PR #3, opened 2019-04-07
Kfold

ysimillides:


ablaom:
@ysimillides Small point, but looks to me as if this belongs in MLJ/utilitiies not MLJBase/utilitiies .

PR #12, opened 2019-05-28
Change `NTuple{<:T}` to `NTuple{N,<:T} where N`

giordano:
See https://github.com/JuliaLang/julia/issues/32170.  This will make the tests of `MLJ.jl` pass on Julia nightly

### DIFF
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -152,7 +152,7 @@ function best end
 clean!(model::Model) = ""
 
 # fallback trait declarations:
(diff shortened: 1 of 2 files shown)


giordano:
I think commits that have been pushed 20 minutes ago already contain this fix?

ablaom:
Yes, I already noticed this error but hadn't pushed yet.

ablaom:
However, nightly still failing. :-( Perhaps I missed something.

giordano:
It's this issue: https://github.com/FugroRoames/TypedTables.jl/issues/48  The package needs a new stable version

PR #17, opened 2019-07-01
Add rng argument to shuffle

ayush-1506:
To support changes made in https://github.com/alan-turing-institute/MLJ.jl/pull/179


ablaom:
This is great. However, for uniformity with ensembles interface, described in https://github.com/alan-turing-institute/MLJ.jl/issues/178, let's also allow rng keyword to be integer, in which case we use MersenneTwisiter(rng).

PR #27, opened 2019-08-17
Fix dataframe deprecation warning

tlienart:
Corresponds to a message on slack,

```
task = load_reduced_ames();
┌ Warning: getindex(df::DataFrame, col_ind::ColumnIndex) is deprecated, use df[!, col_ind] instead.
│   caller = load_reduced_ames() at datasets.jl:18
└ @ MLJBase ~/.julia/packages/MLJBase/vdhww/src/datasets.jl:18
```

This is due to the update https://discourse.julialang.org/t/release-announcements-for-dataframes-jl/18258/38 

PS: this only makes sense along with a compat requirement to have DataFrames be `>=0.19.0` otherwise it will error; so this should be merged when we're happy to use the latest DataFrames (currently clashes with requirements from us to have Categorical Arrays to be of a specific version)

### DIFF
--- a/src/datasets.jl
+++ b/src/datasets.jl
@@ -21,7 +21,7 @@ having six numerical and six categorical features."""
 function load_reduced_ames()
     df = CSV.read(joinpath(datadir, "reduced_ames.csv"), copycols=true,
                   categorical=true)
-    df[:target] = exp.(df[:target])
+    df[!, :target] = exp.(df[:target])
     # TODO: uncomment following after julia #29501 is resolved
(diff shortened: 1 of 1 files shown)


ablaom:
Thanks for that. 

I think a better solution is to use `MLJBase.selectcols`, which does column selection through Tables.jl interface. That way we really don't care what version of DataFrames is being used, assuming DataFrames updates their Tables.jl implementation, which I imagine they would. In any case, this shifts responsibility for the warning to where it belongs. 

As I am refactoring Base anyhow, I will do this. Thanks.

PR #64, opened 2019-10-21
using significant digits rounding in show of univariatefinite

tlienart:
Before

```
UnivariateFinite(f=>0.313470891723578, m=>0.686529108276422)
```

After

```
UnivariateFinite(f=>0.313, m=>0.687)
```

**Note**: uses *sigdigits* rounding so if there are very low numbers, they won't get zeroed but showed as `5.67e-5`:

```
julia> round(0.00001351035, sigdigits=3)
1.35e-5
```


ablaom:
Brilliant. Can we have this rebased onto dev please?

tlienart:
oops

PR #65, opened 2019-10-24
Improve efficiency of LittleDict use

oxinabox:
Where possible it is good to avoid the `Vector{Pair}`
constructor for `LittleDict`
Since the `LittleDict` constructor will have to iterate that and create a Vector of keys, and a Vector of values.

So if you already know your keys and values sperately, you can avoid that.
If they are `Vectors` it will still copy them, but it is a more CPU pipelining friendly copy,

Also being able to have a `Tuple` for keys feels like a good thing.
It certainly is for performance.
But semantically I assume people should not add new classes after the fact

### DIFF
--- a/src/distributions.jl
+++ b/src/distributions.jl
@@ -199,18 +199,14 @@ function average(dvec::AbstractVector{UnivariateFinite{L,U,T}};
     end
 
     # get all refs:
-    refs = reduce(union, [keys(d.prob_given_class) for d in dvec])
-
-    # pad each individual dicts so they have common keys:
-    z = LittleDict{U,T}([x => zero(T) for x in refs])
-    prob_given_class_vec = map(dvec) do d
-        merge(z, d.prob_given_class)
-    end
+    refs = Tuple(reduce(union, [keys(d.prob_given_class) for d in dvec]))
 
     # initialize the prob dictionary for the distribution sum:
-    prob_given_class = LittleDict{U,T}()
-    for x in refs
-        prob_given_class[x] = zero(T)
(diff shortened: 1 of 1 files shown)


src/distributions.jl
```diff
@@ -232,15 +229,10 @@ function average(dvec::AbstractVector{UnivariateFinite{L,U,T}};
...
-
 end
 
 function _pdf(d::UnivariateFinite{L,U,T}, ref) where {L,U,T}
-    if haskey(d.prob_given_class, ref)
-        return d.prob_given_class[ref]
-    else
-        return zero(T)
-    end
+    return get(d.prob_given_class, ref, zero(T))
```
oxinabox: Fasterr since don't need to do two lookups

src/distributions.jl
```diff
@@ -336,7 +328,9 @@ function Distributions.fit(d::Type{<:UnivariateFinite},
     isempty(vpure) && error("No non-missing data to fit. ")
     N = length(vpure)
     count_given_class = Dist.countmap(vpure)
-    prob_given_class = LittleDict([x=>c/N for (x, c) in count_given_class])
+    classes = Tuple(keys(count_given_class))
+    probs = values(count_given_class)./N
+    prob_given_class = LittleDict(classes, probs)
```
oxinabox: This is a less oobvious performance win than the other.
And I haven't benchmarked it.
But I think it i nice for consitency to have `Tuple` keys for the classes,
and the way of expressing values with a broadcast division, is more readable, I think

ablaom:
Many thanks for this!

 Can you please rebase onto dev branch?

tlienart:
closing as superseded by #66; with apologies for being git-illiterate

Issue #77, opened 2019-11-01
Review of loading time


tlienart:
@ablaom skip to the last message for a summary

In line with https://github.com/alan-turing-institute/MLJModels.jl/issues/118 

Initial considerations:

- slowest package to load is `ScientificTypes` by a huge margin (7 seconds;  all other packages take like 0.5)
- the series of export is slower than an unrolled one

I'll keep investigating

I'm a bit confused; loading ST independently takes 0.1s;  loading it with CategoricalArrays is significantly slower but still not 7s.

tlienart:
culprit seems to be categorical arrays:

```
@time using CategoricalArrays
8.963653s 
```

tlienart:
this gives an idea:

```
"1-export / base using" = "1-export / base using"
time() - start = 1.0568718910217285
"2 statsbase,missings" = "2 statsbase,missings"
time() - start = 0.38069915771484375
"3 distrib" = "3 distrib"
time() - start = 1.9556851387023926
"4 st" = "4 st"
time() - start = 7.132582902908325
"5 stats,random,intutils" = "5 stats,random,intutils"
time() - start = 0.0036628246307373047
time() - start = 0.2949249744415283
```

tlienart:
In a fresh session,

```
julia> @time using Tables, OrderedCollections, CategoricalArrays
  9.802188 seconds (19.62 M allocations: 1.367 GiB, 5.16% gc time)

julia> @time using StatsBase, Missings 
  1.672245 seconds (4.69 M allocations: 257.037 MiB, 3.42% gc time)

julia> @time using Distributions
  0.693945 seconds (735.33 k allocations: 40.466 MiB)

julia> @time using ScientificTypes
  1.124992 seconds (2.93 M allocations: 144.996 MiB, 6.22% gc time)
```

tlienart:
seems to point out to  categorical arrays being shitty

```
julia> @time using Tables
  0.099370 seconds (142.67 k allocations: 7.347 MiB)

julia> @time using OrderedCollections
  0.181730 seconds (470.81 k allocations: 22.469 MiB, 6.51% gc time)

julia> @time using CategoricalArrays
  9.064157 seconds (18.09 M allocations: 1.295 GiB, 4.98% gc time)
```

tlienart:
There seems to be a massive discrepancy between a terminal REPL and Juno... this is in a fresh repl session (Same julia version); no problem here...

```
julia> @time using Tables
[...]

PR #78, opened 2019-11-05
Hard deps and CSV->DelimitedFiles

tlienart:
This PR

* removes optional dependencies
* uses the stdlib `DelimitedFiles` to load `boston`, etc

See also https://github.com/alan-turing-institute/ScientificTypes.jl/pull/38; this addresses #77 ; ideally merge this in dev second (after merging https://github.com/alan-turing-institute/ScientificTypes.jl/pull/38)


src/datasets_requires.jl
```diff
@@ -1,32 +1,172 @@
...
+#
+#   for (n, st) in zip(names(data), scitype_union.(eachcol(data)))
+#     println(":$n=>$st,")
+#   end
+#
+# 4. copy and paste the result in a coerce
+#
+#  data_table = coerce(data_table, ...)
+#
+# -------------------------------------------------------
```
ablaom: Thanks for the documentation here!

test/runtests.jl
```diff
@@ -29,9 +29,9 @@ end
   @test include("info.jl")
 end
 
-@testset "datasets" begin
-  @test include("datasets.jl")
-end
+# @testset "datasets" begin
+#   @test include("datasets.jl")
+# end
```
ablaom: Can we please have the tests back?

tlienart: oops

tlienart:
ok done, actually I'm glad you spotted it because the code for the datasets wasn't properly included  🙄  should be all good now

ablaom:
It looks like the macro forms of loading data still import CSV:

```julia
macro load_boston()
    quote
        import CSV
        y, X = unpack(load_boston(), ==(:MedV), x->x != :Chas)
        (X, y)
    end
end
```

(As a side note: the raison d'etre for using macros - importing an optional dependency - is gone. Let's keep the macro - which has a different return value from the functions - because otherwise we break MLJ)

tlienart:
thanks yes will remove those

src/datasets.jl
```diff
@@ -1,9 +1,173 @@
...
+end
+
+
+load_boston()       = load_dataset("Boston.csv", COERCE_BOSTON)
+load_reduced_ames() = load_dataset("reduced_ames.csv", COERCE_REDUCED_AMES)
+load_ames()         = load_dataset("ames.csv", COERCE_AMES)
+load_iris()         = load_dataset("iris.csv", COERCE_IRIS)
+load_crabs()        = load_dataset("crabs.csv", COERCE_CRABS)
+
+load_crabs()
```
ablaom: I'm not sure this fixes the travis fail, but line 166 is presumably not supposed to be here.

Issue #84, opened 2019-11-06
multiclass auc


tlienart:
when the base ConfMat metrics are in and tested, in some future could add multiclass AUC

see https://www.rdocumentation.org/packages/pROC/versions/1.15.3/topics/multiclass.roc for references

Issue #60, opened 2019-10-20
Add performance measures from MLBase


OkonSamuel:
Would there be any need to export MLBase performance evaluation measures like confusion matrix, true and false negative rates, ROC etc. from MLJBase or would it be better using them from MLBase.

tlienart:
Hello @OkonSamuel ! thanks for your help, there is definitely need for such things and I believe @vollmersj has already started looking at it. 

@vollmersj maybe you could list what you've already done / what's up for grabs / what you'd need help with?

Cheers

ablaom:
https://github.com/alan-turing-institute/MLJ.jl/issues/277

tlienart:
Most of it  was done in  #88,  there's still things up for  grabs in #95

PR #88, opened 2019-11-06
Confusion matrix and related metrics

tlienart:
adds functions

* `confusion_matrix` (also `confmat`)
* `accuracy` (1 - mcr)
* `precision`, `recall` (+ `sensitivity`), `auc`, `specificity` (+ `selectivity`), `f1score` as callable measure
* `FScore{beta}` 
* auxilliary functions  `truepositive`, `truepositive_rate` etc

the CM can be used for multiclass as well and eventually we should add a multiclass auc see #84 

Finally there's a bit of re-org in view of possibly modularizing in the future and extracting  metrics, I've just gathered the metrics-related files in one folder. 

Also some minor cleanup here and there (like removing a stray `@show` in `distributions`)

To facilitate the review:

- main changes: https://github.com/alan-turing-institute/MLJBase.jl/pull/88/files#diff-46dbee1cc440de145a5046799515efe1R132-R321
- tests: https://github.com/alan-turing-institute/MLJBase.jl/pull/88/files#diff-0c5aa36b3340a0854f75e096fbfe9cd4R29-R77

cross referencing:

- https://github.com/alan-turing-institute/MLJ.jl/issues/277
- https://github.com/alan-turing-institute/MLJBase.jl/issues/60


tlienart:
Hmm the last fail is due to some traits stuff that wasn’t on Dev before ...

ablaom:
(The fail is because you have taken the TRAIT_GIVEN_FUNCTION declarations out of __init__ again.)

Thanks indeed to @vollmersj  and @tlienart for their work on this important PR. There are a couple of issues here, the first essential, the second more a design preference.

**1. Meausures cannot have keyword arguments**. 

 In particular, in binary classification problems where a choice of "true" class is needed for the measure to make sense, this should be a *field* of the measure *type* which the user must specify on constructing the measure *object*. The rest of the API depends on this kind of convention. For example, in `evaluate!` *there is no mechanism for you to pass keyword arguments to the measures being computed there*.  

For example, if you want to use a cubic loss function from LossFunctions.jl, the existing workflow is like this:

```julia
julia> l3 = LPDistLoss(3)
LPDistLoss{3}()

julia> evaluate!(mach, measure=l3, ....)
```

So, for example, you should handle a measure like "true positive" like this (more-or-less): 

```julia
struct TruePositive{T}
    truth::T
end

(t::TruePositive)(yhat, y) = (yhat .== t.truth) .& (y .== t.truth) |> sum
```

with user workflow

```julia
tp = TruePositive("male")
tp(yhat, y)
evaluate!(mach, measure=tp, ....)
```

**2. Regarding confusion matrices**

I think confusion *matrices* are dumb because you need to specify the "true class" and they don't generalise to the multiclass case. Couldn't we simply have a confusion *table* instead? The column headings are the class labels, in the order the labels appear in `classes(any_element_of_y)`. (You could still *call* it a confusion matrix to avoid the innevitable "Where are confusion matrices?" and I don't think anyone would care it isn't a matrix. )

tlienart:
the keyword thing is addressed;

as for the confusion matrix, I'm not sure I get your points, basically as it stands it's what you'd expect from what you describe to be the "confusion table"; the ordering of classes follows that of `levels` which I think is perfectly reasonable; you can swap it in the binary case with the keyword `rev=true`.

src/measures/finite.jl
```diff
@@ -156,166 +157,187 @@ function ConfusionMatrix(m::Matrix{Int})
...
-struct FScore{β} <: Measure end
-struct AUC <: Measure end
+metadata_measure(Accuracy;
+    name="accuracy",
+    target=AbstractVector{<:Finite},
+    pred=:deterministic,
+    orientation=:score,
+    reports_each=false,
+    feat_dep=false,
+    weights=true)
```
ablaom: Can you please not use abbreviations for the trait names. Then I have to remember *two* handles for the different objects. This is unnecssary terseness, in my opinion.  (Same goes for the model metadata, but I suppose it's too late for that now). 

Also, I think this a bit premature. If we add more traits, you have to change all this later.

tlienart:
A couple of notes from looking at MLR https://mlr.mlr-org.com/articles/tutorial/performance.html

* they expose all the auxiliary metrics like FPR, PPV etc; it wouldn't be very hard to do that too, there's just an API question of whether we'd call them with abbreviated name or not but I guess both can be supported setting `const fpr = false_positive_rate`
* they don't have  an option to set what's positive/negative as far as I can tell; but their effective `show` is something we could mimick which shows clearly the class name above the column / rows

ablaom:
**Regarding AUC**

1. Okay, I've revisited (binary) AUC and can show that the true AUC score of a classifier does not depend on the choice of "true" label (although the ROC naturally does depend on this). I have not checked if this is true for common AUC *estimates*, but would be suspicious of any definition of AUC estimate that does not preserve this invariance. So our discussion on interface points for the "true" label is moot in this case.

2. I have not checked if mutliclass AUC is independent of class ordering, but if it is not it would not want to implement it anyway. 

3. As a side note, I think we should be cautious about using AUC in tutorials. Despite it's widespread use, it has serious flaws, not least of which is that it is not a proper scoring rule.  Maybe `brier` is a better choice when `cross_entropy` is problematic. 

**More discussion about interface points for "true" class**

I have had long think about this issue. Some thoughts:

4. @tlienart I agree with your criticism on slack that it is best to avoid needing to "show" a measure data to instantiate it, for this complicates systematic benchmarking of a model against multiple data sets. So my suggestion `true_positive = TruePositive{"male"}` is flawed. 

5. I think that I could accept the implementation of the binary measures you suggest if their use is restricted to `OrderedFactor{2}` data, where a default order is part of the data prescription. Otherwise, I think there is too great a danger for insidious errors if the user is not confronted to explicitly declare a "true" label. For example, the default "true" (according to the lexicographical ordering implemented by the `categorical` constructor) in the case of levels "accept/reject" is `false="accept"` and `true="reject"`, which is counter-intuitive. 

6. We could deal with `Multiclass` `y` safely, although less satisfactorily,  in two ways:

    (i) The binary measure returns an informative error, explaining the measure called can only be used on `OrderedFactor{2}` data. This is more restrictive but requires less explanation.

    (ii) the binary measures return a callable object instead of a number. For example, `true_positive(yhat, y)` returns a callable object `c`, such that:

    - the `show` methods for `c` return something like `25 if true=$default` (callable object)`, where  `default=levels(y)[2]`
    - `c(levels)` is the true_positive if we declare levels[1] to be the "false" level (see also below)

    To get the default, you just do `c(levels(y))`. 

**confusion matrix**

Similar suggestions concern `confusion_matrix` with the exception that I don't think restricting use to `OrderedFactor` data is an option. We need an analogue of (ii) for `Multiclass` data:  `confusion_matrix(yhat, y)` returns a callable object `c` where `c(levels)` has the obvious meaning (and levels needn't cover all the values of `levels(y)` (more precisely, `classes(y[1])`).  
    
Thoughts?

tlienart:
Ok the recent commit addresses I think everything with a few minor changes which I hope you won't mind  too much, I'll follow your comments to try to simplify review:

### AUC

1. yes, this is now checked explicitly
1. Multiclass AUC is not implemented in this PR (and I agree with your statement)
1. noted

### True class etc

Ok I've pretty much done what you suggested apart from the fact that I do still allow metrics on `Multiclass{2}` BUT they will show a warning with the explicit ordering used; this warning can be suppressed by specifying an ordering explictly. 

In short, all metrics fallback on metrics on the confusion matrix. When the confusion matrix is built, it checks whether an order is specified and if not, if `y` is `OrderedFactor`, if it isn't a warning is displayed:

```julia-repl
julia> y = categorical(['a','b','a','b']);
julia> ŷ = categorical(['a','a','b','b']);
julia> confmat(ŷ, y)
┌ Warning: The classes are un-ordered,
│ using: negative='a' and positive='b'.
│ To suppress this warning, consider coercing to OrderedFactor.
└ @ MLJBase ~/.julia/dev/MLJBase/src/measures/confusion_matrix.jl:74
              ┌───────────────────────────┐
              │       Ground Truth        │
┌─────────────┼─────────────┬─────────────┤
│  Predicted  │      a      │      b      │
├─────────────┼─────────────┼─────────────┤
│      a      │      1      │      1      │
├─────────────┼─────────────┼─────────────┤
│      b      │      1      │      1      │
└─────────────┴─────────────┴─────────────┘
```

this warning will show with any of the derived metrics

```julia
julia> recall(ŷ, y)
┌ Warning: The classes are un-ordered,
│ using: negative='a' and positive='b'.
│ To suppress this warning, consider coercing to OrderedFactor.
└ @ MLJBase ~/.julia/dev/MLJBase/src/measures/confusion_matrix.jl:74
0.5
```

the warnings will  not show if OrderedFactor.

There are also keywords constructor which allow the user to do what I call above "specifying the ordering", in which case it's assumed the user knows what they're doing and all warnings are suppressed:

```julia
julia> rec = Recall(rev=false)
julia> rec(ŷ, y)
0.5
```

### extra

the show is now much nicer as shown above.

-----

I hope that kind of makes sense; I thought it would be nicer to allow user to smoothly transition from stuff they may be used to rather than face them with errors...

ablaom:
@tlienart Thank you for taking my suggestions on board and, I believe, improving on them in the updated implementation! 

My remaining concern is that because `true_positive` etc derive from a confusion matrix, we cannot add them to the list of measures in an `evaluation!` call. So, one would like to do something like this:

```julia
evaluate!(mach, resampling=CV(), measure = [misclassification_rate, recall, true_positive, false_positive])
```

1. Is there a reason you do not want to implement them as stand-alone measures, with behaviour mirroring what have for `recall`? 

2. We should implement `Statistics.mean` for `ConfusionMatrix` objects, so that it can be aggregated in performance evaluations (so that specifying `measure=confusion_matrix` works).

tlienart:
I was just thinking that the implementation does not support missing values in the ground truth; this should be easy to circumvent (just skip those) but I’ll be happy to hear ideas of other things that could be done, it’s a bit pathological though

tlienart:
Right, see also issue #95 , I added a _lot_ of metrics then ended up getting a bit bored and also thinking that this might be a pain to  review... 

Anyway  long story short:

1. all are now metrics, so you can do  `tp` `tpr` and `TPR(rev=true)` in the same way you can call recall etc
1. I did **not** implement the `Statistics.mean` for confusion matrix as I'm not 100% clear on what you want here (do you want  confusion matrix to be a measure itself?), if you could open an issue with very short guidelines I'll do that in another PR
1. I added a bunch of oother stuff like Matthews correlation, Balanced accuracy, PPV etc and  where appropriate I tested the implementations explicitly with sklearn which should help  the review.

I'd suggest not adding anything to this PR (apart from fixes suggested in reviews) and possibly opening new issues for stuff that could be added  in that vein.

I also thought that there's  now a  lot of blueprint that people can  just follow to add more metrics which makes for good and easy isssues.

ablaom:
@tlienart That's fantastic. 

Regarding point 2: You are correct, of course: `confusing_matrix` (and other measures such as `true_positive`) should be summed and not averaged in aggregations. But this points to a gap in the measure interface. We need a new trait `aggregation` defaulting to mean, and a corresponding change to `evaluate!` to call the appropriate method.

Here's a plan to finish this PR:

- [ ] Add `aggregration` trait to this PR @ablaom 

- [ ] Create new branch MLJ#confmat (same name as this one) with necessary changes to `evaluate!` @ablaom 

- [ ] Overload trait value to summation for appropriate measures introduced here (@ablaom to initiate)

- [ ] @ablaom to finish detailed code review

- [ ] @tlienart to check trait values and address any remaining issues

ablaom:
Oops. Closed by mistake.

ablaom:
For the record, I'm implementing :mean for the aggregation of the F-score, although the non-linearity of its definition means "interpret with caution"; see, eg, https://datascience.stackexchange.com/questions/16179/what-is-the-correct-way-to-compute-mean-f1-score . I think the best we can do is document our choice clearly.

ablaom:
Ditto rms. 

Okay, we allow `aggregation` to return any callable object, with built-in `Sum()` and `Mean()` and `Mean()` as default. And we can do a custom one for `rms`, etc

tlienart:
FYI, I've added docstrings for AUC, BACC, and MCC but it's on the `roc` branch which I'll push soon. Just so  that you don't add these yourself 😄

ablaom:
Okay, I am done with my additions (see above) and my review. Not covered above:

- added "registry" for measures (essentially just list of all measure types, auto-generated for now)

- add facility to query measure registry as we do models. Do `?measures` for details.

Although technically non-breaking, I suggest we make this part of a minor release as MLJ will not really work properly for all the new measures that don't use `Mean()` as the aggregation method. (And we  should consider the existing MLJ resampling as flawed in the case of `rms` family of metrics, since these are currently mean-aggregated. My bad.) 

I have left `confusion_matrix` as a non-measure because it looks complicated to fix, and I don't think it is a great loss. It's just a pity that the API for it is slightly different. 

@tlienart Please review the aggregation trait values for each measure and pull when ready. Naturally, for some complicated measures, there is no ideal aggregation method (eg, Score) and I left them as `Mean()`.

```julia
julia> for m in measures()
       println("$(m.name) \t$(m.aggregation)")
       end

auc     MLJBase.Mean()
accuracy        MLJBase.Mean()
balanced accuracy       MLJBase.Mean()
cross_entropy   MLJBase.Mean()
FScore  MLJBase.Mean()
false discovery rate    MLJBase.Mean()
false negative  MLJBase.Sum()
false negative rate (miss rate)         MLJBase.Mean()
false positive  MLJBase.Sum()
false positive rate (fallout)   MLJBase.Mean()
l1      MLJBase.Mean()
l2      MLJBase.Mean()
mav     MLJBase.Mean()
matthews_correlation    MLJBase.Mean()
misclassification_rate  MLJBase.Mean()
negative predictive value       MLJBase.Mean()
precision (positive predictive value)   MLJBase.Mean()
rms     MLJBase.RootMeanSquare()
rmsl    MLJBase.RootMeanSquare()
rmslp1  MLJBase.RootMeanSquare()
rmsp    MLJBase.RootMeanSquare()
true negative   MLJBase.Sum()
true negative rate (specificity, selectivity)   MLJBase.Mean()
true positive   MLJBase.Sum()
true positive rate (sensitivity, recall, hit rate)      MLJBase.Mean()
BrierScore{UnivariateFinite}    MLJBase.Mean()
LossFunctions.DWDMarginLoss     MLJBase.Mean()
LossFunctions.ExpLoss   MLJBase.Mean()
LossFunctions.L1HingeLoss       MLJBase.Mean()
LossFunctions.L2HingeLoss       MLJBase.Mean()
LossFunctions.L2MarginLoss      MLJBase.Mean()
LossFunctions.LogitMarginLoss   MLJBase.Mean()
LossFunctions.ModifiedHuberLoss         MLJBase.Mean()
LossFunctions.PerceptronLoss    MLJBase.Mean()
LossFunctions.ScaledMarginLoss  MLJBase.Mean()
LossFunctions.SigmoidLoss       MLJBase.Mean()
LossFunctions.SmoothedL1HingeLoss       MLJBase.Mean()
LossFunctions.ZeroOneLoss       MLJBase.Mean()
LossFunctions.HuberLoss         MLJBase.Mean()
LossFunctions.L1EpsilonInsLoss  MLJBase.Mean()
LossFunctions.L2EpsilonInsLoss  MLJBase.Mean()
LossFunctions.LPDistLoss        MLJBase.Mean()
LossFunctions.LogitDistLoss     MLJBase.Mean()
LossFunctions.PeriodicLoss      MLJBase.Mean()
LossFunctions.QuantileLoss      MLJBase.Mean()
LossFunctions.ScaledDistanceLoss        MLJBase.Mean()
```

ablaom:
Just adding one more thing: simplification of name trait values.

src/measures/finite.jl
```diff
@@ -375,22 +405,22 @@ metadata_measure.((TruePositive, TrueNegative, TPR, TNR, Precision, FScore, NPV)
...
+name(::Type{<:FNR}) = "fnr"
 
-name(::Type{<:FDR}) = "false discovery rate"
-name(::Type{<:NPV}) = "negative predictive value"
+name(::Type{<:FDR}) = "fdr"
+name(::Type{<:NPV}) = "npv"
 
-name(::Type{<:Precision}) = "precision (positive predictive value)"
+name(::Type{<:Precision}) = "ppv"
+name(::Type{<:FScore{β}}) where β = "FScore{$β}()"
```
tlienart: why is this different from the rest?

ablaom: Okay, so "name" is mostly just a handle to be used by the show methods. The idea though is that you can also just type that name at the REPL and it makes sense (if it's from an external package you will need to do `using Package` first but the package qualifying name does not clutter the show output if you just`import` it). The trait `docstring` is for a short description (a proper name, like "Matthew's correlation" plus aliases.

ablaom:
Yeah, sorry. I just dumped those extra descriptors. I will put them back via `docstring`.

tlienart:
sure no worries, fine by me, we can do that in another PR btw, maybe merge this so that you can do the other stuff and  tag me in an issue to add docstrings?

PR #100, opened 2019-11-13
POC @static macro

tlienart:
Note that ideally we'd need to change this naming for something else because the `@static` macro is already taken by Base and it leads to annoying warning about it breaking compilation and similar  nonsense.

Given that this feature is not set in stone, we could imagine calling the whole thing differently with  possible candidates:

* `Operation` and `@operation`
* `Static` and  `@set_static` or `@declare_static` or some other alternative
* `Fixed` and `@fixed` 
* `StaticTransform` and `@statictf` or `@static_tf` or `@static_transform`

etc

Maybe "fixed" is closest to static, not sure.

Context for this discussion, https://github.com/alan-turing-institute/MLJ.jl/issues/291

### DIFF
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -57,6 +57,7 @@ export std
 
 import Base.==
 import Base: @__doc__
+import Base.@static
 
 using Tables
 using OrderedCollections # already a dependency of StatsBase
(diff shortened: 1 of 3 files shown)


tlienart:
closing this in view of https://github.com/alan-turing-institute/MLJ.jl/issues/291#issuecomment-554064576

PR #120, opened 2019-11-21
removing  set_defaults

tlienart:
It's not used  in MLJModels or MLJ so I think it's safe to just remove it.

Though it would be breaking so for a minor release.


ablaom:
`@set_defaults` is still used in MLJ to defined new model  types in  the `@from_network` macro. So this will have to wait, sorry.

ablaom:
I'm closing as this is now old, there are now merge conflicts, and, given its low priority, is unlikely to see the light of day

PR #122, opened 2019-11-22
fixing AUC

tlienart:
ok there was an embarrasing flipped sign issue in the  AUC computation meaning that what we were getting was `1-auc` instead of `auc` ... 

So this fixes that, and also fixes the tests.

I'll do a patch release as well...


ablaom:
Good catch, thanks!

PR #136, opened 2019-11-29
Adding make_moons, make_circles, make_blobs functions

davidbp:
Here you have some functions to generate basic datasets. I hope it will be usefull for inspecting clustering and classification methods.

### DIFF
--- a/src/datasets_synthetic.jl
+++ b/src/datasets_synthetic.jl
@@ -1,9 +1,6 @@
 module DatasetsSynthetic
 
-using Test
 using Random
-using MLJBase
-
 export make_blobs, make_circles, make_moons
 
 uniform_sample_in_zero_maxval(p, maxval) = maxval .* (1 .- rand(p))
--- a/src/datasets_synthetic.jl
+++ b/src/datasets_synthetic.jl
@@ -1,0 +1,193 @@
+
+using Random
+
+uniform_sample_in_zero_maxval(n_features, maxval) = maxval.*(1 .- rand(n_features)) 
+uniform_sample_in_minval_maxval(n_features, minval, maxval) = (maxval-minval)* rand(n_features) .+ minval 
+normal_sample(n_features, mu, var) = mu .+ sqrt(var) .* randn(n_features)
+
+"""
+    Return evenly spaced numbers over a specified interval.
+    
+    Returns `num` evenly spaced samples, calculated over the
+    interval [`start`, `stop`].
+    
+    The endpoint of the interval can optionally be excluded.
+
+"""
+function linspace(start, stop; num=50, endpoint=true, retstep=false)
+    dif  = stop-start
+    step = dif/num
+    result = []
+       
+    if endpoint 
+        for i in 1:num
+            push!(result, start+i*step)
+        end
+    else
+        for i in 1:num
+            push!(result, start+(i-1)*step)
+        end
+    end
+    return result
+    
+end
+
+
+"""
+Shuffles the rows of an Array `X` and the values of a vector `y` using a randomly
+generated permutation. The same permutation is used to shuffle both `X` and `y`.
+"""
+function shuffle_Xy(X, y; random_seed=1234)
+    Random.seed!(random_seed)
+    perm = randperm(length(y))
+    return X[perm,:], y[perm]
+end
+
+
+"""
+    make_blobs(n_examples=100; 
+               n_features=2, 
+               centers=3, 
+               cluster_std=1.0, 
+               center_box=(-10.,10.),
+               element_type=Float64,
+               random_seed=1234,
+               return_centers=false)
+               
+     Generates a dataset with `n_examples` of dimension `n_features` and returns a vector containing
+     as integers the membership of the different points generated.
+
+     The data is roughly grouped around several `centers`  which are created using `cluster_std`. 
+     The data lives inside `center_box` in the case it is randomly generated.
+     
+     - If `centers` is an integer the centroids are created randomly.    
+     - If `centers` is an Array containing points the centroids are picked from `centers`.
+     - If `return_centers=true` the centroids of the bloods are returned
+
+"""
+function make_blobs(n_examples=100; n_features=2, 
+                    centers=3, cluster_std=1.0, center_box=(-10.,10.),
+                    element_type=Float64, random_seed=1234, return_centers=false)
+
+    Random.seed!(random_seed)
+
+    X = [] 
+    y = []
+
+    if typeof(centers)<: Int
+        n_centers = centers
+        centers = []
+        for c in 1:n_centers
+            center_sample = uniform_sample_in_minval_maxval(n_features, center_box[1], center_box[2])
+            push!(centers, center_sample)
+        end
+    else
+        n_centers = length(centers)
+    end
+
+    if typeof(cluster_std)<: AbstractFloat
+        cluster_std = 0.5 * randn(n_centers)
+    end
+    
+    # generates the nunber of example for each center
+    n_examples_per_center = [div(n_examples, n_centers) for x  in 1:n_centers]
+    for i in 1:(n_examples % n_centers)
+        n_examples_per_center[i] += 1
+    end
+
+    # generates the actual vectors close to each center blob 
+    for (i, (n, std, center)) in enumerate(zip(n_examples_per_center, cluster_std, centers))
+        X_current = center' .+ std .* randn(element_type, (n, n_features))
+        push!(X, X_current)
+        push!(y, [i for k in 1:n]) 
+    end
+
+    # stack all the previous arrays created for each of the centers
+    X = cat(X..., dims=1)
+    y = cat(y..., dims=1)
+    
+    if return_centers
+        return X, y, centers  
+    else
+        return X, y
+    end
+end
+
+
+
+"""
+    make_circles(n=100; shuffle=true, noise=0., random_seed=1234, factor=0.8)
+    
+    Generates a dataset with `n` bi-dimensional examples. Samples are created 
+    from two circles. One of the circles inside the other. The `noise` scalar
+    can be used to add noise to the generation process. The scalar `factor` 
+    can be used to make the radious of the smallest circle smaller.
+
+"""
+function make_circles(n=100; shuffle=true, noise=0., random_seed=1234, factor=0.8)
+    
+    @assert 0 <= factor <=1  #factor in [0,1] 
+
+    n_out = div(n, 2)
+    n_in = n - n_out
+     
+    linspace_out = linspace(0, 2 * pi; num=n_out, endpoint=false)
+    linspace_in  = linspace(0, 2 * pi; num=n_in, endpoint=false)
+
+    outer_circ_x = cos.(linspace_out)
+    outer_circ_y = sin.(linspace_out)
+    inner_circ_x = cos.(linspace_in) .* factor
+    inner_circ_y = sin.(linspace_in) .* factor
+    
+    X = [[outer_circ_x..., inner_circ_x...] [outer_circ_y..., inner_circ_y...]]
+    y = [ones(Int,n_out)..., 2*ones(Int,n_in)...]
+
+    if shuffle
+       X, y = shuffle_Xy(X, y ; random_seed=random_seed)
+    end
+    
+    if noise >0.
+        X .= X .+ noise.* rand(size(X)...)
+    end
+
+    return X,y
+end
+
+
+"""
+    make_moons(n=100; shuffle=true, noise=0., translation=0.5, factor=1.0, random_seed=1234)
+    
+    Generates `n` examples sampling from two moons. The `noise` can be changed to add
+    noise to the samples.
+
+"""
+function make_moons(n=100; shuffle=true, noise=0., translation=0.5, factor=1.0, random_seed=1234)
+    
+    @assert 0 <= factor <=1  
+
+    n_out = div(n, 2)
+    n_in = n - n_out
+     
+    linspace_out = linspace(0, pi; num=n_out, endpoint=true)
+    linspace_in  = linspace(0, pi; num=n_in, endpoint=true)
+
+    
+
+    outer_circ_x = cos.(linspace_out)
+    outer_circ_y = sin.(linspace_out)
+    inner_circ_x = 1 .- cos.(linspace_in) .* factor
+    inner_circ_y = 1 .- sin.(linspace_in) .* factor .- translation
+    
+    X = [[outer_circ_x..., inner_circ_x...] [outer_circ_y..., inner_circ_y...]]
+    y = [ones(Int,n_out)..., 2*ones(Int,n_in)...]
+
+    if shuffle
+       X, y = shuffle_Xy(X, y ; random_seed=random_seed)
+    end
+    
+    if noise >0.
+        X .= X .+ noise.* rand(size(X)...)
+    end
+
+    return X,y
+end
--- a/src/datasets_synthetic.jl
+++ b/src/datasets_synthetic.jl
@@ -1,36 +1,9 @@
 
 using Random
 
-uniform_sample_in_zero_maxval(n_features, maxval) = maxval.*(1 .- rand(n_features)) 
-uniform_sample_in_minval_maxval(n_features, minval, maxval) = (maxval-minval)* rand(n_features) .+ minval 
-normal_sample(n_features, mu, var) = mu .+ sqrt(var) .* randn(n_features)
-
-"""
-    Return evenly spaced numbers over a specified interval.
-    
-    Returns `num` evenly spaced samples, calculated over the
-    interval [`start`, `stop`].
-    
-    The endpoint of the interval can optionally be excluded.
-
-"""
-function linspace(start, stop; num=50, endpoint=true, retstep=false)
-    dif  = stop-start
-    step = dif/num
-    result = []
-       
-    if endpoint 
-        for i in 1:num
-            push!(result, start+i*step)
-        end
-    else
-        for i in 1:num
-            push!(result, start+(i-1)*step)
-        end
-    end
-    return result
-    
-end
+uniform_sample_in_zero_maxval(p, maxval) = maxval .* (1 .- rand(p)) 
+uniform_sample_in_minval_maxval(p, minval, maxval) = (maxval-minval) .* rand(p) .+ minval 
+normal_sample(p, mu, var) = mu .+ sqrt(var) .* randn(p)
 
 
 """
@@ -45,27 +18,26 @@ end
 
 
 """
-    make_blobs(n_examples=100; 
-               n_features=2, 
-               centers=3, 
-               cluster_std=1.0, 
-               center_box=(-10.,10.),
-               element_type=Float64,
-               random_seed=1234,
-               return_centers=false)
-               
-     Generates a dataset with `n_examples` of dimension `n_features` and returns a vector containing
-     as integers the membership of the different points generated.
+make_blobs(n=100; 
+           p=2, 
+           centers=3, 
+           cluster_std=1.0, 
+           center_box=(-10.,10.),
+           element_type=Float64,
+           random_seed=1234,
+           return_centers=false)
+           
+ Generates a dataset with `n` of dimension `p` and returns a vector containing
+ as integers the membership of the different points generated.
 
-     The data is roughly grouped around several `centers`  which are created using `cluster_std`. 
-     The data lives inside `center_box` in the case it is randomly generated.
-     
-     - If `centers` is an integer the centroids are created randomly.    
-     - If `centers` is an Array containing points the centroids are picked from `centers`.
-     - If `return_centers=true` the centroids of the bloods are returned
-
+ The data is roughly grouped around several `centers`  which are created using `cluster_std`. 
+ The data lives inside `center_box` in the case it is randomly generated.
+ 
+ - If `centers` is an integer the centroids are created randomly.    
+ - If `centers` is an Array containing points the centroids are picked from `centers`.
+ - If `return_centers=true` the centroids of the bloods are returned
 """
-function make_blobs(n_examples=100; n_features=2, 
+function make_blobs(n=100; p=2, 
                     centers=3, cluster_std=1.0, center_box=(-10.,10.),
                     element_type=Float64, random_seed=1234, return_centers=false)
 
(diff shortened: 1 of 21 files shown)


src/datasets_synthetic.jl
```diff
@@ -0,0 +1,193 @@
+
+using Random
+
+uniform_sample_in_zero_maxval(n_features, maxval) = maxval.*(1 .- rand(n_features))
```
tlienart: please add spaces around `.*` operator

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,193 @@
+
+using Random
+
+uniform_sample_in_zero_maxval(n_features, maxval) = maxval.*(1 .- rand(n_features)) 
+uniform_sample_in_minval_maxval(n_features, minval, maxval) = (maxval-minval)* rand(n_features) .+ minval
```
tlienart: please add `.` --> `) .* rand`

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,193 @@
...
+            push!(result, start+i*step)
+        end
+    else
+        for i in 1:num
+            push!(result, start+(i-1)*step)
+        end
+    end
+    return result
+    
+end
```
tlienart: use `range`

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,193 @@
...
+     Generates a dataset with `n_examples` of dimension `n_features` and returns a vector containing
+     as integers the membership of the different points generated.
+
+     The data is roughly grouped around several `centers`  which are created using `cluster_std`. 
+     The data lives inside `center_box` in the case it is randomly generated.
+     
+     - If `centers` is an integer the centroids are created randomly.    
+     - If `centers` is an Array containing points the centroids are picked from `centers`.
+     - If `return_centers=true` the centroids of the bloods are returned
+
```
tlienart: Please decrease indentation  of everything, please also kindly abide by the convention so  search and replace all occurences of  `n_features` and `n_samples` or `n_examples` by respectively  `p` and `n`, I  apologise  if you don't like it but it's the convention we use

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,193 @@
...
+    X = [] 
+    y = []
+
+    if typeof(centers)<: Int
+        n_centers = centers
+        centers = []
+        for c in 1:n_centers
+            center_sample = uniform_sample_in_minval_maxval(n_features, center_box[1], center_box[2])
+            push!(centers, center_sample)
```
tlienart: using push! sample by sample is very inefficient, you should pre-generate a bunch of random values  and then act upon them.

tlienart:
Thanks for suggesting this, it would be useful functions to have; a few points:

* please do use the conventions `n` and `p`
* try  to stick with the general coding style, for instance don't indent docstring
* avoid using `push!` and particularly so with a `[]` which is an array of Any
* use `range` instead  of your own `linspace`
* add tests (e.g.: that points are  generated, in the right proportion, with the right dimensionality), you check code coverage to guide you

once that's done I'll do another pass, thanks!

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,193 @@
...
+    else
+        n_centers = length(centers)
+    end
+
+    if typeof(cluster_std)<: AbstractFloat
+        cluster_std = 0.5 * randn(n_centers)
+    end
+    
+    # generates the nunber of example for each center
+    n_examples_per_center = [div(n_examples, n_centers) for x  in 1:n_centers]
```
tlienart: use `fill`

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,193 @@
...
+    for i in 1:(n_examples % n_centers)
+        n_examples_per_center[i] += 1
+    end
```
tlienart: use broadcasting

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,193 @@
...
+
+    # generates the actual vectors close to each center blob 
+    for (i, (n, std, center)) in enumerate(zip(n_examples_per_center, cluster_std, centers))
+        X_current = center' .+ std .* randn(element_type, (n, n_features))
+        push!(X, X_current)
+        push!(y, [i for k in 1:n])
```
tlienart: use fill

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,193 @@
...
+    
+    Generates a dataset with `n` bi-dimensional examples. Samples are created 
+    from two circles. One of the circles inside the other. The `noise` scalar
+    can be used to add noise to the generation process. The scalar `factor` 
+    can be used to make the radious of the smallest circle smaller.
+
+"""
+function make_circles(n=100; shuffle=true, noise=0., random_seed=1234, factor=0.8)
+    
+    @assert 0 <= factor <=1  #factor in [0,1]
```
tlienart: throw proper error messages

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,193 @@
...
+    
+    X = [[outer_circ_x..., inner_circ_x...] [outer_circ_y..., inner_circ_y...]]
+    y = [ones(Int,n_out)..., 2*ones(Int,n_in)...]
+
+    if shuffle
+       X, y = shuffle_Xy(X, y ; random_seed=random_seed)
+    end
+    
+    if noise >0.
+        X .= X .+ noise.* rand(size(X)...)
```
tlienart: `.+=` also since you use  `rand` the noise is never negative

tlienart:
PS: one note, while I would insist that the internal code uses `n` and `p`, feel free to add a keyword argument `n_samples` (which is what is used in sklearn for instance), e.g.:

```julia
function foo(n, p; n_samples=n, n_features=p)
    # ...
end
```

src/datasets_synthetic.jl
davidbp: The code expects to use n >> centers, in the default version centers=3 so I don't think this will be a problem.

src/datasets_synthetic.jl
davidbp: Done, I just removed the if statement and added the `.+=`

src/datasets_synthetic.jl
davidbp: I don´t see how to use broadcasting here. I am picking an index of each prototype at a time untill all n examples are assigned to a prototype. Notices this will be a very small number because `n_per_center = [div(n, n_centers) for x  in 1:n_centers]` was executed before.

src/datasets_synthetic.jl
davidbp: What type of error message you want? The macro already writes precisely the condition that is not met if the user tries to use factor outside [0,1] (that´s a pretty cool use of macros).

src/datasets_synthetic.jl
davidbp: Done!

src/datasets_synthetic.jl
davidbp: Done

src/datasets_synthetic.jl
davidbp: Done

src/datasets_synthetic.jl
davidbp: I use `LinRange` since it is the closest method to my custom `linspace`.

davidbp:
> Thanks for suggesting this, it would be useful functions to have; a few points:
> 
> * please do use the conventions `n` and `p`
> * try  to stick with the general coding style, for instance don't indent docstring
> * avoid using `push!` and particularly so with a `[]` which is an array of Any
> * use `range` instead  of your own `linspace`
> * add tests (e.g.: that points are  generated, in the right proportion, with the right dimensionality), you check code coverage to guide you
> * add  the functions to  `export`
> 
> once that's done I'll do another pass, thanks!

- please do use the conventions n and p
   - Done

- try to stick with the general coding style, for instance don't indent docstring
   - Done

- avoid using push! and particularly so with a [] which is an array of Any
   - I feel the code is more clear this way and it will not affect performance, ( I do it one chunk at a time not one instance at a time).

- use range instead of your own linspace
    -  Done (LinRange instead of my own linspace)

- add tests (e.g.: that points are generated, in the right proportion, with the right dimensionality), you check code coverage to guide you
    -  Need to do this one

- add the functions to export
    - Done

ablaom:
@tlienart Before pulling this, be sure to rebase onto dev first. Currently this PR is a request on to master.

src/datasets_synthetic.jl
tlienart: I understand your  argument but it's  just bad  practice and it's easy to pre allocate here

src/datasets_synthetic.jl
tlienart:
```julia
n_per_center = fill(div(n,n_centers), n_centers)
n_per_center[end] += rem(n,n_centers)
```

src/datasets_synthetic.jl
tlienart:
```julia
0 <= factor <= 1 || throw(ArgumentError("factor should be in [0,1]"))
```

src/datasets_synthetic.jl
```diff
@@ -97,7 +97,7 @@ make_circles(n=100; shuffle=true, noise=0., random_seed=1234, factor=0.8)
 Generates a dataset with `n` bi-dimensional examples. Samples are created
 from two circles. One of the circles inside the other. The `noise` scalar
 can be used to add noise to the generation process. The scalar `factor`
-can be used to make the radious of the smallest circle smaller.
+correspond to the radius of the smallest circle.
```
tlienart: `corresponds`

test/datasets_synthetic.jl
```diff
@@ -1,28 +1,39 @@
...
 
-# using Revise
 using Test
-using MLJBase
 
-# test make_bloobs
-X, y = make_blobs(110; p=2, centers=3)
-@test (110,2) == size(X)
-@test 110 == length(y)
+@testset "make_bloobs Tests" begin
```
tlienart: `blobs`

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,189 @@
...
+make_blobs(n=100;
+           p=2,
+           centers=3,
+           cluster_std=1.0,
+           center_box=(-10.,10.),
+           element_type=Float64,
+           random_seed=1234,
+           return_centers=false)
+
+ Generates a dataset with `n` of dimension `p` and returns a vector containing
```
tlienart: with  `n` samples

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,189 @@
...
+ The data is roughly grouped around several `centers`  which are created using `cluster_std`.
+ The data lives inside `center_box` in the case it is randomly generated.
+
+ - If `centers` is an integer the centroids are created randomly.
+ - If `centers` is an Array containing points the centroids are picked from `centers`.
+ - If `return_centers=true` the centroids of the bloods are returned
+"""
+function make_blobs(n=100; p=2,
+                    centers=3, cluster_std=1.0, center_box=(-10.,10.),
+                    element_type=Float64, random_seed=1234, return_centers=false)
```
tlienart: please add explicit types to things here, `n::Int`, `p::Int`, ... and remove the test whether `cluster_std` is `AbstractFloat`, just make it `::Real` here

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,189 @@
...
+    # generates div(n, n_centers) examples assigned to each center
+    n_per_center = [div(n, n_centers) for x  in 1:n_centers]
+
+    # adds the reamainding examples to each center up to n
+    for i in 1:(n % n_centers)
+        n_per_center[i] += 1
+    end
+
+    # generates the actual vectors close to each center blob
+    for (i, (n_bloob, std, center)) in enumerate(zip(n_per_center, cluster_std, centers))
```
tlienart: `blobs` not `bloob`

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,189 @@
...
+        X_current = center' .+ std .* randn(element_type, (n_bloob, p))
+        push!(X, X_current)
+        push!(y, [i for k in 1:n_bloob])
```
tlienart: same thing, no `push!` if we can easily avoid them.

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,189 @@
...
+
+"""
+make_circles(n=100; shuffle=true, noise=0., random_seed=1234, factor=0.8)
+
+Generates a dataset with `n` bi-dimensional examples. Samples are created
+from two circles. One of the circles inside the other. The `noise` scalar
+can be used to add noise to the generation process. The scalar `factor`
+correspond to the radius of the smallest circle.
+"""
+function make_circles(n=100; shuffle=true, noise=0., random_seed=1234, factor=0.8)
```
tlienart: same comment to  type arguments

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,189 @@
...
+
+    @assert 0 <= factor <=1  "factor is expected to be in [0,1]"
+    @assert 0 <= noise  "noise is expected to be in positive"
+
```
tlienart: see comment on `throw(ArgumentError ... ) `

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,189 @@
...
+    n_out = div(n, 2)
+    n_in = n - n_out
+
+    linrange_out = Array(LinRange(0, 2 * pi, n_out))
+    linrange_in  = Array(LinRange(0, 2 * pi, n_in))
```
tlienart: I don't understand why you're not using `range(0, 2pi, length=n_in)` and if you want a vector, just  use `|> collect`

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,189 @@
...
+
+    outer_circ_x = cos.(linrange_out)
+    outer_circ_y = sin.(linrange_out)
+    inner_circ_x = cos.(linrange_in) .* factor
+    inner_circ_y = sin.(linrange_in) .* factor
+
+    X = [[outer_circ_x..., inner_circ_x...] [outer_circ_y..., inner_circ_y...]]
+    y = [ones(Int,n_out)..., 2*ones(Int,n_in)...]
```
tlienart: splatting is unnecessary here, use `hcat` and `vcat` to the same effect (but better).

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,189 @@
...
+    X .+= noise .* rand(size(X)...)
+
+    return X,y
+end
+
+function make_circles2(n=100; shuffle=true, noise=0., random_seed=1234, factor=0.8)
+
+    @assert 0 <= factor <=1  "factor is expected to be in [0,1]"
+    @assert 0 <= noise  "noise is expected to be in positive"
+
```
tlienart: idem

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,189 @@
...
+    n_out = div(n, 2)
+    n_in = n - n_out
+
+    linrange_out = Array(LinRange(0, 2 * pi, n_out))
+    linrange_in  = Array(LinRange(0, 2 * pi, n_in))
+
```
tlienart: idem

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,189 @@
...
+    outer_circ_x = cos.(linrange_out)
+    outer_circ_y = sin.(linrange_out)
+    inner_circ_x = cos.(linrange_in) .* factor
+    inner_circ_y = sin.(linrange_in) .* factor
+
+    X = [[outer_circ_x..., inner_circ_x...] [outer_circ_y..., inner_circ_y...]]
+    y = [ones(Int,n_out)..., 2*ones(Int,n_in)...]
+
```
tlienart: idem

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,189 @@
...
+"""
+make_moons(n=100; shuffle=true, noise=0.,
+           translation=0.5, factor=1.0, random_seed=1234)
+
+Generates `n` examples sampling from two moons. The `noise` can be changed to add
+noise to the samples.
+"""
+function make_moons(n=100; shuffle=true, noise=0.,
+                   translation=0.5, factor=1.0, random_seed=1234)
```
tlienart: idem

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,189 @@
...
+    y = [ones(Int,n_out)..., 2*ones(Int,n_in)...]
+
+    if shuffle
+       X, y = shuffle_Xy(X, y ; random_seed=random_seed)
+    end
+
+    X .+= noise .* rand(size(X)...)
+
+    return X,y
+end
```
tlienart: so again same comments for this function

tlienart:
Oh,  last thing, please use the code coverage helper indicated above to show how  you can add tests to fully  cover your code; since MLJBase is a base/abstract package, we really want to  strive towards 90%+ coverage. Thanks!

src/datasets_synthetic.jl
davidbp: done

test/datasets_synthetic.jl
davidbp: done

src/datasets_synthetic.jl
davidbp: done

src/datasets_synthetic.jl
davidbp: Done

src/datasets_synthetic.jl
davidbp: The idea here was to allow cluster_std as a real or as a vector of reals . If it is a single real it is used as a scaling factor of the spreads of the cluster_stds. If it is an array of floats then the stds are used form the array. Maybe writting: `Union{Real,Array{Real}}` would be better? or we could simply generate the `stds` without allowing users to pass them (I just translated the methods from sklearn and wanted to allow a very similar API).

src/datasets_synthetic.jl
davidbp: For efficiency I usually use LinRange but the numbers are scale is so tiny (and the amount of times this will called) that is pretty irrelevant.

```julia> @btime collect(range(0, 2pi, length=10))
  209.051 ns (1 allocation: 160 bytes)
10-element Array{Float64,1}:
 0.0               
 0.6981317007977318
 1.3962634015954636
 2.0943951023931953
 2.792526803190927 
 3.490658503988659 
 4.1887902047863905
 4.886921905584122 
 5.585053606381854 
 6.283185307179586 

julia> @btime Array(LinRange(0, 2 * pi, 10))
  31.875 ns (1 allocation: 160 bytes)
10-element Array{Float64,1}:
 0.0               
 0.6981317007977318
 1.3962634015954636
 2.0943951023931953
 2.792526803190927 
 3.490658503988659 
 4.1887902047863905
 4.886921905584122 
 5.585053606381854 
 6.283185307179586 
```

Whatever you decide!

src/datasets_synthetic.jl
davidbp: Would you type all the input aguments?

src/datasets_synthetic.jl
davidbp: done!

src/datasets_synthetic.jl
davidbp: Done

src/datasets_synthetic.jl
davidbp: Done.

You are right, besides it is much faster

```
julia> @btime [x1...,x2...]
  280.772 μs (20003 allocations: 625.20 KiB)
```

```
julia> @btime hcat(x1,x2)
  18.690 μs (5 allocations: 156.42 KiB)
```

src/datasets_synthetic.jl
davidbp: Done

src/datasets_synthetic.jl
davidbp: Done

src/datasets_synthetic.jl
tlienart: sorry, did you forget to  push? I can't see the change :)

src/datasets_synthetic.jl
davidbp: I guess that preallocating and doing ...

```
ind_center = start_ind:(start_ind + n_per_center[i]-1)
X[ind_center,:] .= center' .+ std .* randn(element_type, (n_per_center[i], p));
y[ind_center] .= i
```
is more or less what you meant.

src/datasets_synthetic.jl
davidbp: Done,

```
## previous version with push!
  42.991 μs (209 allocations: 14.72 KiB)
## New version broadbasting with preallocated arrays
25.059 μs (113 allocations: 8.67 KiB)
```

src/datasets_synthetic.jl
davidbp: ^^'  I forgot to push, now it´s done.

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,176 @@
...
+uniform_sample_in_zero_maxval(p, maxval) = maxval .* (1 .- rand(p))
+uniform_sample_in_minval_maxval(p, minval, maxval) = (maxval-minval) .* rand(p) .+ minval
+normal_sample(p, mu, var) = mu .+ sqrt(var) .* randn(p)
+
+
+"""
+Shuffles the rows of an Array `X` and the values of a vector `y` using a randomly
+generated permutation. The same permutation is used to shuffle both `X` and `y`.
+"""
+function shuffle_Xy(X, y; random_seed=1234)
```
tlienart: put  `random_seed=Random.GLOBAL_RNG` otherwise it will always call the `1234` seed unless the user sets it explicitly.

src/datasets_synthetic.jl
tlienart: bump

src/datasets_synthetic.jl
tlienart: remove the `@assert` 

```julia
0 <= factor <=1  || throw(ArgumentError("factor should be in [0,1]"))
```

src/datasets_synthetic.jl
tlienart: + same comment with  random_seed

src/datasets_synthetic.jl
```diff
@@ -0,0 +1,176 @@
...
+
+Generates `n` examples sampling from two moons. The `noise` can be changed to add
+noise to the samples.
+"""
+function make_moons(n::Int=100; shuffle=true, noise=0.,
+                   translation=0.5, factor=1.0, random_seed=1234)
+
+    @assert 0 <= factor <=1  || throw(ArgumentError("factor should be in [0,1]"))
+    @assert 0 <= noise  || throw(ArgumentError("noise should be in [0,inf)"))
+
```
tlienart: remove assert

davidbp:
Hopefully all issues but the test part are done.

I am not sure hot to test for the error cases. I´ve added abstract types as input to avoid users to break the methods. 

I still have the issue with `n` or `p` if introduced as negative but I don't want to write `n::Unsigned` because it is quite annoying. Now, as `n::Int` a user can write `n=-1` and this should not be allowed.

tlienart:
Thanks for all your effort david, I'll finish the last minor things & merge it

davidbp:
Last try before 2020, @tlienart go and rest a little bit :D

tlienart:
superseded by #155

PR #148, opened 2019-12-19
Reduce risk of numerical issues with cross entropy and additional tests

tlienart:
* additional tests comparing with  sklearn `brier_score_loss` and `log_loss`
* clamping of `pdf(d, y)` in `_cross_entropy` to `[eps(), 1-eps()]` to avoid numerical issues
* get rid of stray  `Any[ŷ...]` which are not necessary as far as I can tell
* allow passing a clamping epsilon to cross entropy (most likely this won't be used but it matches what sklearn  provides)

### DIFF
--- a/src/measures/finite.jl
+++ b/src/measures/finite.jl
@@ -12,8 +12,9 @@ Given an abstract vector of distributions `ŷ` and an abstract vector
 of true observations `y`, return the corresponding Cross-Entropy
 loss (aka log loss) scores.
 
-Since the loss is undefined for a score of `0` or `1`, probabilities
-are clipped between `eps` and `1-eps` where `eps` can be specified.
+Since the score is undefined in the case of the true observation has
+predicted probability zero, probablities are clipped between `eps` and `1-eps`
+where `eps` can be specified.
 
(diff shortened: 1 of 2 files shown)


src/measures/finite.jl
```diff
@@ -4,7 +4,29 @@
...
+    ce = CrossEntropy(; eps=eps())
+    ce(ŷ, y)
+
+Given an abstract vector of distributions `ŷ` and an abstract vector
+of true observations `y`, return the corresponding Cross-Entropy
+loss (aka log loss) scores.
+
+Since the loss is undefined for a score of `0` or `1`, probabilities
+are clipped between `eps` and `1-eps` where `eps` can be specified.
+
```
ablaom: I agree it makes sense to clamp at both ends (like Sk learn?). However, the statement "Since the loss is undefined for a score of `0` or `1`," is not completely completely accurate. I can take the log of 1 just fine. The problem is taking the log of 0. 

Perhaps say "Since the score is undefined in the case of the true observation has predicted probability zero, probablities are clipped ..."

Perg

tlienart: Sounds good, I think  I took the undefined note from sklearn's docs. will fix

PR #166, opened 2020-01-28
Light interface

tlienart:
Ok this now works fully locally with Julia 1.3.1 

This can only work after

- [x] ScientificTypes 0.6 has been released [this  PR](https://github.com/alan-turing-institute/ScientificTypes.jl/pull/84)
- [x] MLJScientificTypes 0.1 has been released [this PR](https://github.com/alan-turing-institute/MLJScientificTypes.jl/pull/1)
- [x] MLJModelInterface 0.1 has been  released [this PR](https://github.com/alan-turing-institute/MLJModelInterface.jl/pull/1)


src/MLJBase.jl
```diff
@@ -142,19 +143,24 @@ export trait, Scientific, Found, Unknown, Finite, Infinite,
...
 
 import Base: ==, precision, getindex, setindex!, @__doc__
 
-using Tables
-using PrettyTables
+# Scitype
+using MLJModelInterface
+using MLJScientificTypes
+using ScientificTypes
+
```
ablaom: What's the rationale for importing ScientificTypes? Does MLJScientificTypes not re-export all of ScientificTypes and more?

src/MLJBase.jl
```diff
@@ -166,14 +172,18 @@ import StatsBase: fit, predict, fit!, mode, countmap
...
-
+# ===================================================================
 ## CONSTANTS
 
+# aliases
+const MST = MLJScientificTypes
+const MMI = MLJModelInterface
+const ST  = ScientificTypes
+
 # the directory containing this file:
```
ablaom: Frankly, I don't like all this aliasing. What's wrong with just spelling it out? It makes it much easier to inpect, review and maintain.

ablaom:
Minor point: I'm getting:

```julia
WARNING: replacing module Models.
```

Is it possible  your new `include_everywhere` function is double loading in case of nprocs=1?

ablaom:
Finicky point: Because MLJBase does not own the UnivariateFinite *method*, only it's constructor, so to speak, the documentation string for UnivariateFinite that currently resides in MLJModelInterface is not accessible with `using MLJBase; ?UnivariateFinite`. (This does not apply to other methods, like `int`.) I found this a little annoying, although I realise we cannot have it both ways without duplicating the docstring. Oh, hang on, we could duplicate it. Define it as a const in MLJModelInterface and just reproduce it as a docstring for the Univariate *function* definition in MLJBase?

ablaom:
Why do you pool resampling.jl in the "hyperparam" folder? It does seem to be related to hyperparameters, as such?

ablaom:
Okay, I've looked over this as much as I think is useful for now. I appreciate the huge effort going into this PR and the other supporting ones. Many thanks ❤️ 

If you could address the minors point above, great. Otherwise I'm happy for you to merge and tag a minor release.

ablaom:
Oops, left a task reference in my "minor" commits above. Will fix this now.

PR #177, opened 2020-02-12
OpenML

darenasc:
Integration with OpenML API. It is possible to pull datasets from [OpenML.org]((https://www.openml.org)) from `MLJBase`.

For example, the iris dataset has the ID 61 in [OpenML.org](https://www.openml.org/d/61) (`https://www.openml.org/d/61`). The `OpenML.load(61)` function will return a `Vector` of `NamedTuples` that can be transformed with a DataFrame or other tabular formats.

```
ntp = OpenML.load(61)
df = DataFrame(ntp)
```


tlienart:
did you mean to close this?

darenasc:
yes, I did a new PR with a few changed (removed) lines.

PR #181, opened 2020-02-12
OpenML API

darenasc:
OpenML.org API integration, the `OpenML.load()` function will pull a dataset from the openml platform given its ID number, and return it as a `Vector` of `NamedTuples`.

For example, the iris dataset has the ID 61 in [OpenML](https://www.openml.org/d/61). The vector of named tuples can be later converted to tabular format such as a `DataFrame`.

```
named_tuples = OpenML.load(61)
df = DataFrame(named_tuples)
```


darenasc:
It is failing the tests, will check it out and create a new PR.

PR #182, opened 2020-02-12
OpenML API

darenasc:
OpenML.org API integration, the `OpenML.load()` function will download a dataset from the OpenML platform given its ID number, and return it as a `Vector` of `NamedTuples`.

For example, the iris dataset has the ID 61 in OpenML. The named tuples can be converted to tabular format such as a `DataFrame`.

```
iris_named_tuples = OpenML.load(61)
df = DataFrame(iris_named_tuples)
```

### DIFF
--- a/src/openml.jl
+++ b/src/openml.jl
@@ -3,7 +3,7 @@ module OpenML
 using HTTP
 using JSON
 
-api_url = "https://www.openml.org/api/v1/json"
+const API_URL = "https://www.openml.org/api/v1/json"
 
 # Data API
 # The structures are based on these descriptions
@@ -24,7 +24,7 @@ information about the creator, URL to download it and more.
 112 - No access granted. This dataset is not shared with you.
 """
 function load_Dataset_Description(id::Int; api_key::String="")
-    url = string(api_url, "/data/$id")
+    url = string(API_URL, "/data/$id")
     try
         r = HTTP.request("GET", url)
         if r.status == 200
@@ -37,8 +37,10 @@ function load_Dataset_Description(id::Int; api_key::String="")
             println("No access granted. This dataset is not shared with you.")
         end
     catch e
-        return "Error occurred : $e"
+        println("Error occurred : $e")
+        return nothing
     end
+    return nothing
 end
 
 """
@@ -63,12 +65,12 @@ function convert_ARFF_to_rowtable(response)
                 elseif occursin("@relation", lowercase(line))
                     nothing
                 elseif occursin("@data", lowercase(line))
-                    # it means the data starts so we can create the data frame
+                    # it means the data starts
                     nothing
                 else
                     values = split(line, ",")
-                    for i = 1:length(featureNames)
-                        if in(lowercase(dataTypes[i]), ["real","numeric"])
+                    for i in eachindex(featureNames)
+                        if lowercase(dataTypes[i]) in ["real","numeric"]
                             push!(d, featureNames[i] => Meta.parse(values[i]))
                         else
                             # all the rest will be considered as String
@@ -102,7 +104,7 @@ Returns a list of all data qualities in the system.
 370 - No data qualities available. There are no data qualities in the system.
 """
 function load_Data_Qualities_List()
-    url = string(api_url, "/data/qualities/list")
+    url = string(API_URL, "/data/qualities/list")
     try
         r = HTTP.request("GET", url)
         if r.status == 200
@@ -111,8 +113,10 @@ function load_Data_Qualities_List()
             println("No data qualities available. There are no data qualities in the system.")
         end
     catch e
-        return "Error occurred : $e"
+        println("Error occurred : $e")
+        return nothing
     end
+    return nothing
 end
 
 """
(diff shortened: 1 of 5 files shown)


src/openml.jl
```diff
@@ -0,0 +1,230 @@
+module OpenML
+
+using HTTP
+using JSON
+
+api_url = "https://www.openml.org/api/v1/json"
```
tlienart: change for

```julia
const API_URL ...
```

src/openml.jl
```diff
@@ -0,0 +1,230 @@
...
+Returns information about a dataset. The information includes the name,
+information about the creator, URL to download it and more.
+
+110 - Please provide data_id.
+111 - Unknown dataset. Data set description with data_id was not found in the database.
+112 - No access granted. This dataset is not shared with you.
+"""
+function load_Dataset_Description(id::Int; api_key::String="")
+    url = string(api_url, "/data/$id")
+    try
```
tlienart: Here you could add a request to Julia's discourse, if it fails you can set a flag to OFFLINE and show an  error message that the user seems offline.

src/openml.jl
```diff
@@ -0,0 +1,230 @@
...
+            println("Please provide data_id.")
+        elseif r.status == 111
+            println("Unknown dataset. Data set description with data_id was not found in the database.")
+        elseif r.status == 112
+            println("No access granted. This dataset is not shared with you.")
+        end
+    catch e
+        return "Error occurred : $e"
+    end
+end
```
tlienart: the end should look like:

```julia
catch e
  @warn "Error occurred : $e"
  return nothing
end
return nothing # for all the  cases where you do  `println`
```

also replace the  `println` with `@warn`

src/openml.jl
```diff
@@ -0,0 +1,230 @@
...
+function convert_ARFF_to_rowtable(response)
+    data = String(response.body)
+    data2 = split(data, "\n")
+
+    featureNames = String[]
+    dataTypes = String[]
+    named_tuples = []
+    for line in data2
+        if length(line) > 0
+            if line[1:1] != "%"
```
tlienart: this is weird, `line[1] != '%'` is better (note single quote)

src/openml.jl
```diff
@@ -0,0 +1,230 @@
...
+                    push!(featureNames, replace(split(line, " ")[2], "'" => ""))
+                    push!(dataTypes, split(line, " ")[3])
+                elseif occursin("@relation", lowercase(line))
+                    nothing
+                elseif occursin("@data", lowercase(line))
+                    # it means the data starts so we can create the data frame
+                    nothing
+                else
+                    values = split(line, ",")
+                    for i = 1:length(featureNames)
```
tlienart: can use `for  i in eachindex(featureNames)` but doesn't matter

src/openml.jl
```diff
@@ -0,0 +1,230 @@
...
+                        if in(lowercase(dataTypes[i]), ["real","numeric"])
```
tlienart: prefer the operation syntax `lowercase(...) in [...]` as it's more readable

src/openml.jl
```diff
@@ -0,0 +1,230 @@
...
+370 - No data qualities available. There are no data qualities in the system.
+"""
+function load_Data_Qualities_List()
+    url = string(api_url, "/data/qualities/list")
+    try
+        r = HTTP.request("GET", url)
+        if r.status == 200
+            return JSON.parse(String(r.body))
+        elseif r.status == 370
+            println("No data qualities available. There are no data qualities in the system.")
```
tlienart: again `@warn` and return something after the try catch

src/openml.jl
```diff
@@ -0,0 +1,230 @@
...
+    elseif r.status == 271
+        println("Unknown dataset. Data set with the given data ID was not found (or is not shared with you).")
+    elseif r.status == 272
+        println("No features found. The dataset did not contain any features, or we could not extract them.")
+    elseif r.status == 273
+        println("Dataset not processed yet. The dataset was not processed yet, features are not yet available. Please wait for a few minutes.")
+    elseif r.status == 274
+        println("Dataset processed with error. The feature extractor has run into an error while processing the dataset. Please check whether it is a valid supported file. If so, please contact the API admins.")
+    end
+
```
tlienart: same here

src/openml.jl
```diff
@@ -0,0 +1,230 @@
...
+        println("Unknown dataset. The data set with the given ID was not found in the database, or is not shared with you.")
+    elseif r.status == 362
+        println("No qualities found. The registered dataset did not contain any calculated qualities.")
+    elseif r.status == 363
+        println("Dataset not processed yet. The dataset was not processed yet, no qualities are available. Please wait for a few minutes.")
+    elseif r.status == 364
+        println("Dataset processed with error. The quality calculator has run into an error while processing the dataset. Please check whether it is a valid supported file. If so, contact the support team.")
+    elseif r.status == 365
+        println("Interval start or end illegal. There was a problem with the interval start or end.")
+    end
```
tlienart: and  here

src/openml.jl
```diff
@@ -0,0 +1,230 @@
...
+        return JSON.parse(String(r.body))
+    elseif r.status == 370
+        println("Illegal filter specified.")
+    elseif r.status == 371
+        println("Filter values/ranges not properly specified.")
+    elseif r.status == 372
+        println("No results. There where no matches for the given constraints.")
+    elseif r.status == 373
+        println("Can not specify an offset without a limit.")
+    end
```
tlienart: and here

test/openml.jl
```diff
@@ -0,0 +1,28 @@
+using Test
```
tlienart: did you add those tests in runtests.jl?

src/openml.jl
darenasc: done.

src/openml.jl
darenasc: modified.

src/openml.jl
darenasc: yeah, looks more readable, changed.

test/openml.jl
darenasc: thanks, it is added now.

src/openml.jl
darenasc: sorry, it didn't work, I kept it with double quote.

ablaom:
@tlienart Thanks for reviewing!

ablaom:
@darenasc Looks like there's just the one suggestion of @tlienart about testing the network is online. Did you want to do something about this?

darenasc:
Thanks @ablaom, I'll leave that one for an update @tlienart, we don't want to generate a DoS of the discourse page from MLJ :) 

There are error messages are printed for all the statuses and that should work for now.

src/openml.jl
darenasc: Thanks @tlienart I think something around that would be good, maybe using a dedicated service for that or even an internal feature in the HTTP library that can check it. For this PR I think it is ready to go, and I'll see what would be a good solution for that checking and include it in an update soon. There are more functions using the API key that I'd like to work next.

Issue #169, opened 2020-02-05
Error with evaluate! tests in the distributed setting


tlienart:
I've not looked into this in details but the tests in  `resampling.jl` cause problems as far as I can tell. Specifically tests following `evaluate!`.

It may be that it's actually a bug with `evaluate!`  and not a problem with the tests per se but either way there are only issues with this in `resampling` when using multiple workers (on CI the tests pass). 

Here's an example of test that sometimes fail:

```julia
# cv:
x1 = ones(4)
x2 = ones(4)
X = (x1=x1, x2=x2)
y = [1.0, 2.0, 3.0, 1.0]
w = 1:4
cv=CV(nfolds=2)
model = Models.DeterministicConstantRegressor()
mach = machine(model, X, y)
e = evaluate!(mach, resampling=cv, measure=l1,
              weights=w, verbosity=0, acceleration=accel).measurement[1]

@test e ≈ (1/3 + 13/14)/2 #  THIS DOES NOT  ALWAYS PASS
```

**I've commented out all those tests in #166 and marked them with a `XXX`** as they sometimes fail but are unrelated to the changes I made (these tests sometimes fail locally on `master` branch as well). They should be fixed. (PS: I've set the seed used by all workers to be the same but this does not fix the issue).

I suggest we first get the light-interface in then fix this in order to not get side tracked.

ablaom:
Okay this is wierd. There is no randomness in the CV(nfolds=2) resampling strategy, and certainly none in `DeterministicConstantRegressor`. 

I assume here that `accel=CPUProcesses` and you have nprocs() == 3 or something?

ablaom:
Can you say how badly this fails? That is, what is the size of the error?

tlienart:
Full stacktrace:

```

Evaluating over 5 folds: 100%[=========================] Time: 0:00:00
folds specified (accelerated with CPUThreads): Test Failed at /Users/tlienart/.julia/dev/MLJBase/test/hyperparam/resampling.jl:81
  Expression: result.per_fold[1] ≈ v
   Evaluated: [0.25, 0.75, 0.25, 0.75, 0.5] ≈ [0.5, 0.75, 0.5, 0.75, 0.5]
Stacktrace:
 [1] top-level scope at /Users/tlienart/.julia/dev/MLJBase/test/hyperparam/resampling.jl:81
[...]

PR #201, opened 2020-02-19
Fix resampling race condition, and so forth

ablaom:
- [x] (**bug**) Remove multithreading race conditions in resampling (use of `evaluate!`) (#169)

- [x] (**testing**) Fix and re-instate all `CPUThreads` and `CPUProcesses` tests in resampling

- [x] (**enhancement**) Add progress meter to all forms of accelerated resampling

- [x] (**testing**) Increase number of workers in travis testing to 2 (3 processes); increase number of threads to 2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@
 language: julia
 os:
   - linux
+env:
+  - JULIA_NUM_THREADS=30
 julia:
   - 1.1
   - 1.2
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -28,9 +28,7 @@ using Distributed
 using ComputationalResources
 using ComputationalResources: CPUProcesses
 using ProgressMeter
-@static if VERSION >= v"1.3.0-DEV.573"
-    import .Threads
-end
+import .Threads
 
 # Operations & extensions
 import LossFunctions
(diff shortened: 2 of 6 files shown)


ablaom:
Five times speed up on my laptop using multithreading. Yay!

```julia
julia> using MLJ, BenchmarkTools
julia> X, y = @load_boston;
julia> model = @load DecisionTreeRegressor;
julia> mach = machine(model, X, y):

julia> @btime evaluate!(mach, resampling=CV(nfolds=12), 
                              verbosity=0)
  15.821 ms (8557 allocations: 2.23 MiB)

julia> using .Threads; nthreads()
12

julia> import ComputationalResources.CPUThreads

julia> @btime evaluate!(mach, resampling=CV(nfolds=12), 
                              verbosity=0,
                              acceleration=CPUThreads())
  3.460 ms (8151 allocations: 2.13 MiB)
```

ablaom:
BTW, travis provides some limited possibilities for testing parallelization:

Screen Shot 2020-02-19 at 9 11 46 PM

ablaom:
Still a few wrinkles to iron out in this PR, which I will take up tomorrow.

ablaom:
@jpsamaroo Perhaps if you have time, you could have a look over this PR.

ablaom:
@OkonSamuel Many thanks for your review. You make a good point (and it sounds like you have more expertise than I in this area). The only real requirement is that two threads never touch the same machine, and that each thread will need a "clone" of the specified machine to work with. 

In next commit I use @spawn and clone the machines as needed, putting adding them to a dictionary keyed on threadid. Testing locally with nthreads=4 (nrprocs =3). Performance is about the same.

PR #203, opened 2020-02-24
WIP: Unnormalize weighted loss functions

juliohm:
Fix https://github.com/alan-turing-institute/MLJ.jl/issues/445

I am not familiar with the resampling API, appreciate if you can fix the tests.

After this is merged, I will prepare a second PR in an attempt to reorganise all these traits and measure implementations.

### DIFF
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -165,7 +165,7 @@ export orientation, reports_each_observation,
        aggregate, default_measure, value
 
 # measures/continuous.jl:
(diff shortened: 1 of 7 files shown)


juliohm:
Notice I've also cleaned the synonym `mav` used as an alternative to `mae`. Synonyms aren't very helpful, and can be quite confusing actually. The term `mae` plays well with `mse` for mean square error. In a future PR, I will work on this naming to make sure that all the acronyms follow a minimum convention: mse, mae, rmse, rmae, etc.

PR #205, opened 2020-02-24
Bump Tables.jl version in Project.toml

juliohm:
Could you please update to Tables.jl 1.0? The API didn't change AFAICT, it just settled on a stable version. I've also removed the carrot symbol as it is the default in Julia v1.x


tlienart:
dupe of #204 this will likely be merged within 24h

PR #208, opened 2020-03-03
Suppressing weight normalization

ablaom:
(**breaking**) Remove the normalisation of weights in measures.

### DIFF
--- a/src/measures/continuous.jl
+++ b/src/measures/continuous.jl
@@ -12,8 +12,8 @@ Mean absolute error.
 
 For more information, run `info(mae)`.
 """
-mae = MAE()
-mav = MAE()
+const mae = MAE()
+const mav = MAE()
 
 metadata_measure(MAE;
     name                     = "mae",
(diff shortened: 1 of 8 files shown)


src/measures/continuous.jl
```diff
@@ -1,60 +1,51 @@
...
+metadata_measure(MAE;
+    name                     = "mae",
     target_scitype           = Union{Vec{Continuous},Vec{Count}},
     prediction_type          = :deterministic,
     orientation              = :loss,
     reports_each_observation = false,
     is_feature_dependent     = false,
     supports_weights         = true,
-    docstring                = "mean absolute value; aliases: `mav`.")
+    docstring                = "mean absolute error; aliases: `mae`, `mav`")
```
tlienart: `mav` should go away?

ablaom: Good catch. I put the `mav` alias in the docstring but forgot to put the definition back. I guess `mav` isn't used that much, but I it is currently in one of the tutorials (A-ensembles-3).

ablaom:
Merged onto `early-march` branch instead, in preparation for next minor release.

PR #218, opened 2020-03-13
Add serialization

ablaom:
This PR resolves https://github.com/alan-turing-institute/MLJ.jl/issues/138. It:

 - adds fallback methods `save` and `restore` dispatched on  *models* (the method stubs are already provided by MLJModelInterface) which save using JSLO. 

- defines `MLJBase.save` for machines and extends the `machine` constructor for de-serialization, as demonstrated in the docstring below:

- refactors some of the `machine` constructor code to simplify, improve readability, and resolve some type ambiguities in corner cases

**edited:** the docstring

---
    MLJ.save(filename, mach::Machine; kwargs...)
    MLJ.save(io, mach::Machine; kwargs...)

    MLJBase.save(filename, mach::Machine; kwargs...)
    MLJBase.save(io, mach::Machine; kwargs...)

Serialize the machine `mach` to a file with path `filename`, or to an
input/output stream `io` (at least `IOBuffer` instances are
supported). 

The format is JLSO (a wrapper for julia native or BSON serialization)
unless a custom format has been implemented for the model type of
`mach.model`. The keyword arguments `kwargs` are passed to
the format-specific serializer, which in the JSLO case include these:

keyword        | values                        | default
---------------|-------------------------------|-------------------------
`format`       | `:julia_serialize`, `:BSON`   | `:julia_serialize`
`compression`  | `:gzip`, `:none`              | `:none`

See (see
[https://github.com/invenia/JLSO.jl](https://github.com/invenia/JLSO.jl)
for details.

Machines are de-serialized using the `machine` constructor as shown in
the example below. Data may be optionally passed to the constructor
for retraining on new data using the saved model.

### Example

    using MLJ
    tree = @load DecisionTreeClassifier
    X, y = @load_iris
    mach = fit!(machine(tree, X, y))

    MLJ.save("tree.jlso", mach, compression=:none)
    mach_predict_only = machine("tree.jlso")
    predict(mach_predict_only, X)

    mach2 = machine("tree.jlso", selectrows(X, 1:100), y[1:100])
    predict(mach2, X) # same as above

    fit!(mach2) # saved learned parameters are over-written
    predict(mach2, X) # not same as above

    # using a buffer:
    io = IOBuffer()
    MLJ.save(io, mach)
    seekstart(io)
    predict_only_mach = machine(io)
    predict(predict_only_mach, X)


ablaom:
Feedback welcome.

@tlienart Only if you have time

@OkonSamuel Would be great to get your feedback

ablaom:
cc @oxinabox

src/machines.jl
```diff
@@ -45,46 +46,54 @@ function machine(model::Supervised, args...)
...
+    return nothing
 end
 
-# public constructors for unsupervised models
-function machine(model::Unsupervised, args...)
-    # check arguments
+function check(model::Unsupervised, args...)
     nargs = length(args)
-    length(args) == 1 ||
+    nargs in [0,1] ||
```
tlienart: nargs <= 1

src/machines.jl
```diff
@@ -45,46 +46,54 @@ function machine(model::Supervised, args...)
...
+function machine(model::Model, args...)
+    isempty(args) &&
+        error("`machine(model)` is ambiguous. Use `Machine(model)` or "*
+              "`NodalMachine(model)` (for machines in a learning network). ")
+    check(model, args...)
     return Machine(model, args...)
 end
 
-machine(model::Static, args...) = Machine(model, args...)
+# machine(model::Static, args...) = Machine(model, args...)
```
tlienart: why is this commented out?

src/machines.jl
ablaom: :smile

src/machines.jl
ablaom: The `machine(model::Unsupervised, args...)` constructor now suffices (`Static <: Unsupervised`)

OkonSamuel:
@ablaom. This is great.
 But you haven't yet exported the `save` function.

src/machines.jl
```diff
@@ -200,3 +217,79 @@ is_stale(mach::Machine) =
...
+
+    # using a buffer:
+    io = IOBuffer()
+    MLJ.save(io, mach)
+    seekstart(io)
+    predict_only_mach = machine(io)
+    predict(predict_only_mach, X)
+
+"""
+function MMI.save(file, mach::Machine; verbosity=1, kwargs...)
```
OkonSamuel: The `MMI.save` function now has two methods which will be both be exported. Except you wish for both methods to be exported maybe this should just be named `save`.

src/MLJBase.jl
```diff
@@ -13,7 +13,8 @@ using MLJModelInterface
 import MLJModelInterface: fit, update, update_data, transform,
                           inverse_transform, fitted_params, predict,
                           predict_mode, predict_mean, predict_median,
-                          evaluate, clean!, is_same_except
+                          evaluate, clean!, is_same_except,
+                          save, restore
```
OkonSamuel: No need to specify `save` since it are already specified in the code as `MMI.save`. This is to prevent ambiguity in export. See my other comment below

src/machines.jl
```diff
@@ -200,3 +217,79 @@ is_stale(mach::Machine) =
...
+    fit!(mach2) # saved learned parameters are over-written
+    predict(mach2, X) # not same as above
+
+    # using a buffer:
+    io = IOBuffer()
+    MLJ.save(io, mach)
+    seekstart(io)
+    predict_only_mach = machine(io)
+    predict(predict_only_mach, X)
+
```
oxinabox:
```suggestion
!!! warning "Only load files from trusted sources"
    Maliciously constructed JLSO files, like pickles, and most other general purpose
    serialization formats can allow for arbitrary code execution during loading.
    This means it is possible for someone to use a JLSO file that looks like a serialized
    MLJ machine as a [Trojan horse](https://en.wikipedia.org/wiki/Trojan_horse_(computing)).
```

ablaom: Perfect. Will add.

src/MLJBase.jl
ablaom: Many thanks for the review!

First let me clarify that I am disinclined to *export* `save` (unqualified) because I feel it's likely to come into conflict with the packages MLJ users might want to use. What do you think?

I'm not sure I follow your comments. Yes, I can extend `MMI.save` with the explicit import, but it then it will not available to an `MLJBase` user:

```julia
julia> module FooBar
       import MLJModelInterface
       MLJModelInterface.save() = print("save extened")
       end

julia> using .FooBar

julia> FooBar.save()
ERROR: UndefVarError: save not defined
Stacktrace:
 [1] getproperty(::Module, ::Symbol) at ./Base.jl:13
 [2] top-level scope at REPL[5]:1
```

My intention is to exend the `MLJModelInterface.save` stub with two methods, and for these to both be available to `MLJBase` users using `using MLJBase; MLJBase.save` (but `using MLJBase; save` should not work). This is what my implementation does, as far as I can tell:

```julia
julia> using MLJBase
[ Info: Precompiling MLJBase [a7f614a8-145f-11e9-1d2a-a57a1082229d]

help?> save
search: StackOverflowError AbstractVector AbstractVecOrMat bytesavailable

Couldn't find save
Perhaps you meant scale, size, Base, mae, mav, name, isa, stat, std, Some or sec
  No documentation found.

  Binding save does not exist.

help?> MLJBase.save
  MLJ.save(filename, mach::Machine; kwargs...)
  MLJ.save(io, mach::Machine; kwargs...)
  
  MLJBase.save(filename, mach::Machine; kwargs...)
  MLJBase.save(io, mach::Machine; kwargs...)

  Serialize the machine mach to a file with path filename, or to an
  input/output stream io (at least IO...
```

Possibly I misunderstand your comment?

OkonSamuel: It's  okay. Since you have clarified that you intended both methods to be available to the user

ablaom:
After reflection. Have reverted `machine(m::Unsupervised) = AbstractMachine(m)` as there is really no difference in behaviour of `Machine` and `AbstractMachine` in the case of no training argments ("static" transformers). Added more tests to check this in practice.

PR #223, opened 2020-03-19
Documentation update for github pages

Arkoniak:
This is some initial setup for MLJBase API docs.

Notes:
- `matrix` has been changed to `jobs` per this recommendation: https://juliadocs.github.io/Documenter.jl/stable/man/hosting/index.html#Travis-CI-1
- There are couple of warnings in docs build section, honestly do not know how to solve them, but it seems that final docs are fine
- One should of course setup DOCUMENTER_KEY and all other things in repo and travis CI. Also I do not know at what point travis creates `stable` version in `gh-pages`, I think simpliest solution is just to add copy of `dev` folder manually in `gh-pages` branch. 
- Unfortunately I do not know intrinsic details of MLJBase, so some docs were rather loosely put to categories, hope that's ok.

Result can be found here: https://arkoniak.github.io/MLJBase.jl/dev/

PR #224, opened 2020-03-19
Auto docs

tlienart:
Thanks to #223


ablaom:
@tlienart @Arkoniak FYI, the documentation failure has been fixed by switch from julia 1.3 to julia 1.2. I do not know why that fixed the problem.

Issue #76, opened 2019-10-30
info/info_dict  bottleneck


tlienart:
In view of the investigation https://github.com/alan-turing-institute/MLJModels.jl/issues/112 it seems that these functions are much slower than they should be; this warrants an investigation.

I haven't looked in depth but a first thing to note is that calling these functions seems to  allocate like crazy; which  seems weird.

Anyway proper benchmarking of `info_dict` and `info` and fixes would probably be a good idea

Arkoniak:
I doubt it is good first issue :-))

Some investigation

```julia
using MLJBase
using BenchmarkTools

struct DummyProb <: Probabilistic end

@btime info_dict($DummyProb)
#  84.096 ms (171929 allocations: 10.68 MiB)
```

After diiging through `info_dict` it turns out that most time expensive method is `implemented_methods` which in turn expensive because it uses `InteractiveUtils` `methodswith`

```julia
@btime methodswith($DummyProb)
#   83.581 ms (171701 allocations: 10.67 MiB)
```

So, unless `methodswith` can be avoided, there is no way to make this function faster.

tlienart:
That’s a relative newcomer I think (?) anyway 80ms for that is atrocious; @ablaom do you reckon we could get around this? What’s the use case for having methodswith in the infos? 

Also I wonder whether we could do a lazy info dict where (expensive) things are only computed upon explicit querying

Arkoniak:
For future references, it is rather easy to implement lazy dict with `DefaultDict` from `DataStructures.jl`

instead of
```julia
    return LittleDict{Symbol, Any}(trait => getfield(MLJBase, trait)(M)
                                    for trait in MLJModelInterface.MODEL_TRAITS)
```

use
```julia
dd = DefaultDict{Symbol, Any}(passkey=true) do key
         getfield(MLJBase, key)(DummyProb)
     end

return dd
```

The drawback of this approach is that `keys` and `iterate` wouldn't work correctly and you need to know all keys beforehand. Since `info_dict` is internal structure, may be it's ok.

Alternatively, one can implement all `DefaultDict` functionality plus some defaults for keys. But this requires some work to carefully forward all necessary methods (only sketch of the idea)

```julia
struct LazyDict{K, V, M} <: AbstractDict{K, V}
    d::Dict{K, V}
    keys::Vector{K}
    m::M
end

Base.:keys(d::LazyDict) = union(keys(d.d), d.keys)

[...]

PR #227, opened 2020-03-24
lazy version of info_dict

Arkoniak:
This is "lazy" implementation of `info_dict` as discussed in #76

1. `OrderedCollections` was changed to `DataStructures`, fortunately, they reexport all necessary structures, so one can safely swap one to another.
2. It looks like `info_dict` is passing all tests, but not used internally. In original issue was also request regarding proper benchmarking. Is there a relevant piece of code that can be benchmarked? It is interesting to see the effect of this change.


tlienart:
great thanks! I'm not so keen for the swap orderedcollections <> datastructures as the latter is 4-5x slower to load (0.055 vs 0.245 on my machine), datastructures loads probably too  many things. 

Would it be annoying / difficult to pull out the  relevant part of the code and just have it here?

Arkoniak:
Removed `DataStructures` dependency. It seems that `info_dict` is needed in read-only mode, so implemented the bare minimum of the functions. All tests are green, but it should be checked in the packages where `info_dict` is used, maybe something is missed.

On a side note, `LittleDict{Symbol, Any}` is very bad for type stability. Maybe it would make sense to verify it's usage and change it to structure with well-defined fields.

There are multiple ways to implement lazy structure I think, mainly through metaprogramming. But it's possible in purely functional way, yet slow. Here is the prototype

```julia
mutable struct Goo
    x::Int
    y::String
    activated::Dict{Symbol, Bool}

    function Goo()
        res = new()
        res.activated = Dict(:x => false, :y => false)
        return res
    end
end

function Base.:getproperty(g::Goo, x::Symbol)
    getfield(g, :activated)[x] && return getfield(g, x)
    if x == :x
        setfield!(g, x, rand(1:1000))
    elseif x == :y
        setfield!(g, x, join(rand('a':'z') for _ in 1:4))
    end

    getfield(g, :activated)[x] = true
    return getfield(g, x)
end

function addx(g1::Goo, g2::Goo)
    return g1.x + g2.x
end

function addy(g1::Goo, g2::Goo)
    return g1.y * g2.y
end

g1 = Goo()
g2 = Goo()

@code_warntype addx(g1, g2)

Variables
  #self#::Core.Compiler.Const(addx, false)
  g1::Goo
  g2::Goo

Body::Int64
1 ─ %1 = Base.getproperty(g1, :x)::Int64
│   %2 = Base.getproperty(g2, :x)::Int64
│   %3 = (%1 + %2)::Int64
└──      return %3
```

tlienart:
Fantastic thanks, I think it would be good to have @ablaom ‘s thoughts on this but otherwise looks good to me

Arkoniak:
Haha, thank you for new idiom :-) No, my bubble is fine :-) Lazy dictionaries and lazy structures is a useful thing by itself, so it was really nice to play with these things.

As for `MLJ` and `MLJModels`.

`MLJ` - all tests are green

`MLJModels` - all tests are green, except `NaiveBayes` with 2 broken tests, which I think unrelated.

A couple of notes.
1. I have found https://github.com/alan-turing-institute/MLJModels.jl/blob/master/src/registry/src/Registry.jl#L93 with `meta_given_package` used later in `Manifest.toml`. So to be on a safe side I added `iterate` function to `LazyInfoDict`.
2. MLJModels uses `info_dict` in very strange way:
https://github.com/alan-turing-institute/MLJModels.jl/blob/master/src/model_search.jl#L52 - info_dict is cast to `Tuple` structure and not used by itself any longer. I think it kind of annulate the usefulness of the lazy approach. Maybe it would make sense to remove this intermediate structure?

ablaom:
> Haha, thank you for new idiom :-) No, my bubble is fine :-) Lazy dictionaries and lazy structures is a useful thing by itself, so it was really nice to play with these things.
> 
> As for `MLJ` and `MLJModels`.
> 
> `MLJ` - all tests are green
> 
> `MLJModels` - all tests are green, except `NaiveBayes` with 2 broken tests, which I think unrelated.
> 
> A couple of notes.
> 
> 1. I have found https://github.com/alan-turing-institute/MLJModels.jl/blob/master/src/registry/src/Registry.jl#L93 with `meta_given_package` used later in `Manifest.toml`. So to be on a safe side I added `iterate` function to `LazyInfoDict`.

Good.

> 2. MLJModels uses `info_dict` in very strange way:
>    https://github.com/alan-turing-institute/MLJModels.jl/blob/master/src/model_search.jl#L52 - info_dict is cast to `Tuple` structure and not used by itself any longer. I think it kind of annulate the usefulness of the lazy approach. Maybe it would make sense to remove this intermediate structure?

You are right. Ultimately everything is cast into named tuples. Apologies for forgetting this detail.  A long, long time ago, in a galaxy far away, we thought dictionaries were the way to go. But I think we switched to named-tuples on the user-interaction side, because they were more convenient for users in model search workflow. Switching back to dictionaries certainly isn't going to happen. In this case, convenience trumps performance.

Given this, I suggest pulling this PR really serves no purpose and, as it adds a little complexity to the code, I suggest closing. 

@Arkoniak Any objections?

Again my apologies.

ablaom:
@Arkoniak Do you have a slack handle? On julia slack channel?

Arkoniak:
Yes, I am agreed, it makes no sense to go down this route.

For future references (if someone stumbles on this issue), it is actually possible to imitate `NamedTuple` API but it is extremely slow.

```julia
julia> Base.:getproperty(d::LazyInfoDict, key::Symbol) = key in [:d, :keys, :m] ? getfield(d, key) : d[key]

julia> dd = info_dict(DummyProb)

julia> @btime $dd.name
  132.539 ns (3 allocations: 336 bytes)
```

PR #240, opened 2020-04-13
Spread remainder evenly among folds in train_test_pairs(::CV, ...)

### LINKED ISSUE #232: Spread remainder evenly among folds in train_test_pairs
The current behavior of `train_test_pairs` is to place all extra indices in the last fold. This is fine for large datasets, but is rather uneven for small data sets:

```julia
julia> cv = CV(nfolds=10);

julia> rows = 1:99;

julia> mlj_tt_pairs = MLJBase.train_test_pairs(cv, rows);

julia> map(length ∘ last, mlj_tt_pairs)
10-element Array{Int64,1}:
  9
  9
  9
  9
  9
  9
  9
  9
  9
 18
```

Instead, the remainder `r = N mod K` could be [distributed evenly](https://stats.stackexchange.com/a/48161/225507) across the `r` first folds. Here are a couple of ways to implement that:

```julia
using Base.Iterators: partition, flatten
using MLJBase


function my_train_test_pairs1(cv::CV, rows)
    N = length(rows)
    K = cv.nfolds

    if cv.shuffle
        rows=shuffle!(cv.rng, collect(rows))
    end

    # (floor(N/K), remainder)
    n, r = divrem(N, K)
    n > 0 || error("Inusufficient data for $K-fold cross-validation.\n"*
                   "Try reducing nfolds. ")

    m = n + 1 # number of observations in first r folds

    itr1 = partition( 1 : m*r , m)
    itr2 = partition( m*r+1 : N , n)
    test_folds = flatten((itr1, itr2))

    map(test_folds) do test_indices
        test_rows = rows[test_indices]

        train_rows = vcat(
            rows[ 1 : first(test_indices)-1 ],
            rows[ last(test_indices)+1 : end ]
        )

        (train_rows, test_rows)
    end
end


function my_train_test_pairs2(cv::CV, rows)
    N = length(rows)
    K = cv.nfolds

    if cv.shuffle
        rows=shuffle!(cv.rng, collect(rows))
    end

    # (floor(N/K), remainder)
    n, r = divrem(N, K)
    n > 0 || error("Inusufficient data for $K-fold cross-validation.\n"*
                   "Try reducing nfolds. ")

    m = n + 1
    M = m * r

    test_folds = map(1:K) do k
        if k <= r
            1 + (k-1)*m : k*m
        else
            j = k - r
            M + 1 + (j-1)*n : M + j*n
        end
    end

    map(test_folds) do test_indices
        test_rows = rows[test_indices]

        train_rows = vcat(
            rows[ 1 : first(test_indices)-1 ],
            rows[ last(test_indices)+1 : end ]
        )

        (train_rows, test_rows)
    end
end
```

Here's the new implementation in action:

```julia
julia> my_tt_pairs = my_train_test_pairs1(cv, rows);

julia> map(length ∘ last, my_tt_pairs)
10-element Array{Int64,1}:
 10
 10
 10
 10
 10
 10
 10
 10
 10
  9
```

I don't think partitioning the data is a bottleneck in ML pipelines, but if performance is a concern, `my_test_train_pairs1` is slightly slower than `my_test_train_pairs2` and `MLJBase.test_train_pairs`. Here are the benchmarks:

```julia
julia> using BenchmarkTools

julia> cv = CV(nfolds=10);

julia> rows = 1:100_009;

julia> @btime MLJBase.train_test_pairs($cv, $rows);
  445.345 μs (117 allocations: 6.87 MiB)

julia> @btime my_train_test_pairs1($cv, $rows);
  655.561 μs (234 allocations: 8.55 MiB)

julia> @btime my_train_test_pairs2($cv, $rows);
  444.728 μs (115 allocations: 6.87 MiB)
```

I think the small underperformance of `my_train_test_pairs1` could be ameliorated if Base had a specialized `Iterators.partition` method for `AbstractRange`. Currently the `eltype` of `partition(::UnitRange, ::Int)` is `Vector{Int}` rather than `UnitRange`. However, I think `my_train_test_pairs1` is conceptually more clear and more readable than `my_train_test_pairs2`.

I can put in a PR for this if there is interest.

CameronBieganek:
Fixes #232.

In order to not break the stratified CV unit test, I slightly modified the test so that the count for each level is a multiple of 3. That way the difference between the new and the old `train_test_pairs(::CV, ...)` has no effect on the unit test.

I noticed an interesting behavior for stratified CV. Again, it doesn't really matter much for large data sets, but for small data sets you can see a small anomaly. With the old `train_test_pairs`, if you take `y = categorical(collect('a'^4 * 'b'^4 * 'c'^4))` and use a 3-fold stratified CV, the classes would stack into the folds like this:

```
      c
c  c  c
      b
b  b  b
      a
a  a  a
```

where columns represent folds. With the new `train_test_pairs`, now they stack like this:

```
c
c  c  c
b
b  b  b
a
a  a  a
```

Probably the right way to stack them would be like this:

```
c  c  c
b  b  c
a  b  b
a  a  a
```

^ I think this would actually be pretty easy to implement.


CameronBieganek:
Odd that some of your merge commits are shown in the commit history. Is it because I rebased on dev at some point?

ablaom:
> ^ I think this would actually be pretty easy to implement.

Any help welcome. Could you say (in a new issue) more precisely what the algorithm would be?

ablaom:
Oops. This is breaking and should wait before merging onto dev.

CameronBieganek:
Ok. Do you mean breaking in the sense that if somebody is using a non-shuffled CV they'll get slightly different results? I suppose somebody could have a small unit test with an `evaluate!` of their model, and this could break their test.

Issue #235, opened 2020-04-08
`evaluate!` hangs instead of throwing exceptions (flawed exception handling)


ablaom:
A couple of times I have run into this problem. There is an Exception being thrown during `evaluate!` but instead of it making the light of day, the method call just hangs, because of the way we're currently using channels/tasks (even in CPU1() mode). 

Here's an example to reproduce the hanging:

```julia
using MLJ

X, y = @load_iris
model = @load ConstantClassifier

bad_loss(yhat, y) = throw(Exception)
evaluate(model, X, y, measure=bad_loss) # hangs!!!
```

Here's relevant code (from src/resampling):

```julia
    p = Progress(nfolds, ... )
    channel = RemoteChannel(()->Channel{Bool}(nfolds) , 1)

    function get_measurements(mach, k)
        < do stuff that might throw Exception >
        ...
        put!(channel, true)
    end

    @sync begin
        # printing the progress bar
        @async while take!(channel)
            verbosity < 1 || next!(p)
        end

        @async global measurements_flat =
            _evaluate!(get_measurements,
                       machines,
                       acceleration,
                       nfolds,
                       channel)
    end
```
where `_evaluate!` is this:

```julia
function _evaluate!(func, machines, ::CPU1, nfolds, channel)
    generator = (begin
                     r = func(machines[1], k)
                     put!(channel, true)
                     r
                 end for k in 1:nfolds)
    ret = reduce(vcat, generator)
    put!(channel, false)
    return ret
end
```
I should like to keep the task mechanism (b/s it is needed in the other acceleration modes) but am not sure how to handle the exceptions properly.

@OkonSamuel  @jpsamaroo

OkonSamuel:
@ablaom the hanging is probably due to the blocking effect of `take!` in 
```julia
    @sync begin
        # printing the progress bar
        @async while take!(channel)
            verbosity < 1 || next!(p)
        end

        @async global measurements_flat =
            _evaluate!(get_measurements,
                       machines,
                       acceleration,
                       nfolds,
                       channel)
    end
```
The code runs well with just one task. There is no need for the other `@async` hence the above code should be.
```julia
    @sync begin
        # printing the progress bar
        @async while take!(channel)
            verbosity < 1 || next!(p)
        end

        global measurements_flat =
            _evaluate!(get_measurements,
                       machines,
                       acceleration,
                       nfolds,
                       channel)
    end
```
this  allows the exception to be properly thrown.
[...]

PR #245, opened 2020-04-14
fixes issue 235

OkonSamuel:
see #235


ablaom:
Wonderful. I have noticed an issue, and I don't know if this is new or not. The progress bars work as expected for CPUProcesses and CPUThreads, but for CPU1 the completed progress bar is output all at once at the end of the computation, it seems. To reproduce:

```julia
using MLJ
X, y = @load_iris
model = @load DecisionTreeClassifier

# works as expected:
evaluate(model, X, y, acceleration=CPUThreads(), repeats=10000);

# sudden end-of-job progress:
evaluate(model, X, y, repeats=5000);
```

Do you have any idea why this might be so?

OkonSamuel:
@ablaom. This is due to the way the tasks are managed. One way around this is to use the `sleep` method to block the current task so that the task displaying the values operates, But this might affect performance duce to frequent display to screen.

OkonSamuel:
@ablaom i was able to work around this. Please check again to see if this fixes it.

OkonSamuel:
@ablaom . Here are the results of benchmarks ran on my system comparing  options `verbosity = 0` and `verbosity = 1` using  `X, y = @load_iris` and `model = @load ConstantClassifier` 
```julia
julia> @benchmark evaluate(model, $X, $y, measure=cross_entropy, verbosity=1)
BenchmarkTools.Trial: 
  memory estimate:  179.50 KiB
  allocs estimate:  2806
  --------------
  minimum time:     10.560 ms (0.00% GC)
  median time:      15.835 ms (0.00% GC)
  mean time:        15.749 ms (0.23% GC)
  maximum time:     26.033 ms (43.49% GC)
  --------------
  samples:          317
  evals/sample:     1
julia> @benchmark evaluate(model, $X, $y, measure=cross_entropy, verbosity=0)
BenchmarkTools.Trial: 
  memory estimate:  163.89 KiB
  allocs estimate:  2512
  --------------
  minimum time:     1.066 ms (0.00% GC)
  median time:      1.089 ms (0.00% GC)
  mean time:        1.142 ms (2.70% GC)
  maximum time:     10.837 ms (83.18% GC)
  --------------
  samples:          4359
  evals/sample:     1
.

**Comment by ablaom** [comment]
@OkonSamuel That does work for me.

It seems that there ought to be a better way to do this, but you have more expertise than I in this area and I am very grateful to have this fix. Many thanks!! 

I wonder if it is necessary to add the workaround code to CPUProcesses and CPUThreads, as they were already behaving well, as far as I could tell? Do you have a reason?

As far as the benchmarks go, I think we can live with this. If I use DecisionTree instead of ConstantClassfier, and if I make the data 50 times bigger (iris is tiny) then the benchmarks show little difference:

```julia
julia> @benchmark evaluate(model, $X, $y,
       measure=cross_entropy,
       verbosity=1)

BenchmarkTools.Trial: 
  memory estimate:  67.72 MiB
  allocs estimate:  1147217
  --------------
  minimum time:     262.030 ms (2.81% GC)
  median time:      272.789 ms (4.57% GC)
  mean time:        274.262 ms (3.92% GC)
  maximum time:     292.158 ms (4.75% GC)
  --------------
  samples:          19
  evals/sample:     1

julia> @benchmark evaluate(model, $X, $y,
       measure=cross_entropy,
       verbosity=0)
BenchmarkTools.Trial: 
  memory estimate:  67.70 MiB
  allocs estimate:  1146885
  --------------
  minimum time:     217.157 ms (2.87% GC)
  median time:      226.551 ms (5.15% GC)
  mean time:        225.528 ms (4.17% GC)
  maximum time:     240.730 ms (2.72% GC)
  --------------
  samples:          23
  evals/sample:     1
```

OkonSamuel:
@ablaom. There is a better way. That is avoiding `sleep` altogether. I'll make a commit in that regard.

ablaom:
💯 Awesome! Many thanks for this.

PR #253, opened 2020-04-19
improve progressmeter display

OkonSamuel:
After my last PR i decided to push the limits of the progress meter display. This lead me to make the following four changes.
1. limit the buffer size of the `RemoteChannel` to a maximum of 1000. This is to limit the size of memory allocated which may affect performance of the program depending on the platform it is ran on. This choice follows that used in [ProgressMeter](https://github.com/timholy/ProgressMeter.jl) codebase.
2. Added a yield() in the case of `CPU1` and `CPUThreads`. This is to ensure smooth updating of the progress bar especially in the case of `CPUThreads.
3. Added an  Info message in the case of `CPUThreads` similar to `CPUProcesses` to tell the user the number of threads being used.
4. Add fallback to `CPU1` when `nthreads==1`. This was done to smooth-en the display in that case.

### DIFF
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -549,11 +549,9 @@ evaluate(model::Supervised, args...; kwargs...) =
 function _evaluate!(func, machines, ::CPU1, nfolds, channel, verbosity)
     
     ret = mapreduce(vcat, 1:nfolds) do k
-                begin
-                     r = func(machines[1], k)
-                     verbosity < 1 || put!(channel, true) 
-                     r
-                 end
+             r = func(machines[1], k)
+             verbosity < 1 || put!(channel, true);yield() 
+             r
     	  end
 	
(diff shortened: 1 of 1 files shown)


ablaom:
Awesome! 💪

ablaom:
Oops. This should have been merged to dev first. 

I'll just update dev manually. Last time I tried to fix this with a revert and new merge I got into trouble.

ablaom:
Done.

ablaom:
Now you have perfected these, would be great if you get a chance to consider  https://github.com/alan-turing-institute/MLJTuning.jl/issues/14 and https://github.com/alan-turing-institute/MLJTuning.jl/issues/15 ? 🥺

OkonSamuel:
@ablaom  I'll take a look at it in my spare time.

PR #256, opened 2020-04-22
Adds cross-validation for time series 

aa25desh:
Cross-validation for time series by implementing  train_test_pairs and tsCV.
Working example can be found over
https://github.com/aa25desh/graph/blob/master/add_tscv_to_MLJbase.ipynb

PR #278, opened 2020-04-30
adds cross-validation for time series

aa25desh:
see  #256

Cross-validation for time series by implementing  train_test_pairs and TimeSeriesCV. 
look here, https://robjhyndman.com/hyndsight/tscv/


ablaom:
Okay. This is looking pretty good, thanks!

Thanks for your patience while I found time to review. 

Can I suggest:

- A.  You rename `nsteps` to `horizon` (slightly more informative)

- B.  You use the following revised doc-string, which assumes you go with A (but please proof read):

"""
    tscv = TimeSeriesCV(; horizon=4)

Cross-validation resampling strategy, for use in `evaluate!`,
`evaluate` and tuning, when observations are chronological and not
expected to be independent.

    train_test_pairs(tscv, rows)

Return an iterator of `(train, test)` pairs of vectors in which `train` progressively grows and `test` consists of a single row index `horizon` steps later in `rows`. Specifically, 

    train[k] = rows[1:k]
    test[k] = [rows[k + horizon], ]

for `k in 1:(length(rows) - horizon)`.

### Example

```julia
julia> tscv = TimeSeriesCV(horizon=3)
julia> MLJBase.train_test_pairs(tscv, collect(1:2:20))
7-element Array{Tuple{Array{Int64,1},Array{Int64,1}},1}:
 ([1], [7])
 ([1, 3], [9])
 ([1, 3, 5], [11])
 ([1, 3, 5, 7], [13])
 ([1, 3, 5, 7, 9], [15])
 ([1, 3, 5, 7, 9, 11], [17])
 ([1, 3, 5, 7, 9, 11, 13], [19])

```

"""

- C. You check that `sort(rows) == rows` and issue a warning otherwise (see test below for suggestion).

- D.. You add some tests, eg:

```julia
tscv = TimeSeriesCV(horizon=3)
@test MLJBase.train_test_pairs(tscv, collect(1:2:20)) == 
[ ([1], [7]),
 ([1, 3], [9]),
 ([1, 3, 5], [11]),
 ([1, 3, 5, 7], [13]),
 ([1, 3, 5, 7, 9], [15]),
 ([1, 3, 5, 7, 9, 11], [17]),
  ([1, 3, 5, 7, 9, 11, 13], [19])]

@test_logs((:warn, r"TimeSeriesCV being applied to `rows` not in sequence. "),
           MLJBase.train_test_pairs(tscv, reverse(1:10)))
```

How do these suggestions sound to you?

I have resolved the merge conflicts on [this branch](https://github.com/alan-turing-institute/MLJBase.jl/tree/time-series-cv) if you care to rebase. If you don't know how to do that, don't worry, I'll fix later.

ablaom:
cc @CameronBieganek

aa25desh:
> * C. You check that `sort(rows) == rows` and issue a warning otherwise (see test below for suggestion).
Yeah, we should check the sequence.
> How do these suggestions sound to you?
Thank you for very well informed reply, I will make changes by tomorrow & will try rebase.

@CameronBieganek sorry I did't get on rebase last time, thought I would make mess:)

CameronBieganek:
Awesome, @aa25desh! Thanks for working on this.

I can see how this would be useful when working with an ARIMA model on a small data set. However, for a large data set I think it would create too many training sets. For example, if I had a data set with 50,000 rows and I was using a random forest model, I wouldn't want to train ~50,000 random forests. :)

I think the usual approach, which you can see in sci-kit learn ([here](https://scikit-learn.org/stable/modules/cross_validation.html#time-series-split) and [here](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.TimeSeriesSplit.html)), is to divide the data set into K+1 folds, and then iteratively train on folds 1 through M and test on fold M+1. In this approach each fold (and therefore each test fold) contains more than one data point.

If you add a field or two to your `TimeSeriesCV` type, you can probably make it general enough to cover both cases. I wouldn't get too fancy, though. There are all kinds of more sophisticated time series CV approaches that folks can implement if they want something more specific, as described in [this blog post](https://medium.com/@samuel.monnier/cross-validation-tools-for-time-series-ffa1a5a09bf9).

On a related note, I've wondered if it would make sense to keep each of the training sets the same size. For example, if the data were partitioned into 6 partitions, the train test pairs could look like this:

```julia
([1, 2, 3], [4])  # where an integer here represents a whole partition.
([2, 3, 4], [5])
([3, 4, 5], [6])
```

It seems like in at least some cases it would make sense to ensure that each model is trained on the same amount of data, but I don't know if people actually do it this way.

aa25desh:
> I can see how this would be useful when working with an ARIMA model on a small data set. 

Yes, you are correct `TimeSeriesCV` in this shape would work only on small data sets.

> I think the usual approach, which you can see in sci-kit learn ([here](https://scikit-learn.org/stable/modules/cross_validation.html#time-series-split) and [here](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.TimeSeriesSplit.html)), is to divide the data set into K+1 folds, and then iteratively train on folds 1 through M and test on fold M+1. In this approach each fold (and therefore each test fold) contains more than one data point.
>

Sure, I was reading about `TimeSeriesCV`(kfold) [here](https://www.researchgate.net/publication/256720783_On_the_use_of_cross-validation_for_time_series_predictor_evaluation) which i think is same as implemented in sklearn. Again `TimeSeriesCV`(kfold) need assumption that data (Time Series) is stationary.
I will try it out:)  

> If you add a field or two to your `TimeSeriesCV` type, you can probably make it general enough to cover both cases. I wouldn't get too fancy, though. There are all kinds of more sophisticated time series CV approaches that folks can implement if they want something more specific, as described in [this blog post](https://medium.com/@samuel.monnier/cross-validation-tools-for-time-series-ffa1a5a09bf9).
> 

Sure thing!

> On a related note, I've wondered if it would make sense to keep each of the training sets the same size. 
> It seems like in at least some cases it would make sense to ensure that each model is trained on the same amount of data, but I don't know if people actually do it that way.

Yes, we can. I was following [blog](https://robjhyndman.com/hyndsight/tscv/) so kept it that way, & I too think it is better to have same size training set.

CameronBieganek:
@aa25desh Thanks for the links! The blog post by Rob Hyndman got me thinking. First, let me set up some example data with a time series variable `y` and a time series covariate `x`:

```julia
using DataFrames
using ShiftedArrays

y = 1:15

df = DataFrame(
    x = 103:114,
    ylag2 = lag(y, 2)[3:14],
    ylag1 = lag(y, 1)[3:14],
    y = y[3:14],
    ylead1 = lead(y, 1)[3:14]
)
```

```julia
julia> df
12×5 DataFrame
│ Row │ x     │ ylag2  │ ylag1  │ y     │ ylead1 │
│     │ Int64 │ Int64⍰ │ Int64⍰ │ Int64 │ Int64⍰ │
├─────┼───────┼────────┼────────┼───────┼────────┤
│ 1   │ 103   │ 1      │ 2      │ 3     │ 4      │
│ 2   │ 104   │ 2      │ 3      │ 4     │ 5      │
│ 3   │ 105   │ 3      │ 4      │ 5     │ 6      │
│ 4   │ 106   │ 4      │ 5      │ 6     │ 7      │
│ 5   │ 107   │ 5      │ 6      │ 7     │ 8      │
│ 6   │ 108   │ 6      │ 7      │ 8     │ 9      │
│ 7   │ 109   │ 7      │ 8      │ 9     │ 10     │
│ 8   │ 110   │ 8      │ 9      │ 10    │ 11     │
│ 9   │ 111   │ 9      │ 10     │ 11    │ 12     │
│ 10  │ 112   │ 10     │ 11     │ 12    │ 13     │
│ 11  │ 113   │ 11     │ 12     │ 13    │ 14     │
│ 12  │ 114   │ 12     │ 13     │ 14    │ 15     │
```

The training data set is set up for any flavor of AR(2) (auto-regressive) model that makes predictions for `y` one and two steps ahead. So, the first three columns are the input features, and the last two columns are the (multivariate) response. A 3-fold time series cross-validation of the sci-kit learn variety would split the data into train-test pairs like this:

```julia
[
    (1:3, 4:6),
    (1:6, 7:9),
    (1:9, 10:12)
]
```

That's all fine so far, but it raises a question. @ablaom Does MLJ currently support models with multivariate responses? Also, will MLJ accept custom metrics/measures written to work on multivariate predictions?

I think in most time series applications you predict 1 through H steps ahead, and you want a separate accuracy estimate for each step, h, ahead. Generally of course the accuracy will go down the farther into the future that you are predicting. Then you can create plots of predictions with error bands like this:

image

(Image borrowed from Rob Hyndman's [online time series book](https://otexts.com/fpp2/prediction-intervals.html).)

ablaom:
> That's all fine so far, but it raises a question. @ablaom Does MLJ currently support models with multivariate responses? Also, will MLJ accept custom metrics/measures written to work on multivariate predictions?

A model can return a multivariate response; a few already do:

```julia
julia> models() do m
       m.target_scitype <: Table && m.is_supervised
       end
4-element Array{NamedTuple{(:name, :package_name, :is_supervised, :docstring, :hyperparameter_ranges, :hyperparameter_types, :hyperparameters, :implemented_methods, :is_pure_julia, :is_wrapper, :load_path, :package_license, :package_url, :package_uuid, :prediction_type, :supports_online, :supports_weights, :input_scitype, :target_scitype, :output_scitype),T} where T<:Tuple,1}:
 (name = MultiTaskElasticNetCVRegressor, package_name = ScikitLearn, ... )
 (name = MultiTaskElasticNetRegressor, package_name = ScikitLearn, ... )  
 (name = MultiTaskLassoCVRegressor, package_name = ScikitLearn, ... )     
 (name = MultiTaskLassoRegressor, package_name = ScikitLearn, ... )       
```

Yes, you can specify a custom metric: 

https://alan-turing-institute.github.io/MLJ.jl/dev/performance_measures/#Traits-and-custom-measures-1

So, your custom metric would have a signature like `multi_rms(yhat, y)` where `yhat` and `y` are tables with the multi-targets as columns. There is a tool to tabularize vectors, `MLJBase.table` (basically the same as `Tables.table`). 

However, it would be great to add to the metrics API a way to construct "products" of existing metrics, and that is not there yet. In any case, proof of concept along above lines using a custom metric would be useful, I reckon.

ablaom:
> On a related note, I've wondered if it would make sense to keep each of the training sets the same size.

I'm no expert, but I do see people using "telescoping" train sets. I guess the idea is that you want to maximise the train set size when you can, to improve generalization, although this naturally complicates theoretical considerations.

CameronBieganek:
> A model can return a multivariate response; a few already do:
 
> So, your custom metric would have a signature like `multi_rms(yhat, y)` where `yhat` and `y` are tables with the multi-targets as columns.

Great, so it sounds like there wouldn't be any issues using `TimeSeriesCV` when the response is multivariate.

@aa25desh I didn't realize this initially, but I think we can normally expect that if someone is building a machine learning time series model, they will arrange their training data similarly to the example in my previous post. So, in that case the `horizon` option is not really needed, because each row is complete. In other words, each row contains all the features and all of the multivariate response variables. (Well, I guess the response would be split into a table `y`.) So in that case there's no need to skip rows between the training set and the test set. Does that make sense?

> I do see people using "telescoping" train sets.

Yeah, from what I see on the internet, the "telescoping" method does seem to be the standard, so we should probably make that the default.

@aa25desh To summarize:

- The `nfolds` field is probably the only necessary field for `TimeSeriesCV`.
- If you want to, you could add an `equal_train_size` field which defaults to `false`.

ablaom:
> Great, so it sounds like there wouldn't be any issues using TimeSeriesCV when the response is multivariate.

That's correct. You just actually do it yet without defining your custom metric.

ablaom:
@CameronBieganek What is your slack handle? Would like to invite you to `#mlj-dev` if that's okay.

CameronBieganek:
I'm not on Slack yet, but I suppose I might as well create an account. :)

vollmersj:
@CameronBieganek I second the request for joining slack. We trying to start MLJTime and would be great to have your feedback.

CameronBieganek:
@aa25desh Thanks for the updates!

There appears to be a small bug in the code:

```julia
julia> tscv = TimeSeriesCV(nfolds=3)
TimeSeriesCV(
    nfolds = 3) @ 1…64

julia> MLJBase.train_test_pairs(tscv, collect(1:9))
3-element Array{Tuple{Array{Int64,1},Array{Int64,1}},1}:
 ([1, 2, 3], [4, 5, 6])
 ([1, 2, 3, 4, 5, 6], [7, 8, 9])
 ([1, 2, 3, 4, 5, 6, 7, 8, 9], [])
```

Note that the third test set is empty. It's probably a good idea to include a unit test like this where the number of observations is not an even multiple of `n_folds + 1`.

We didn't address this before, but when `n_obs / (n_folds + 1)` has a remainder, it would probably be nice to put the remainder in the first partition (which is only used for training). Then each test set has exactly the same size. That's what scikit-learn does. [Quoting from their manual](https://scikit-learn.org/stable/modules/cross_validation.html#time-series-split):

> Also, it adds all surplus data to the first training partition, which is always used to train the model.

You can also see that behavior in the example that they have in the same section of the manual:

> ```python
> >>> from sklearn.model_selection import TimeSeriesSplit
> >>> X = np.array([[1, 2], [3, 4], [1, 2], [3, 4], [1, 2], [3, 4]])
> >>> y = np.array([1, 2, 3, 4, 5, 6])
> >>> tscv = TimeSeriesSplit(n_splits=3)
> >>> print(tscv)
> TimeSeriesSplit(max_train_size=None, n_splits=3)
> >>> for train, test in tscv.split(X):
> ...     print("%s %s" % (train, test))
> [0 1 2] [3]
> [0 1 2 3] [4]
> [0 1 2 3 4] [5]
> ```

Here's a unit test that captures the desired behavior:

```julia
tscv = TimeSeriesCV(nfolds=3)
@test MLJBase.train_test_pairs(tscv, 1:11) == [
    (1:5, 6:7),
    (1:7, 8:9),
    (1:9, 10:11)
]
```

You might find the methods in the [Base.Iterators](https://docs.julialang.org/en/v1/base/iterators/) module to be useful when creating the partitions. [Here's an example](https://github.com/alan-turing-institute/MLJBase.jl/blob/7f023250dd497c3ead048c06f894bcbeee0113f7/src/resampling.jl#L151) of the usage of some of those methods to create partitions.

src/resampling.jl
```diff
@@ -165,6 +165,73 @@ function train_test_pairs(cv::CV, rows)
 end
 
 
+# ----------------------------------------------------------------
+# Cross-validation (TimeSeriesCV)
+"""
+tscv = TimeSeriesCV(;folds=4)
```
CameronBieganek, line 171: I think the actual name of the argument is `nfolds`, right?

aa25desh: Yes, it is nfolds.

src/resampling.jl
```diff
@@ -165,6 +165,73 @@ function train_test_pairs(cv::CV, rows)
...
+       TimeSeriesCV(nfolds)
+
+ function train_test_pairs(tscv::TimeSeriesCV, rows)
+     if rows != sort(rows)
+         @warn("TimeSeriesCV being applied to `rows` not in sequence. ")
+     end
+     n_obs = length(rows)
+     nfolds = tscv.nfolds
+     # number of observations per fold
+     k = floor(Int, n_obs/nfolds)
```
CameronBieganek, line 216: I think this is supposed to be `n_obs / (nfolds + 1)`.

By the way, I know that the old `CV` code used the variable `k` here, but I find it a little confusing in this context, because when talking about cross-validation `k` is usually the number of folds, e.g. "k-fold cross-validation". :)

src/resampling.jl
```diff
@@ -165,6 +165,73 @@ function train_test_pairs(cv::CV, rows)
...
+     k > 0 || error("Inusufficient data for $nfolds-fold cross-validation.\n"*
+                    "Try reducing nfolds. ")
+     # define the (trainrows, testrows) pairs:
+     firsts = 1:k:((nfolds)*k + 1) # itr of first `test` rows index
```
CameronBieganek, line 220: You can omit the nested parentheses here if you want, since the `:` operator has lower precedence than the arithmetic operations.

src/resampling.jl
aa25desh: For sure:)

CameronBieganek:
@aa25desh I apologize for springing new requirements on you. But I think if you add that new unit test to the test set that you already have, then that should fully capture the behavior we're looking for.

src/resampling.jl
```diff
@@ -165,6 +165,73 @@ function train_test_pairs(cv::CV, rows)
...
+
+train_test_pairs(tscv, rows)
+
+Return an iterator of `(train, test)` pairs of vectors in which `train` progressively
+grows in  `nfolds` and `test` consists of a `1+nfolds` indexs in `rows`.
+Specifically,
+
+train[k] = rows[1:k]
+test[k] = [rows[k + 1, ]
+for k in 2:(nfold).
```
CameronBieganek, line 185: I'm a little confused by this example. However, it will probably need to be updated anyways to reflect the behavior in the new unit test. :)

ablaom:
Be good to see this PR move forward. 

@aa25desh It seems that @CameronBieganek is still waiting for some feedback. Are you happy to address the comments, or do you to consider the suggested changes out of scope for now?

@CameronBieganek Are happy with the PR as is?

cc: @vollmersj

ablaom:
Closing in favour of #331 where merge conflicts are resolved.

PR #283, opened 2020-05-01
removed channels for CPU1 and CPUThreads

OkonSamuel:
This PR does 2 things.
1. Removes  the use of channels for displaying progress meters  for `CPU1` and `CPUThreads`. Channels are still being used for `CPUProcesses`
2. Makes progress meter display  functionality compatible with that of `MLJTuning` (PR fro this would be created at MLJTuning soonest)

### DIFF
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -91,12 +91,9 @@ end
 
 """
     cv = CV(; nfolds=6,  shuffle=nothing, rng=nothing)
-
 Cross-validation resampling strategy, for use in `evaluate!`,
 `evaluate` and tuning.
-
     train_test_pairs(cv, rows)
-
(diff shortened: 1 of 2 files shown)


ablaom:
@OkonSamuel Thanks for this!

I see you have also made changes to the CV resampling strategy as well. I wonder if you could comment on these. We recently had a very thorough review and a merged PR on CV from @CameronBieganek, who may wish to comment.

(I haven't looked at the channel stuff, but imagine there won't be issues there.)

OkonSamuel:
@ablaom . Although it appears so. I haven't really made any changes to CV resampling. It happens to be that my local repo wasn't up to date with the last merge you made so i kinda fixed that after opening this PR

PR #288, opened 2020-05-04
For 0.13.4 release - Take 2

ablaom:
The new implementation of `schema` requires `MLJModelInterface.schema` to exist, so the [compat] needed bumping.

Luckily I caught this before auto-merge.

cc @aviatesk


aviatesk:
nice catch, I was just about to notice this to you.  LGTM.

PR #293, opened 2020-05-09
Add MAPE

azev77:


OkonSamuel:
Thanks so much for this @azev77 . I have gone through and it looks good. Please could you export `mae` at **src/MLJBase.jl** under *measure exports* section. Also some tests at **test/measures/continuous.jl** would be great.

azev77:
You mean export ‘male’?

azev77:
@OkonSamuel  how can I export `mape` in the same PR?

OkonSamuel:
>  how can I export `mape` in the same PR?

just edit the code from the branch at https://github.com/azev77/MLJBase.jl/tree/patch-1 at it will update this pr

src/measures/continuous.jl
```diff
@@ -272,3 +272,43 @@ function (::RMSP)(ŷ::Vec{<:Real}, y::Vec{<:Real})
...
+    reports_each_observation = false,
+    is_feature_dependent     = false,
+    supports_weights         = false,
+    docstring                = "Mean Absolute Percentage Error; aliases: `mape`.")
+
+function (::MAPE)(ŷ::Vec{<:Real}, y::Vec{<:Real})
+    check_dimensions(ŷ, y)
+    ret = zero(eltype(y))
+    count = 0
+    for i in eachindex(y)
```
tlienart, line 306: can add `@inbounds` here

src/measures/continuous.jl
```diff
@@ -272,3 +272,43 @@ function (::RMSP)(ŷ::Vec{<:Real}, y::Vec{<:Real})
...
+        if y[i] != zero(eltype(y))
```
tlienart, line 307: this is not great numerically, might be better to use approx with an absolute tolerance that can be passed as parameter of the measure something like 

```julia
abs(y[i]) < thresh
```

you could also store `ayi = abs(y[i])` as it's used in the next expression.

src/measures/continuous.jl
```diff
@@ -272,3 +272,43 @@ function (::RMSP)(ŷ::Vec{<:Real}, y::Vec{<:Real})
...
+            dev = abs((y[i] - ŷ[i]) / y[i])
+            ret += dev
+            count += 1
+        end
+    end
+    return ret / count
```
tlienart, line 313: there's a possible corner case where `count  = 0`.

Issue #95, opened 2019-11-12
Review of MLR metrics


tlienart:
https://mlr.mlr-org.com/articles/tutorial/measures.html

### classification

MLR | MLJBase | Comment
---- | --- | ---
acc | accuracy | 
auc | auc | 
bac | **missing** | 
ber | **missing** |
brier | BrierScore |
brier scaled | **missing?** | maybe worth checking?
f1 | f1score |
fdr | **missing** | #88 
fn | **missing** | #88 
fnr | **missing** | #88 
fp | **missing** #88 
fpr | **missing** | #88 
gmean | **missing** | 
gpr | **missing** |
kappa | **missing** |
logloss | cross entropy ?  | **check**
lsr | **missing** |
mcc | **missing** |
mmce | **missing** |
multiclass au1p | **missing** | 👀 
multiclass au1u | **missing** |
multiclass aunp | **missing** |
mutliclass aunu | **missing** |
multiclass brier | **missing** |
npv | **missing** | #88 
ppv | **missing** | #88 
qsr | **missing** | 
ssr | **missing** |
tn | **missing** | #88 
tnr | **missing** | #88 
tp | **missing** | #88 
tpr | **missing** | #88 
wkappa | **missing** | #88 

### regression

some of these may be available in LossFunctions (?) + I did this one on the top of my head so may be worth double checking

MLR | MLJBase | Comment
---- | --- | ---
arsq | **missing** |
expvar | **missing** |
kendalltau | **missing** |
mae | mav | **check**
mape | **missing** | 
medae | **missing** |
medse | **missing** |
mse | mse |
msle | **missing** |
rae | **missing** |
rmse | rms | 
rmsle | rmsl |
rrse | **missing** |
rsq | **missing** |
sae | **missing** |
spearmanrho | **missing** |
sse | **missing** |

tlienart:
Might be good to also add stuff like rsquared, SSE, TSS etc for regression metrics

azev77:
@tlienart I submitted a [PR](https://github.com/alan-turing-institute/MLJBase.jl/pull/293/files) to add MAPE. 

Copying #294:
Rob's paper: https://robjhyndman.com/papers/mase.pdf (might be a newer version)
has a nice review of different measures of forecast accuracy.

Would be nice to incorporate at some point.
(They focus on univariate time series, but their results apply more broadly.)

2.1 Scale-dependent measures:   **e := ŷ - y**
Mean Square Error (MSE)
[...]

PR #296, opened 2020-05-09
MAPE

azev77:


azev77:
@OkonSamuel I think this one has the stuff

src/measures/continuous.jl
```diff
@@ -272,3 +272,39 @@ function (::RMSP)(ŷ::Vec{<:Real}, y::Vec{<:Real})
     end
     return sqrt(ret / count)
 end
+
+struct MAPE <: Measure end
+
+"""
+     mape(ŷ, y)
+Mean Absolute Percentage Error:
+``\\text{MAPE} =  m^{-1}∑ᵢ|{yᵢ-ŷᵢ \\over yᵢ}|``
```
OkonSamuel, line 281: am not too familiar with latex but shouldn't there be brackets here like ``\\text{MAPE} =  m^{-1}∑ᵢ|{(yᵢ-ŷᵢ )\\over yᵢ}|``

azev77: @OkonSamuel  yea

OkonSamuel: let me help you edit that

tlienart:
ah crap I commented on the wrong PR, please see comments in #293 there's a few numerical issues to sort out.

azev77:
@tlienart 
1 added `@inbounds`.
This can be added other places in the same program (I don't wanna touch other people's code)

2 regarding a tolerance, this is an issue w/ any measure based on percentage errors #95 
(currently RMSP does the same thing)
New code:
```julia
    @inbounds for i in eachindex(y)
        ayi = abs(y[i])
        if ayi > tol
        #if y[i] != zero(eltype(y))
            dev = abs((y[i] - ŷ[i]) / ayi)
            #dev = abs((y[i] - ŷ[i]) / y[i])
            ret += dev
            count += 1
        end
    end
```

3 regarding the corner case where `count = 0`, that is an inevitable [known problem](https://robjhyndman.com/papers/mase.pdf) w/ any percentage based score-measure (including RMSP).

tlienart:
Great thanks, @ablaom will probably complain on how you pass the tolerance parameter, it should be part of the MAPE struct.

Re inevitable errors etc, I would suggest showing a warning? 

Re improving RMSP to also have a threshold + use inbounds, we should do it 👍🏼

azev77:
None of the current measures have a tolerance parameter as part of the struct. 
How can this be done?
I don't know how to throw a warning...

azev77:
perhaps the current code could be written more parsimoniously using mean(), median etc? 
(unless loops are faster?)
Define vector: **e := ŷ - y**
image

Define vector: **p := e/y**  {for y[i]>tol}
image

etcetera?

tlienart:
I think @ablaom can better comment on this PR, he’s the one who thought carefully about measures.

I think the PR is great.

There’s a wider discussion of performance, warnings etc but it can be done separately. And yes afaik `mean` is two-three times faster than the loop, possibly bc it uses SIMD (?) , not sure.

Warning can be shown with `@warn` but better to do this for everything rather than have only one metric that would do it, so could be a separate issue.

azev77:
> Ps; performance is not really a problem with metrics...
I fully agree, that's why I wasn't sure about the need for `@inbounds`, but why not...

azev77:
@ablaom @tlienart see how much easier this could be: 
(the only things missing are weights & tol)
```julia
function err(ŷ, y)
    return ŷ - y
end

function perr(ŷ, y; tol = eps())
    e = err(ŷ, y)
    p = 100 * e[y .!= 0] ./ y[y .!= 0]
    return p
end

using Statistics
mse(ŷ, y)  = err(ŷ, y).^2 |> mean
rmse(ŷ, y) = err(ŷ, y).^2 |> mean |> sqrt
mae(ŷ, y)  = abs.(err(ŷ, y)) |> mean
mdae(ŷ, y) = abs.(err(ŷ, y)) |> median
rmdse(ŷ, y) = err(ŷ, y).^2 |> median |> sqrt

mape(ŷ, y)    = abs.(perr(ŷ, y)) |> mean
mdape(ŷ, y)   = abs.(perr(ŷ, y)) |> median
rmspe(ŷ, y)   = perr(ŷ, y).^2 |> mean |> sqrt
rmdspe(ŷ, y)  = perr(ŷ, y).^2 |> median |> sqrt

"https://en.wikipedia.org/wiki/Symmetric_mean_absolute_percentage_error"
smape(ŷ, y)   = 200. * abs.(err(ŷ, y)) ./ (ŷ + y) |> mean
smdape(ŷ, y)   = 200. * abs.(err(ŷ, y)) ./ (ŷ + y) |> median

# TEST
y    = [-1, 0, 1, 2, 3, 4]
ŷ    = [-3, 1, 2, 2, 4, 5]

e = err(ŷ, y)
p = perr(ŷ, y)

mse(ŷ, y)
rmse(ŷ, y)
mae(ŷ, y) 
mdae(ŷ, y) 
rmdse(ŷ, y)

mape(ŷ, y)
mdape(ŷ, y)
rmspe(ŷ, y)
rmdspe(ŷ, y)
smape(ŷ, y)
smdape(ŷ, y) 
```

azev77:
I'm not sure about struct & I took out kwarg, but put in a default tol.
Can a measure have arguments w/ default values? 
```julia
struct MAPE <: Measure 
    tol::Real
end

"""
     mape(ŷ, y)
Mean Absolute Percentage Error:
``\\text{MAPE} =  m^{-1}∑ᵢ|{(yᵢ-ŷᵢ) \\over yᵢ}|`` 
where the sum is over indices such that `yᵢ≂̸0` and `m` is the number
of such indices.
For more information, run `info(mape)`.
"""
const mape = MAPE()

metadata_measure(MAPE;
    name                     = "mape",
    target_scitype           = Union{Vec{Continuous},Vec{Count}},
    orientation              = :loss,
    supports_weights         = false,
    docstring                = "Mean Absolute Percentage Error; aliases: `mape`.")

function (::MAPE)(ŷ::Vec{<:Real}, y::Vec{<:Real}, tol = eps())
    check_dimensions(ŷ, y)
    ret = zero(eltype(y))
    count = 0
    @inbounds for i in eachindex(y)
        ayi = abs(y[i])
        if ayi > tol
        #if y[i] != zero(eltype(y))
            dev = abs((y[i] - ŷ[i]) / ayi)
            #dev = abs((y[i] - ŷ[i]) / y[i])
            ret += dev
            count += 1
        end
    end
    return ret / count
end
```

ablaom:
No, the idea is that you choose the tolerance at time of instantiation of the measure. As in 

```julia
mape_dangerous = MAPE(tol=0)
mape_dangerous(yhat, y) 
```

The signature of the measure can only be `m(yhat, y)` or `m(yhat, y, w)`.

azev77:
@ablaom sorry, not clear
1. can you show me w/ the mape() example 
2. include it w/ the new upgraded measures.jl (if we're gonna throw out most of the code anyway)

ablaom:
> 1\. can you show me w/ the mape() example

See the new `mape` fork. Here's the diff: https://github.com/alan-turing-institute/MLJBase.jl/commit/9aabed73407090fc5b07b013c61748fd1c054e90

> 2\. include it w/ the new upgraded measures.jl (if we're gonna throw out most of the code anyway)

Are you volunteering to carry out the refactor at #299, then? I'm not going to get around to that for some time.

azev77:
1. thanks for the code 
2. while I'd love to help w/ the refactor I don't know enough about the details of MLJ. 
I spilled my guts [here](https://github.com/alan-turing-institute/MLJBase.jl/issues/95#issuecomment-626411420).

ablaom:
Closed in favour of #302 - essentially identical

ablaom:
@azev77 Many thanks for the PR.

Issue #300, opened 2020-05-11
replace dict of machine


OkonSamuel:
Just got to know that Julia collections are not thread safe when at least one thread is manipulating the collection. 
This example from discourse for `Dict` explains what i mean
```julia
dict = Dict{Int64,Vector{Float64}}()
nentries = 5e5

Threads.@threads for n in 1:nentries
    dict[n] = [rand() for i in 1:100]
end

@assert length(dict) == nentries
```
The assertion fails.

Currently MLJBase creates a `Dict` of `machines` clones  with a machine for each thread. There is another way to do this without which is thread safe. I'll open a PR soon to fix this.

ablaom:
I think there is dictionary in MLJTuning too.

https://github.com/alan-turing-institute/MLJTuning.jl/blob/f01d231b1a64b7e19d37adffa590c417f9a06268/src/tuned_models.jl#L520

PR #307, opened 2020-05-14
replace unsafe use of dict in Multithreaded code.

OkonSamuel:
Closes issue #300 .
### Further Additions in this PR include
1. `CPUThreads` now accepts an (Optional) integer (e.g `CPUThreads(5)` ) which can be used to control the number of threads that are spawned (as spawning threads has some overhead). Most times the default parameters do the trick.
2. A `Chunks` iterator is implemented to help (1) above.
3. The return values of `_evaluate` method now has the right type (You could check this if you like)

## A slight Note on the use of theses Computational Resources in the `evaluate` method
1. `CPU1()` can be used in all cases
2. when using `CPUThreads(n) or CPUThreads()` speed up is only present if the data is small and model involves heavy computations. The package implementing the code for the model must be thread safe. The magnitude of the speedup depends on the number of threads.
3. when using `CPUProcesses` speed up is only present if the data is small and model involves heavy computations. 
   .

### DIFF
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -573,33 +573,32 @@ end
 
 function _evaluate!(func, mach, ::CPUProcesses, nfolds, verbosity) 
     
-local ret
-@sync begin
+    local ret
+    @sync begin
         channel = RemoteChannel(()->Channel{Bool}(min(1000, nfolds)), 1)
-    verbosity < 1 || (p = Progress(nfolds,
-                 dt = 0,
-                 desc = "Evaluating over $nfolds folds: ",
-                 barglyphs = BarGlyphs("[=> ]"),
-                 barlen = 25,
-                 color = :yellow))
+        verbosity < 1 || (p = Progress(nfolds,
+                     dt = 0,
+                     desc = "Evaluating over $nfolds folds: ",
+                     barglyphs = BarGlyphs("[=> ]"),
+                     barlen = 25,
+                     color = :yellow))
         # printing the progress bar
-       verbosity < 1 || @async begin
-                    while take!(channel)
-                    next!(p)
-                    end
-                    end
-        
-    
-     @sync begin
-            ret = @distributed vcat for k in 1:nfolds
-        	r = func(mach, k)
-        	verbosity < 1 || begin
-                            put!(channel, true)
-                            yield()
-                            end
-        	r
-    	   end
-    verbosity < 1 || put!(channel, false)
+        verbosity < 1 || @async begin
+                          while take!(channel)
+                            next!(p)
+                          end
+                         end
+
+         @sync begin
+           ret = @distributed vcat for k in 1:nfolds
+                     r = func(mach, k)
+                     verbosity < 1 || begin
+                                        put!(channel, true)
+                                        yield()
+                                      end
+                     r
+        	     end
+         verbosity < 1 || put!(channel, false)
     end
     close(channel)
     end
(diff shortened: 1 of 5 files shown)


src/resampling.jl
```diff
@@ -609,47 +608,43 @@ end
...
-
+          end
+      end
+        
     end
-  
-    return reduce(vcat, results)
+    results = reduce(vcat, ret) ## has the right Type
+     return results
 end
```
ablaom, line 648: Not sure I understand how this small change effects the return type but I'll take your word for it.

src/resampling.jl
```diff
@@ -701,7 +696,17 @@ function evaluate!(mach::Machine, resampling, weights,
...
         end
     end
-    if acceleration isa CPUThreads
+     if acceleration isa CPUThreads
+        if acceleration.settings === nothing 
+            nthreads = Threads.nthreads()
+            #Tasks have some ovehead so tried to avoid excessive creation
+             # set default threshold at 50
+            n = div(nfolds, 50)
+            # The parameter passes as input to CPUThreads is the ntasks
```
ablaom, line 705: I suggest we make this "50" a global constant declared in src/MLJBase.jl.

src/resampling.jl
OkonSamuel:
```julia
julia> a = Vector(undef,3)
3-element Array{Any,1}:
 #undef
 #undef
 #undef

julia> for i in 1:length(a)
       a[i] = [1,2,3]
       end

julia> a
3-element Array{Any,1}:
 [1, 2, 3]
 [1, 2, 3]
 [1, 2, 3]

julia> reduce(vcat, a)
9-element Array{Int64,1}:
 1
 2
 3
 1
 2
 3
 1
 2
 3
```

ablaom: Got it, thanks.

ablaom: Still, I think the compiler still has to box things, no? Look this example:

```julia
function foo(I)
    a = Array{Any}(undef, 2)
    a[1] = I
    a[2] = I
    return [a...]
end

julia> @code_warntype foo(3)
```
```
Variables
  #self#::Core.Compiler.Const(foo, false)
  I::Int64
  a::Array{Any,1}

Body::Any
1 ─ %1 = Core.apply_type(Main.Array, Main.Any)::Core.Compiler.Const(Array{Any,N} where N, false)
│        (a = (%1)(Main.undef, 2))
│        Base.setindex!(a, I, 1)
│        Base.setindex!(a, I, 2)
│   %5 = Core._apply(Base.vect, a)::Any
└──      return %5
```

(not that I think this matters for this PR)

OkonSamuel: yeah it still boxes things but the `reduce` methods returns the right type reducing performance losses

ablaom:
Okay, sorry for being a little dense, but why do we set the threshold so high? 

So, a common use case would be CV with say, `nfolds` between about 2 and 10; no repeats.  If `RESAMPLING_THRESHOLD=50`, then we get `n=0` every time and so the default that the code picks out is `CPUThreads(1)`, no matter how many threads are available. What am I missing?

OkonSamuel:
> Okay, sorry for being a little dense, but why do we set the threshold so high?
> 
> So, a common use case would be CV with say, `nfolds` between about 2 and 10; no repeats. If `RESAMPLING_THRESHOLD=50`, then we get `n=0` every time and so the default that the code picks out is `CPUThreads(1)`, no matter how many threads are available. What am I missing?

Thats the idea. I set the minimum work load per task is 50.(The idea is that We don't want to use spawn on tasks that finish in less than a second). This choice is based on the assumption that it takes roughly 0.1 sec to call fit per fold.(0.1*50 is well above a second).

ablaom:
But `fit` might take 2 hours or 2 days, right? If `fit` is taking a few seconds or less, we probably don't care about acceleration at all. The exception might be a very small data set and a model with a big hyper-parameter space we want to explore in tuning. But I would say that's an exceptional use-case, not the norm. 

Indeed, on reflection I would be happy to use all available threads when I don't specify the number explicitly (the existing behaviour, I believe)

 [Side note: In any case, we should make sure our testing here is actually using more than one thread]

OkonSamuel:
> But `fit` might take 2 hours or 2 days

Oh that's very true. I guess I'll just set the default to the number of threads.

test/utilities.jl
```diff
@@ -98,5 +98,17 @@ end
...
+    #test for cases with exactly same work as threads
+    @test length(MLJBase.chunks(1:nthreads, nthreads)) == nthreads
+    #test for cases with more work than threads
+    @test length(MLJBase.chunks(1:nthreads + rand(1:nthreads), nthreads)) == nthreads
+    #test for cases with less work than threads
+    @test length(MLJBase.chunks(1:nthreads-1, nthreads)) == nthreads - 1
+end
+end
+
 end # module
```
ablaom, line 113: Cool.

Issue #267, opened 2020-04-27
New syntax for @from_network @pipeline and beyond


ablaom:
As I turn to define more canned composite models (eg `@stack`, `@continuous`) and with the issues #258, https://github.com/alan-turing-institute/MLJ.jl/issues/412, https://github.com/alan-turing-institute/MLJ.jl/issues/311 in mind, I find that I am not happy with the existing syntax for these macros. 

In short, the general purpose advanced-user macro `@from_network` is not sufficiently expressive, while the common use-case `@pipeline` syntax could be simpler (because if it really doesn't actually do what you want it to, you can always use the more powerful @from_newtork to fix that). In an attempt to cast both these from the same syntactic mould, I settled on an unfortunate compromise.

The following examples show more-or-less what I have in mind as substitutions:

**A simpler `@pipeline` macro**

```julia
# provide either types
@pipeline OneHotEncoder(drop_last=true) LinearRegressor() target=BoxCoxTransformer()

# or instances (to insantiate with defaults)
@pipeline OneHotEncoder LinearRegressor target=BoxCoxTransformer

```
The user doesn't get to name the field names of the composite - they are always just `model1`, `models2`, and so forth, and the name of the new composite type is autogenerated, unless overwritten with kwarg  `name=...`. 

If a scitype trait for the composite cannot be reliably determined, then it falls back to `Unknown` unless the user specifies a kwarg (as in `target_scitype = AstractVector{<:Finite}`). If the prediction type cannot be reliably determined, then a warning is issued and `:deterministic` is used, unless user overrides with `prediction_type=:probabilistic`.

**A more expressive (and Julia idiomatic) `@from_network` macro**

The following example how to handle multiple operations (see issue cited above):

```julia 
@from_network begin
    mutable struct MyComposite <: UnsupervisedNetwork
        preprocessor::OneHotEncoder=one_hot
        clusterer=kmeans
    end
    predict = yhat    # the node that furnishes predictions
    transform = W     # the node that furnishes transforms
    input_scitype = Table(Continous,Finite)
end
```

Again scitype traits that cannot be reliably deduced fallback to `Unknown` unless explicitly provided by kwargs.

Thoughts anyone?

ablaom:
Ultimately breaking but suggest keeping old syntax temporarily.

tlienart:
[...]

Issue #258, opened 2020-04-23
Implement learning network export for multiple-operations


ablaom:
At the moment if `N` is a node in a learning network, then the syntax

```julia
@from_network MyCompositeModel(....) <= N
```
effectively defines a `predict` (resp. `transform`) method that returns `N()` if the network is supervised (resp. unsupervised). 

Now a supervised model can have both a `predict` and a `transform`, as can an unsupervised model. As a first step towards #257, we need to extend the syntax to bind multiple operations to the output of calling multiple nodes. So, something like

```julia
@from_network MyComposite(....) predict = yhat() transform = Xout()
```
Here `yhat`  likely represents a node with the `predict` operation (or could be `predict_mean`, and so forth) and `Xout` likely represents a node with the `transform` operation. 

We might also want to allow `inverse_transform` while we are at it. 

Thoughts anyone?

ablaom:
Update: revised syntax proposal in #267

PR #310, opened 2020-05-18
Model composition refactor Part I

ablaom:
In preparation for addressing https://github.com/alan-turing-institute/MLJ.jl/issues/72, https://github.com/alan-turing-institute/MLJBase.jl/issues/267 and https://github.com/alan-turing-institute/MLJBase.jl/issues/258, this PR is Part I of a refactoring of the model composition code. It adds a number of enhancements at a relatively low level (building and training elements of a learning
network, and exporting it) and is only mildly breaking, and only for low-level interaction. It has no effect on the `@pipeline` or `@from_network` syntax - those changes will be Part II.

Here are the main changes:

1. Remove the distinction between `Machine` and `NodalMachine`,
   leaving only `Machine`.  Make the training arguments of any machine
   `AbstractNode`s (i.e., `Source` or `Node`). Previously, an argument
   could be ordinary data, but now the constructor will wrap the data
   in a source node in that case. This has no effect on how the user
   *constructs* machines.
   
2. (**mildly breaking**) If a machine has `Node`s as training
   arguments (i.e, is part of a learning network) then make
   `fit!(mach)` first train all machines on which `mach` depends. The
   old behavior, in which no other machine is touched, is obtained by
   calling `fit_only!(mach)` but this is not likely to be of interest
   to the general user.
   
3. Replace the serial scheduler that trains a learning network `Node`
   with an asynchronous scheduler using tasks.
   
4. Introduce a new machine constructor `machine!` which directly wraps
   a machine around a learning network.  For example,
   `mach=machine!(predict=yhat, transform=W)` defines a machine for
   which `predict(mach, Xnew) = yhat(Xnew)` and `transform(mach, Xnew)
   = W(Xnew)`. Whereas a custom `fit` method for a composite model
   previously returned `fitresults(yhat)`, now it returns
   `mach.fitresult, mach.cache, mach.report`, where `mach` is the
   machine wrapped around the network. This allows composite models
   exported "by hand" to implement multiple operations (addressing
   https://github.com/alan-turing-institute/MLJBase.jl/issues/258 for that case)
   
5. (**enhancement**) Add a macro `@node` for overloading static
   operations on nodes. For example, given a node `X` that returns a
   table, `Xtop = @node selectrows(X, 1:4)` now defines a new node
   returning the first four rows of the table. It is equivalent to the
   less parsimonious `Xtop = node(X -> selectrows(X, 1:4), X)`.
   
6. (**code re-organization**) Completely re-organize the material in
   `src/machines.jl` and `src/composites` to assist maintenance.

7. (**mildly breaking**) Reverse order of machines in `report(mach).machines` and `fitted_params(mach).machines` when `mach` is a machine bound to a composite model. In particular arrange the order to coincide with the natural one for `@pipeline` models (#314)

Notes:

- A byproduct of (1) is that even a machine with `Node` training
  arguments can be evaluated using `evaluate!(mach, ...)`, so long as
  these arguments have themselves been trained. 
  
- Because of (3), it should be relatively straightforward in the
  future to make the scheduler multi-threading or distributed,
  addressing https://github.com/alan-turing-institute/MLJ.jl/issues/72. Anticipating this, the PR already adds the `acceleration` keyword  to `fit!(::Node, ...)`, but disallows all but `CPU1()`. 

- Regarding (4), we now view machines wrapped around a learning network
  as the first step in any process that exports a learning network as
  a stand-alone model. As the `machine!` doc-string details, this adds
  interface points at an earlier stage for declaring:
  
  - what roles each source node plays in the exported composite model
    (`:input`, `:target` or `:weight`) and the user need not declare
    this at the time of constructing the sources (but can if they want)
	
  - whether the composite model is probabilistic, deterministic or
    unsupervised, where this cannot be inferred from the other data provided
	
  - what values the composite model should have for its traits
    `input_scitype`, `output_scitype` and `target_scitype`, which
    ordinarily cannot be safely inferred from the network.

- Hopefully, (5) will remove a some of the mystery around the construction
  of the more sophisticated learning networks, such as stacks.
    
- Hopefully this PR improves the maintainability of the composition
  code base, safe-guarding against @ablaom + 💥  + 🚌


ablaom:
While I don't anticipate anyone will want to a review this entire PR, all feedback welcome. I will be leaving this unmerged for at least a week.

Issue #282, opened 2020-05-01
Help people quickly build UnivariateFinite object


tlienart:
It seems that people are often getting tripped up by `UnivariateFinite` object. Typical situation:

- User has some way of getting a vector of **scores** `s` 
- User assumes that given it's just a single vector of scores it's obvious it's a binary case
- User gets annoyed because they can't just use `auc` or some similar measures.
- User asks and we end up saying something like:

```julia
c = classes(categorical([:label0, :label1])[1])
y = [UnivariateFinite(c, [1-s[i], s[i]]) for i in s]
``` 

what would be good is to just have a simple way to do this directly something like

```
UnivariateFinite(s) # ==> does the above
UnivariateFinite(s, [:label_0, :label_1]) # ==> does the above with specific labels
UnivariateFinite(w) # ==> with w has c columns, do it with c labels
UnivariateFinite(w,  [:label_0, :label_1, :label_2]) # ==> same but specifying labels
```

Solving this would go along with solving https://github.com/alan-turing-institute/MLJ.jl/issues/511

ablaom:
I think this is a great idea. While not abolutely necessary, it would be less work overall to incorporate with [this](https://github.com/alan-turing-institute/MLJ.jl/issues/511) remedy to another issue.

PR #317, opened 2020-05-22
UnivariateFiniteVector

tlienart:
With apologies for the big PR.

Attached PR to MMI: https://github.com/alan-turing-institute/MLJModelInterface.jl/pull/45 (**required for  CI**) and 

Related PR to  MLJLM using it: https://github.com/alan-turing-institute/MLJLinearModels.jl/pull/74

Related issues

* #312 (StableRNGs)
* #282 , https://github.com/alan-turing-institute/MLJ.jl/issues/511

--- 

### StableRNG

The StableRNGs stuff is pretty straightforward, it was a bit of a PITA to  put in place and check everything but nothing major to report apart from broken tests at the end of the `composites` one (https://github.com/alan-turing-institute/MLJBase.jl/blob/9231531371bf0c576e2fdb099a86164ac630b75c/test/composition/composites.jl#L429-L442

it's probably something silly

### UnivariateFiniteVector

This is basically a polished version of what I suggested in #282 which works. It essentially does all that we discussed, and is orders of magnitude faster than the status quo.

If you access a single element, then the corresponding `UnivariateFinite` object is fully constructed. Otherwise this is basically a  lazy container which allows the two main operations (`pdf` and `mode`) to be computed much faster, it also allows access to  the raw scores and classes in a much more straightforward way (both are just fields of the object).

```julia
julia> using MLJBase
julia> s = rand(100_000);
julia> y = categorical([-1,1]);
julia> c = classes(y[1]);
julia> @time u = [UnivariateFinite(c, [1-si,si]) for si in s];
  1.990588 seconds (13.77 M allocations: 1.008 GiB, 30.04% gc time)
julia> @time uv = UnivariateFiniteVector(s, c);
0.070527 seconds (389.29 k allocations: 17.569 MiB)
...
julia> @time mode.(u);
  0.120148 seconds (948.78 k allocations: 22.482 MiB)
julia> @time mode(uv);
  0.016595 seconds (23 allocations: 392.406 KiB)
...
julia> @time pdf.(u, c[1]);
  0.033399 seconds (199.50 k allocations: 3.807 MiB)
julia> @time pdf(uv, c[1]);
  0.000913 seconds (6 allocations: 781.438 KiB)
```

So basically it's massively faster. It also sorts the problem #282 allowing people to write stuff like

```julia
s = rand(100_000);
u = UnivariateFiniteVector(s) 
```

and it will slap some basic class name to it (`:class_k` for multiclass and `:negative` / `:positive` for the binary).  

### Behaviour

Here's a  gist where I show how it can be used, hopefully it should all feel natural: https://gist.github.com/tlienart/d2c555f46e9972ff63833d73ebd3a7c2


src/interface/univariate_finite_vector.jl
```diff
@@ -0,0 +1,102 @@
...
+        end
+        E = R
+        cl = c isa NTuple ? c : tuple(c...)
+        new{R,C,L,E}(s, cl)
+    end
+    # Simple constructor for multi-class case (array of scores)
+    function UnivariateFiniteVector(s::Arr{<:Real,2}, c)
+        size(s, 2) == length(c) ||
+            throw(DimensionMismatch("Number of classes don't match."))
+        all(sum(s[i,:]) .≈ 1. for i in size(s, 1)) ||
```
OkonSamuel, line 29: Shouldn't you use `≈ 1.` instead of `==1.` due to Floating point arithmetic issues. It would affect calculations for large class size. ??

tlienart: Hey sam thanks! I’m a bit confused isn’t what I did precisely what you’re suggesting?

tlienart:
I’m thinking an easy way out of the caveat is to just not make this a subtype of abstractvector, but leave the convenience functions. I think then `f.(...)` will amount to `f(...)`. I need to check this.

Also I think there needs to be additions for the metrics to cope with this object

OkonSamuel:
> Also I think there needs to be additions for the metrics to cope with this object

Yeah. But that shouldn't be too hard

src/interface/univariate_finite_vector.jl
OkonSamuel: Oops am kinda stressed out.  I thought you did otherwise. (sorry)

OkonSamuel:
@tlienart. This may be a dumb idea. 
Have you tried Overloading the `Base.Broadcast.broadcasted` function for `pdf` function.
All you need to do this is to define the following functions for  the `UnivariateFiniteVector` type
1. `length`
 2. `iterate`
3. `Base.Broadcast.broadcasted(::typeof(pdf), h::UnivariateFiniteVector)`

tlienart:
@OkonSamuel thanks a lot for your suggestion, I didn't know that `typeof(foo)` was a specific subtype of `Function` (I thought it was just... `Function`) so that you can hijack the broadcasting mechanism for specific functions. So that fixes the  caveat, thanks!

I fixed and simplified a bunch of things and will go over the metrics and then  it should be good.

ablaom:
@tlienart @OkonSamuel Thanks indeed for this ❤️  I will try to review soon.

ablaom:
The CI error (julia 1.4) would appear to be unrelated to the PR; I am investigating.

ablaom:
Thanks @tlienart for this terrific PR!

Firstly, I am wondering why the  classes/scores  in the `UnivariateFiniteVector` constructor are in the reverse order found in its scalar ananlogue. So, we already have `UnivariateFinite(classes, p)` but you implement `UnivariateFiniteVector(p, classes)`.  For the record, I started fixing this in this MLJModelInterface PR, but will wait for your feedback before proceeding: https://github.com/alan-turing-institute/MLJModelInterface.jl/pull/47

Secondly, there is some inconsistency with the scalar case regarding how to use probabilities in the absence of a pool. I realize that some of the existing API is very new and you may have overlooked it. The recent change is the addition of key-word arguments. Here is the doc-string (see below for examples):

---

    UnivariateFinite(classes, p; pool=nothing, ordered=false)

Construct a discrete univariate distribution whose finite support is
the elements of the vector `classes`, and whose corresponding
probabilities are elements of the vector `p`, which must sum to one $REQUIRE.

*Important.* Here `classes` must have type
 `AbstractVector{<:CategoricalValue}` and all elements are assumed to
share the same categorical pool. Raw classes *may* be used, but only provided
`pool` is specified. The possible values are:

- some `v::CategoricalVector`  such that `classes` is a subset of `levels(v)`

- some `a::CategoricalValue` such that `classes` is a subset of `levels(a)`

- `missing`, in which case a new categorical pool is created which has
  `classes` as its only levels.

In the last case specify `ordered=true` to order the new pool.

---

I'm open to changing this to help convergence of the two. But I feel that there should be some kind of "cognitive barrier" to constructing a distribution without a pre-existing pool. That is, the user should think twice before doing this and not be sucked into thinking this is the modus operandum. 

In the scalar case above, if you try to construct a distribution using "raw" classes (rather than categorical elements) then you get an error message telling you what to do. If you want to use raw classes you must explicitly declare `pool=missing`:

```julia
julia> UnivariateFinite(["M", "F"], [0.3, 0.7])
ERROR: ArgumentError: No `CategoricalValue` found from which to extract a complete pool of classes.
Possible remedies: 
(i) Specify `pool=c` where `c` is a `CategoricalArray` or CategoricalValue` with a complete pool;
(ii) Specify `pool=missing` to create a new pool limited to the classes given. 
 In case (ii) you must also specify `ordered=true` to order the pool. 
Stacktrace:
 [1] _classes(::Nothing) at ...

julia> UnivariateFinite(["M", "F"], [0.3, 0.7], pool=missing)
UnivariateFinite(F=>0.7, M=>0.3)
``` 

Or you could do this:

```julia
julia> v=categorical(rand(["M", "F", "N"], 7))
7-element CategoricalArray{String,1,UInt32}:
 "F"
 "M"
 "M"
 "M"
 "F"
 "F"
 "N"

julia> UnivariateFinite(["M", "F"], [0.3, 0.7], pool=v)
UnivariateFinite(F=>0.7, M=>0.3)

julia> classes(ans)
3-element CategoricalArray{String,1,UInt32}:
 "F"
 "M"
 "N"
```

By the way, when a pool is constructed from scratch the user should be able to say if it is ordered or not, a detail I think is missing in the current PR.

tlienart:
Thanks! I'll work on the argument ordering and will do what I can with the keyword arguments (I might need further guidance on that later on as some of this pool business is a bit foreign to me).

ablaom:
Many thanks @tlienart . I'll work on pushing this along and get your feedback before merging.

tlienart:
In light of your comment in MMI, should I rename the field “probabilities” instead of “scores”? I guess we should be consistent here :)

ablaom:
I will do this. I am also changing `classes` to `support` in a lot of places, to distinguish the complete pool, from those classes for which probabilities have been explicitly set (typically the non-zero ones). I'm in the middle of working on this branch.

 If you don't mind, leave this for now and I will get you to review in due course. Many thanks again for the tremendous job.

ablaom:
While I appreciate the huge effort that has gone into this - and I have learned a few things from my review and discussion above - I am not happy with a major detail I overlooked before, namely that it introduces `UnivariateFiniteVector`s as new kind of real array, rather than implementing it as a subtype of `AbstractVector{<:UnivariateFinite}`, my original suggestion. 

I would like to propose the alternative at https://github.com/alan-turing-institute/MLJBase.jl/pull/327

Issue #137, opened 2019-12-05
Allow `pdf`  to return scores for multiple classes in  one shot


tlienart:
At the moment `pdf` is well adapted for binary, but in the multiclass case, we may want to be able to do something like 

```julia
pdf(yhat, labels)
```

and get a vector of floating point scores with `c` entries where `c` is the  number of classes.

ablaom:
I think some care should be taken here as `pdf` is part of another package's interface, namely Distributions.jl. The suggestion above could be construed as type piracy, no? I mean, `yhat` is not a distribution, for starters. 

Is there a pressing use case for this functionality? I wonder if this is more a question of users wanting the output in a form with which they are familiar, rather than a form they actually need it in. But I am happy to be contradicted. 

For occasional use, `hcat([pdf.(yhat, c) for c in labels]...)` is not so bad, IMHO.

tlienart:
I think that's fair enough, the one extra case where I could see this being useful in some form is if you pass on the results to a higher level learner where the score per class would be features (maybe that's convoluted). 

Maybe this could be solved by using another name which would not clash like `get_scores` (that would apply to a vector of univariate finite) or `predict_scores` (which would do  both  in one shot) defined precisely as you said.

ablaom:
I think a remedy proposed at https://github.com/alan-turing-institute/MLJ.jl/issues/511#issuecomment-622666705 would allow us to avoid type piracy here.

ablaom:
Resolved by above PR merge. Syntax as suggested by issue poster.

PR #327, opened 2020-06-01
UnivariateFiniteArray <: AbstractArray{<:UnivariateFinite}

### LINKED ISSUE #282: Help people quickly build UnivariateFinite object

ablaom:
This PR, an alternative to https://github.com/alan-turing-institute/MLJBase.jl/pull/317, does this:

1. Implements a wrapper `UnivariateFiniteArray <:
AbstractArray{<:UnivariateFinite}` for arrays of `UnivariateFinite`
objects that has: (i) efficient and convenient constructors (resolves #282); and (ii)
performant broadcasting of `pdf` and `mode`, a major step to resolving #282.

2. Includes @tlienart's fix for #312 (StableRNGs)

3. Implements  ~~`pdf.(d_vec, [:class1, :class3, :class7])`~~  `pdf(d_vec, [:class1, :class2, :class3])` for any `u::AbstractArray{<:UnivariateFinite}` (resolving #137). This is maybe type piracy but by far the best alternative I've been able to think of after thinking about this a while (and is wise @tlienart's original suggestion). 

*Important* This PR depends on https://github.com/alan-turing-institute/MLJModelInterface.jl/pull/47 ✔️ merged

Further notes on the new wrapper in 1:

- TODO: 
   - [x] Make finite measures (`auc`, etc) broadcast `pdf` to automatically buy into the performance benefits. 
    - [ ] Models need to `predict` `UnivariateFiniteArray`s instead of `Array{<:UnivariateFinite, 1}` objects (more on this below). This will resolve #282.

- The new `UnivariateFiniteArray` type is a trivial extension of the
  existing `UnivariateFinite` type (same fields, new parameter `N`) and it 
  shares the same interface, and code to implement this.

- For now, no `UnivariateFiniteArray` constructor is provided. Rather,
  the existing constructor `UnivariateFinite` is overloaded to
  return arrays when the supplied probabilities have the appropriate shape:

```julia
julia> UnivariateFinite([:no, :yes], 0.4 , augment=true, pool=missing)
UnivariateFinite{Multiclass{2}}(no=>0.6, yes=>0.4)

julia> UnivariateFinite([:no, :yes], rand(3) , augment=true, pool=missing)
3-element UnivariateFiniteArray{Multiclass{2},Symbol,UInt8,Float64,1}:
 UnivariateFinite{Multiclass{2}}(no=>0.332, yes=>0.668)
 UnivariateFinite{Multiclass{2}}(no=>0.563, yes=>0.437)
 UnivariateFinite{Multiclass{2}}(no=>0.966, yes=>0.034)
```

- Setting a new key-word argument `augment=true` allows
  probabilities for the first class  to be inferred by summing all
  the others, as in the examples above. I do not see how this enhancement 
  can be provided without this (ie, inferred automatically) because some corner
  cases otherwise lead to ambiguity

- Performance demonstration:

```julia
N = 10_000;
probs = rand(N);

# old way:
@time begin
       u = [UnivariateFinite([:no, :yes], probs[i], augment=true, pool=missing)
       for i in 1:N]
       pdf.(u, :yes)
       end;
  0.883413 seconds (3.79 M allocations: 470.678 MiB, 17.03% gc time)

# new way:
@time begin
    u = UnivariateFinite([:no, :yes], probs, augment=true, pool=missing)
    pdf.(u, :yes)
end;
0.000688 seconds (350 allocations: 520.609 KiB)
```

#### Multi-class example

```julia
y = categorical(rand("abcdef", 100));
C = classes(y);

# got some subclasses encountered in training:
support = ['a', 'b', 'f'];

# got some probabilities
probs = rand(50, 3)
probs ./= sum(probs, dims=2)

# predict returns
UnivariateFinite(support, probs, pool=C)
```

#### Binary example

```julia
# got some classes:
y = categorical(rand(Bool, 100));
C = classes(y)

# got some probabilities associated with C[2]:
probs = rand(50);

# predict returns
UnivariateFinite(C, probs, augment=true)
```

More examples below.

### API

Here's the doc-string that will replace the existing
`UnivariateFinite` docstring:

---

    UnivariateFinite(support,
                     probs;
                     pool=nothing,
                     augmented=false,
                     ordered=false)

Construct a discrete univariate distribution whose finite support is
the elements of the vector `support`, and whose corresponding
probabilities are elements of the vector `probs`. More generally,
construct an abstract *array* of `UnivariateFinite` distributions by
choosing `probs` to be an array of one higher dimension than the array
generated.

Unless `pool` is specified, `support` should have type
 `AbstractVector{<:CategoricalValue}` and all elements are assumed to
 share the same categorical pool, which may be larger than `support`.

*Important.* All levels of the common pool have associated
probabilites, not just those in the specified `support`. However,
these probabilities are always zero (see example below).

If `probs` is a matrix, it should have a column for each class in
`support` (or one less, if `augment=true`). More generally, `probs`
will be an array whose size is of the form `(n1, n2, ..., nk, c)`,
where `c = length(suppport)` (or one less, if `augment=true`) and the
constructor then returns an array of size `(n1, n2, ..., nk)`.

```
using CategoricalArrays
v = categorical([:x, :x, :y, :x, :z])

julia> UnivariateFinite(classes(v), [0.2, 0.3, 0.5])
UnivariateFinite{Multiclass{3}}(x=>0.2, y=>0.3, z=>0.5)

julia> d = UnivariateFinite([v[1], v[end]], [0.1, 0.9])
UnivariateFiniteMulticlass{3}(x=>0.1, z=>0.9)

julia> rand(d, 3)
3-element Array{Any,1}:
 CategoricalArrays.CategoricalValue{Symbol,UInt32} :z
 CategoricalArrays.CategoricalValue{Symbol,UInt32} :z
 CategoricalArrays.CategoricalValue{Symbol,UInt32} :z

julia> levels(d)
3-element Array{Symbol,1}:
 :x
 :y
 :z

julia> pdf(d, :y)
0.0
```

### Specifying a pool

Alternatively, `support` may be a list of raw (non-categorical)
elements if `pool` is:

- some `CategoricalArray`, `CategoricalValue` or `CategoricalPool`,
  such that `support` is a subset of `levels(pool)`

- `missing`, in which case a new categorical pool is created which has
  `support` as its only levels.

In the last case, specify `ordered=true` if the pool is to be
considered ordered.

```
julia> UnivariateFinite([:x, :z], [0.1, 0.9], pool=missing, ordered=true)
UnivariateFinite{OrderedFactor{2}}(x=>0.1, z=>0.9)

julia> d = UnivariateFinite([:x, :z], [0.1, 0.9], pool=v) # v defined above
UnivariateFinite(x=>0.1, z=>0.9) (Multiclass{3} samples)

julia> pdf(d, :y) # allowed as `:y in levels(v)`
0.0

v = categorical([:x, :x, :y, :x, :z, :w])
probs = rand(3, 100)
probs = probs ./ sum(probs, dims=1)
julia> UnivariateFinite([:x, :y, :z], probs, pool=v)
100-element UnivariateFiniteVector{Multiclass{4},Symbol,UInt32,Float64}:
 UnivariateFinite{Multiclass{4}}(x=>0.194, y=>0.3, z=>0.505)
 UnivariateFinite{Multiclass{4}}(x=>0.727, y=>0.234, z=>0.0391)
 UnivariateFinite{Multiclass{4}}(x=>0.674, y=>0.00535, z=>0.321)
   ⋮
 UnivariateFinite{Multiclass{4}}(x=>0.292, y=>0.339, z=>0.369)
```

### Probability augmentation

Unless `augment=true`, sums of elements along the last axis (row-sums
in the case of a matrix) must be equal to one, and otherwise such an
array is created by inserting appropriate elements *ahead* of those
provided. This means the provided probabilities are associated with
the the classes `c2, c3, ..., cn`.

---

    UnivariateFinite(prob_given_class; pool=nothing, ordered=false)

Construct a discrete univariate distribution whose finite support is
the set of keys of the provided dictionary, `prob_given_class`, and
whose values specify the corresponding probabilities.

The type requirements on the keys of the dictionary are the same as
the elements of `support` given above. If the values (probabilities)
are arrays instead of scalars, then an abstract array of
`UnivariateFinite` elements is created, with the same size as the
array.


tlienart:
This is great! Thanks for taking the time to do it and sorry I didn’t manage to do it myself.

It will be really great when this is in and we magically remove the overhead of dealing with an array of probabilities :)

src/univariate_finite/arrays.jl
```diff
@@ -0,0 +1,71 @@
...
+       setindex!(u.prob_given_ref[ref], v.prob_given_ref[ref], i...)
+    end
+    return u
+end
+
+# performant broadcasting of pdf - sample is a cat. value:
+Base.Broadcast.broadcasted(::typeof(pdf),
+                           u::UniFinArr{S,V,R,P,N},
+                           cv::CategoricalValue) where {S,V,R,P,N} =
+    get(u.prob_given_ref, int(cv), zero(P))
```
tlienart, line 37: could you clarify the `zero(P)` fallback?

src/univariate_finite/arrays.jl
```diff
@@ -0,0 +1,71 @@
...
+    mode_flat = map(1:length(u)) do i
+        max_prob = maximum(dic[ref][i] for ref in keys(dic))
+        m = zero(R)
+        for ref in keys(dic)
+            if dic[ref][i] == max_prob
+                m = ref
+                break
+            end
+        end
+        return u.decoder(m)
```
tlienart, line 67: I might be missing something here but can't you achieve essentially the same but maybe a bit more readable using `findmax` ?

src/univariate_finite/types.jl
```diff
@@ -0,0 +1,299 @@
...
+    ::FI,
+    prob_given_class::AbstractDict{<:CategoricalValue, <:Prob})
+
+    probs = values(prob_given_class) |> collect
+    _check_probs_01.(probs)
+    _check_probs_sum(probs)
+
+    # retrieve decoder and classes from element
+    class1         = first(keys(prob_given_class))
+    S = scitype(class1)
```
tlienart, line 126: maybe put this closer to where `S` is used?

tlienart:
I should transform your comment here in a small tutorial 🤔

src/univariate_finite/arrays.jl
ablaom: Values in not in the support (but in the pool of all classes) get assigned probablity zero. But there should be a check here that `cv` is in the pool of all classes, which is there now.

src/univariate_finite/arrays.jl
ablaom: `findmax` finds the index in an iterable but we have a dictionary here.

src/univariate_finite/types.jl
ablaom: done

ablaom:
Still got AUC to do, but that should be all for this PR.

PR #338, opened 2020-06-08
fixes strange warnings

OkonSamuel:
see https://github.com/alan-turing-institute/MLJTuning.jl/pull/49#issuecomment-640089283

### DIFF
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -634,11 +634,11 @@ end
 function _evaluate!(func, mach, accel::CPUThreads, nfolds, verbosity)
    nthreads = Threads.nthreads()
     
-   if nthreads == 1
+   if nthreads == 1 || nfolds == 1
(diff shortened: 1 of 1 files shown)


OkonSamuel:
The Caveat to this is that it's slightly slower than the previous implementation if the time to run `evaluate` on each fold is large

ablaom:
If "slightly slower" means within a factor of 1.5, I'm not fussed. Better safe than sorry. 

You could put a note in the code saying "This could be faster if ... But leads to warnings..."

ablaom:
Mmm. Now that I look at the code, this adds quite a bit of complexity. Let's sleep on it.

ablaom:
Okay, after sleeping on it I will go with your suggestion at the other issue to live with the warnings for now until we have a better understanding of what's causing them. 

Many thanks for this and well-done discovering the work-around. We may yet need to use it.

PR #346, opened 2020-06-11
added _acceleration to Resampler update method

OkonSamuel:
- added _acceleration parameter to `update` method of `Resampler` (Important omission)
- removed redundant `@sync` and `close` methods. (Turns out the lock in `close` method also contributed to the issue)
- Better workaround to the silent warnings issue for the `evaluate` method.
cc @ablaom

### DIFF
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -595,36 +595,40 @@ function _evaluate!(func, mach, ::CPU1, nfolds, verbosity)
     return ret
 end
 
+
 function _evaluate!(func, mach, ::CPUProcesses, nfolds, verbosity) 
     
     local ret
     @sync begin
-        channel = RemoteChannel(()->Channel{Bool}(min(1000, nfolds)), 1)
-        verbosity < 1 || (p = Progress(nfolds,
+        verbosity < 1 || begin
+                      p = Progress(nfolds,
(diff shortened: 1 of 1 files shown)


src/resampling.jl
```diff
@@ -925,11 +931,14 @@ function MLJBase.update(resampler::Resampler{Holdout},
         _process_weights_measures(resampler.weights, resampler.measure,
                                   mach, resampler.operation,
                                   verbosity, resampler.check_measure)
+
+    _acceleration = _process_accel_settings(resampler.acceleration)
+
```
ablaom, line 936: Good catch

ablaom:
Many thanks. Merge when ready.

OkonSamuel:
@ablaom should i tag a new release. Or should would you mind doing that??

ablaom:
Oops. This was merged directly to master and not through dev. 

@OkonSamuel Would be great if you could update dev and tag a new release. Otherwise, I will do it on Monday.

OkonSamuel:
@ablaom. I tagged a new release earlier. 
At that time the **master** and **dev** were in sync so there isn't any need to update the **dev**. (the **dev** is just a commit ahead of **master**)

PR #348, opened 2020-06-15
avoid Unboxing and mem alloc 

OkonSamuel:
currently the current implementation boxes `channels` and `progressmeter` (because of the way they are defined and later used in closures passed to `mapreduce`) . This PR fixes that. 
Also the previous code does `[measurements_matrix[:,k]...]` which has unnecessary allocations (@ablaom is there any reason for this). I replaced with `measurements_matrix[:,k]` which has same behaviour(i.e creates a copy)
```julia
julia> A = permutedims(reshape(rand(100), (1,100)));
julia> @btime [$(A)[:,1]...]
  3.491 μs (102 allocations: 3.31 KiB)
julia> @btime $(A)[:,1]
  184.344 ns (1 allocation: 896 bytes)
```

### DIFF
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -755,7 +755,7 @@ function evaluate!(mach::Machine, resampling, weights,
     per_observation = map(1:nmeasures) do k
         m = measures[k]
         if reports_each_observation(m)
-            [measurements_matrix[:,k]...]
+            measurements_matrix[:,k]
         else
             missing
         end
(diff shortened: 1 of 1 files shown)


src/resampling.jl
```diff
@@ -755,7 +755,7 @@ function evaluate!(mach::Machine, resampling, weights,
     per_observation = map(1:nmeasures) do k
         m = measures[k]
         if reports_each_observation(m)
-            [measurements_matrix[:,k]...]
+            measurements_matrix[:,k]
         else
```
ablaom, line 759: There was a reason, but I forget what it was. Quite possibly all the refactoring has rendered the original reason obsolete. Tests are passing, so it's probably safe.

Issue #314, opened 2020-05-19
Reverse order of machine in composite model reports and fitted_params


ablaom:
This should make order consistent with "natural" order for special case of `@pipeline` models.

ablaom:
Closing in favour of https://github.com/alan-turing-institute/MLJ.jl/issues/553

Issue #301, opened 2020-05-11
Implement `measures(matching(y))`


ablaom:
We already have `models(matching(X, y))`.

Need to migrate  these parts of `MLJ/src/matching.jl` to MLJBase: `ModelChecker` struct (better called `Checker`, since it is model and measure checker now), `matching(X)`, `matching(X, y)`, `matching(X, y, w)` method definitions. The rest can stay but becomes a model for the `measures` code that lives in MLJBase. 

This will break MLJ which currently owns the `matching` function.

PR #350, opened 2020-06-24
Model composition refactor, Part II - new and improved syntax for @from_network and @pipelines

ablaom:
- Simplify the `@pipeline` syntax. Keep the old syntax but deprecate (#267)

- (**breaking**) Make the `@from_network` more expressive (#267). Old syntax is abandoned.

- (**breaking**) Add more user-friendly interface for inspecting
  training reports and fitted parameters of composite models. For
  example, if `composite = @pipeline OneHotEncoder KNNRegressor` and
  `mach = machine(composite, X, y)`, then access the fitted parameters
  of the machine associated with `KNNRegressor` using
  `fitted_params(mach).knn_regressor` (https://github.com/alan-turing-institute/MLJ.jl/issues/553, #314)

-  Remove the requirement to specify the `kind` of
  source node, as in `source(y, kind=:target)`. Arrange that the role
  of source nodes be inferred from the order in which they appear in
  learing network machine constructors (see PR #310). Deprecate 
  specification of `kind`. 

- Miscellaneous tweaks to the constructor for learning network machines 
   given in PR #310 (which is not yet on `dev`). This is now non-mutating and renamed 
   to `machine` instead of `machine!`.

**edit**

Also:

- (**enhancement**) Allow new kinds of measure queries, such as `measures(matching(y))` and `measures("rms")` (#301 

- Resolve minor bug with displaying results of some performance evaluations (#352)


ablaom:
Here are some of the relevant updated doc-strings.

---

    @pipeline model1 model2 ... modelk

Create an instance of an automatically generated composite model type,
in which the specified models are composed in order. This means
`model1` receives inputs, whose output is passed to `model2`, and so
forth. Model types or instances may be specified.

At most one of the models may be a supervised model, but this model
can appear in any position.

The `@pipeline` macro accepts several key-word arguments discussed
further below.

Static (unlearned) transformations - that is, ordinary functions - may
also be inserted in the pipeline as shown in the following example:

    @pipeline X->coerce(X, :age=>Continuous) OneHotEncoder ConstantClassifier,


### Target transformation and inverse transformation

A learned target transformation (such as standardization) can also be
specified, using the key-word `target`, provided the transformer
provides an `inverse_transform` method:

    @pipeline OneHotEncoder KNNRegressor target=UnivariateTransformer

A static transformation can be specified instead, but then
an `inverse` must also be given:

    @pipeline(OneHotEncoder, KNNRegressor,
              target = v -> log.(v),
              inverse = v -> exp.(v))

*Important.* By default, the target inversion is applied *immediately
 following* the (unique) supervised model in the pipeline. To apply
 at the end of the pipeline, specify `invert_last=true`.


### Optional key-word arguments

- `target=...` - any `Unsupervised` model or `Function`

- `inverse=...` - any `Function` (unspecified if `target` is `Unsupervised`)

- `invert_last` (default=`true`) - set to `true` to delay target
  inversion to end of pipeline

- `prediction_type` (default=`:deterministic` if not inferable) -
  prediction type of the pipeline; possible values: `:deterministic`,
  `:probabilistic`, `:interval`

- `operation` (default=`predict`) - operation applied to the
  supervised component model, when present; possible values:
  `predict`, `predict_mean`, `predict_median`, `predict_mode`

- `name` (defaults to auto-generated) - new composite model type name;
  can be any name not already in current global namespace

See also: [`@from_network`](@ref)

---

    @from_network mach mutable struct NewCompositeModel
           ...
    end

    @from_network mach begin
        [mutable] struct NewCompositeModel
           ...
        end
        <optional trait declarations>
    end

Create a new stand-alone model type called `NewCompositeModel`, using
the specified learning network machine `mach` as a blueprint.

For more on learning network machines, see [`machine`](@ref).


### Example

Consider the following simple learning network for training a decision
tree after one-hot encoding the inputs, and forcing the predictions to
be point-predictions (rather than probabilistic):

```julia
Xs = source()
ys = source()

hot = OneHotEncoder()
tree = DecisionTreeClassifier()

W = transform(machine(hot, Xs), Xs)
yhat = predict_mode(machine(tree, W, ys), W)
```

A learning network machine is defined by

```julia
mach = machine(Deterministic(), Xs, ys; predict=yhat)
```

To specify a new `Deterministic` composite model type `WrappedTree` we
specify the model instances appearing in the network as "default"
values in the following decorated struct definition:

```julia
@from_network mach struct WrappedTree
    encoder=hot
    decision_tree=tree
end
```
and create a new instance with `WrappedTree()`.

To allow the second model component to be replaced by any other
probabilistic model we instead make a mutable struct declaration and,
if desired, annotate types appropriately.  In the following code
illustration some model trait declarations have also been added:

```julia
@from_network mach begin
    mutable struct WrappedTree
        encoder::OneHotEncoder=hot
        classifier::Probabilistic=tree
    end
    input_scitype = Table(Continuous, Finite)
    is_pure_julia = true
end
```

---

    machine(Xs; oper1=node1, oper2=node2, ...)
    machine(Xs, ys; oper1=node1, oper2=node2, ...)
    machine(Xs, ys, ws; oper1=node1, oper2=node2, ...)

Construct a special machine called a *learning network machine*, that
"wraps" a learning network, usually in preparation to export the
network as a stand-alone composite model type. The keyword arguments
declare what nodes are called when operations, such as `predict` and
`transform`, are called on the machine.

In addition to the operations named in the constructor, the methods
`fit!`, `report`, and `fitted_params` can be applied as usual to the
machine constructed.

    machine(Probablistic(), args...; kwargs...)
    machine(Deterministic(), args...; kwargs...)
    machine(Unsupervised(), args...; kwargs...)
    machine(Static(), args...; kwargs...)

Same as above, but specifying explicitly the kind of model the
learning network is to meant to represent.

Learning network machines are not to be confused with an ordinary
machine that happens to be bound to a stand-alone composite model
(i.e., an *exported* learning network).


### Examples

Supposing a supervised learning network's final predictions are
obtained by calling a node `yhat`, then the code

```julia
mach = machine(Deterministic(), Xs, ys; predict=yhat)
fit!(mach; rows=train)
predictions = predict(mach, Xnew) # `Xnew` concrete data
```

is  equivalent to

```julia
fit!(yhat, rows=train)
predictions = yhat(Xnew)
```

Here `Xs` and `ys` are the source nodes receiving, respectively, the
input and target data.

In a unsupervised learning network for clustering, with single source
node `Xs` for inputs, and in which the node `Xout` delivers the output
of dimension reduction, and `yhat` the class labels, one can write

```julia
mach = machine(Unsupervised(), Xs; transform=Xout, predict=yhat)
fit!(mach)
transformed = transform(mach, Xnew) # `Xnew` concrete data
predictions = predict(mach, Xnew)
```

which is equivalent to

```julia
fit!(Xout)
fit!(yhat)
transformed = Xout(Xnew)
predictions = yhat(Xnew)
```

ablaom:
One behaviour change worth mentioning: Inverse target transformations are now performed *immediately after* the supervised model makes its prediction. Old behaviour was to apply at the end of the pipeline, which can be enforced with `invert_last=true`. If the supervised model is *probablistic*, then it does not make sense to immediately apply the inversion; however, one can now remedy this by setting `operation=predict_mode` (for, example) and if the user does not reset the `operation` (which is `predict` by default) then a warning is issued suggesting possible remedies:

```julia
julia> @pipeline OneHotEncoder ConstantRegressor target=UnivariateStandardizer
┌ Warning: Pipeline is applying a target inversion immediately after a probabilistic target prediction. Perhaps you want to set `operation=predict_mean` or `operation=predict_mode`. Alternatively, you may want to insert a function/model after the supervised model and set `invert_last=true`. 
```

ablaom:
Bad code coverage report is due to my turning testing of deprecated code off. Have re-instated.

PR #351, opened 2020-06-25
create a model for each thread

OkonSamuel:
The current implementation of `_evaluate!(func, mach, accel::CPUThreads, nfolds, verbosity` shares the state of `mach.model` across all threads. 
But the call to the`fit!` method from `_evaluate` calls `clean!(model)` which mutates the model(a shared state) on each thread causing data race(due to reading and writing from different threads). This so far hasn't caused any issues but there's no telling what would happen when more threads are used. Hence informing the decision to create a clone of the model on each thread(I really still don't like copying)
Another option is to maybe to call the `clean!` method outside the threads (To clean up) so subsequent calls to `clean` within `Threads.@spawn` would only be reading (provided mutation doesn't occur on a previously cleaned model i.e it depends on the user's implementation of `clean(model)` ) but not writing hence avoiding any possible data race.

The first option is a more general solution since it is independent of the users implementation of `clean(model)`
@ablaom Thoughts.

### DIFF
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -664,7 +664,7 @@ function _evaluate!(func, mach, accel::CPUThreads, nfolds, verbosity)
                         end
       
     #One tmach for each task:
-    machines = [mach, [machine(mach.model, mach.args...) for _ in 2:length(partitions)]...]  
+    machines = [mach, [machine(deepcopy(mach.model), mach.args...) for _ in 2:length(partitions)]...]  
    @sync for (i, parts) in enumerate(partitions)    
      Threads.@spawn begin
        results[i] = mapreduce(vcat, parts) do k  
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -662,9 +662,9 @@ function _evaluate!(func, mach, accel::CPUThreads, nfolds, verbosity)
                                 ProgressMeter.updateProgress!(p)
                               end
                         end
-      
-    #One tmach for each task:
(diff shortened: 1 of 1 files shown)


ablaom:
Another good catch, thanks! 

I think we can safely say that `clean!(clean!(model)) = clean!(model)` everywhere at present, and make that a requirement of the API. In that case the "pre-clean" alternative you suggest sounds good to me.

OkonSamuel:
>  `clean!(clean!(model)) = clean!(model)`

How can one enforce  the behaviour above in the API
i.e mutating based on some codition.
 ```julia
clean!(model) = cleaned_model
clean!(cleaned_model) = cleaned_model`
```

OkonSamuel:
Sorry a side note on something unrelated. Was going through some of the code you merged in a recent PR #310 
i can't find the definition of `fit!_only` method you defined [here](https://github.com/alan-turing-institute/MLJBase.jl/blob/breaking/src/machines.jl#L399)

ablaom:
> Maybe we can add the following to docs.

Yes, this is what I had in mind.  But I think it suffices in the docs to state that `clean!(clean!(model)) = clean!(model)` must hold under all circumstances. The example already in the docs is of this kind, no?

> fit!_only

Great catch! That  should be `fit_only!`. I will fix this in my newer PR. Looks like that `wait_for_upstream` is always called with `true` so this never got tripped. If I remember right, there isn't any real use-case for `false` - this is just some dispatch trickery.

OkonSamuel:
> The example already in the docs is of this kind, no?

Yes. I took the example from the docs

PR #359, opened 2020-07-06
Hyphens are replaced by underscores in column names from OpenML

darenasc:
Column names with hyphens are replaced with underscores in datasets from OpenML.

Resolves https://github.com/alan-turing-institute/MLJ.jl/issues/579


ablaom:
@darenasc Many thanks!

PR #375, opened 2020-07-22
Fixes

OkonSamuel:
This PR fixes 
1. A bug that ocurs due to indexing a non Array in `getter` method
2. `UnivariateFiniteArray` method definition for `cross_entropy` and `BrierScore()`

### DIFF
--- a/src/measures/finite.jl
+++ b/src/measures/finite.jl
@@ -61,7 +61,7 @@ function (c::CrossEntropy)(ŷ::Vec{<:UnivariateFinite},
     return broadcast(_cross_entropy, ŷ, y, c.eps)
 end
 # performant in case of UnivariateFiniteArray:
-function (c::CrossEntropy)(ŷ::Vec{<:UnivariateFinite{S,V,R,P}},
+function (c::CrossEntropy)(ŷ::UnivariateFiniteVector{S,V,R,P},
                            y::Vec) where {S,V,R,P}
     check_dimensions(ŷ, y)
     check_pools(ŷ, y)
@@ -166,7 +166,7 @@ function (::BrierScore{<:UnivariateFinite})(
 end
 # performant version in case of UnivariateFiniteArray:
 function (::BrierScore{<:UnivariateFinite})(
-    ŷ::Vec{UnivariateFinite{S,V,R,P}},
+    ŷ::UnivariateFiniteVector{S,V,R,P},
     y::Vec,
     w::Union{Nothing,Vec{<:Real}}=nothing) where {S,V,R,P<:Real}
 
--- a/src/univariate_finite/arrays.jl
+++ b/src/univariate_finite/arrays.jl
@@ -134,10 +134,17 @@ function Base.Broadcast.broadcasted(
 
     # will use linear indexing:
     v_flat = ((v[i], i) for i in 1:length(v))
-    getter((cv, i)) = get(u.prob_given_ref, int(cv), zero(P))[i]
+    
+    getter((cv, i)) = _getindex(get(u.prob_given_ref, int(cv), zero(P)), i)
     ret_flat = getter.(v_flat)
     return reshape(ret_flat, size(u))
 end
+#dummy function
+# returns `x[i]` for `Array` inputs `x` 
+# For non-Array inputs returns `x`
+#This avoids using an if statement 
+_getindex(x::Array,i)=x[i]
+_getindex(x, i) = x
 
 # pdf.(u, raw) where raw is scalar or vec
 function Base.Broadcast.broadcasted(
--- a/src/univariate_finite/arrays.jl
+++ b/src/univariate_finite/arrays.jl
@@ -135,16 +135,18 @@ function Base.Broadcast.broadcasted(
     # will use linear indexing:
     v_flat = ((v[i], i) for i in 1:length(v))
     
-    getter((cv, i)) = _getindex(get(u.prob_given_ref, int(cv), zero(P)), i)
-    ret_flat = getter.(v_flat)
+    getter((cv, i), dtype) = _getindex(get(u.prob_given_ref, int(cv), nothing), i, dtype)
+    
+    ret_flat = getter.(v_flat, P)
     return reshape(ret_flat, size(u))
 end
+        
 #dummy function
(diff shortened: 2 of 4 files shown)


src/measures/finite.jl
```diff
@@ -61,7 +61,7 @@ function (c::CrossEntropy)(ŷ::Vec{<:UnivariateFinite},
     return broadcast(_cross_entropy, ŷ, y, c.eps)
 end
 # performant in case of UnivariateFiniteArray:
-function (c::CrossEntropy)(ŷ::Vec{<:UnivariateFinite{S,V,R,P}},
+function (c::CrossEntropy)(ŷ::UnivariateFiniteVector{S,V,R,P},
```
ablaom, line 64: I don't understand the need for this restriction. Only the element type needs to be restricted. (Here `Vec` is `AbstractVector`.) The user ought to be able to supply `yhat` as an ordinary `Vector{<:UnivariateFinite}` if they want to, yes?

OkonSamuel: without this restriction julia is confused as to which method to choose.

> The user ought to be able to supply `yhat` as an ordinary `Vector{<:UnivariateFinite}` if they want to, yes?

Yes. The user can still do this with `(c::CrossEntropy)(ŷ::Vec{<:UnivariateFinite}, y::Vec)` method

OkonSamuel:
@ablaom Tests are failing due to an unrelated OpenML issue

src/measures/finite.jl
ablaom: Ah, I see. Like I did correctly for BrierScore. Good catch!

OkonSamuel: using the MLJBase master branch  you can confirm what am saying with the following code snippet
```julia
using CategoricalArrays, MLJ, MLJBase
d = UnivariateFinite(classes(categorical([1,0,1])), [0.2,0.8])
l = [d,d,d] #Vector{<:UnivariateFinite}
@which MLJBase.BrierScore()(l, y[1:3])
````
this out to point to https://github.com/alan-turing-institute/MLJBase.jl/blob/master/src/measures/finite.jl#L151 but it instead used the definition at https://github.com/alan-turing-institute/MLJBase.jl/blob/master/src/measures/finite.jl#L168

File: src/univariate_finite/arrays.jl [hunk 2]
```diff
@@ -134,10 +134,17 @@ function Base.Broadcast.broadcasted(
...
+    getter((cv, i)) = _getindex(get(u.prob_given_ref, int(cv), zero(P)), i)
     ret_flat = getter.(v_flat)
     return reshape(ret_flat, size(u))
 end
+#dummy function
+# returns `x[i]` for `Array` inputs `x` 
+# For non-Array inputs returns `x`
+#This avoids using an if statement 
+_getindex(x::Array,i)=x[i]
+_getindex(x, i) = x
```
ablaom, line 148: Okay, I'm guessing I was missing some corner case here. I'm having trouble seeing it. Could you add a test exposing the flaw? I imagine that will make it clear for me.

ablaom: Yes, OpenML server is down.

OkonSamuel: Okay.

OkonSamuel: using master the following snippet should result in an error
```julia
using CategoricalArrays, MLJ, MLJBase
c  = categorical([0,1,1])
d = UnivariateFinite(c[1:1], [1 1 1]')
v = categorical([0,1,1,1])
broadcast(pdf, d, v[2:end])
```
The above case seems wierd but can occur when resampling.
This reuslts in a `BoundsError` due to indexing a non- Array type in the`get` function from the `getter` method [here](https://github.com/alan-turing-institute/MLJBase.jl/blob/master/src/univariate_finite/arrays.jl#L137)
this is shown futher below
```julia
julia> get(d.prob_given_ref, 1, zero(Float64))
3-element Array{Int64,1}:
 1
 1
 1
julia> get(d.prob_given_ref, 2, zero(Float64))
0.0
julia> get(d.prob_given_ref, 2, zero(Float64))[2] #fails for every index apart from `1`
ERROR: BoundsError
Stacktrace:
 [1] getindex(::Float64, ::Int64) at ./number.jl:78
 [2] top-level scope at REPL[18]:1

Comment (ablaom) [reply src/univariate_finite/arrays.jl:148]: Oh, I see now. I was returning a zero scalar for missing index instead of zero *array*. A more serious error which I should have written a test to cover. 

I like your fix (no need to create the zero array), but could I suggest a minor tweak (that will help me understand the code later): Change the `zero(P)` to `nothing` and dispatch your `_getindex` on `Nothing` like this:

```
    getter((cv, i)) = _getindex(get(u.prob_given_ref, int(cv), nothing), i)
    ret_flat = getter.(v_flat)
    return reshape(ret_flat, size(u))
end
_getindex(x, i)=x[i]
_getindex(::Nothing, i) = zero(P)
```

And if you could add your MWE to the test. Also happy to do this myself - let me know.

Again many thanks.

Comment (OkonSamuel) [reply src/univariate_finite/arrays.jl:148]: @ablaom  but that would require me to move `_getindex(x, i)=x[i]`
and `_getindex(::Nothing, i) = zero(P)` inside  `Base.broadcasted` function. Happy to implement your suggestion if this is fine by you.

Comment (ablaom) [reply src/univariate_finite/arrays.jl:148]: Yes, move inside.

Comment (OkonSamuel) [reply src/univariate_finite/arrays.jl:150]: Yeah. I worked around that

File: test/univariate_finite/arrays.jl [hunk 3]
```diff
@@ -128,6 +128,11 @@ end
     @test broadcast(pdf, d, v[2:end]) == [0,0,0]
 end
 
+@testset "_getindex" begin
+   @test MLJBase._getindex(collect(1:4), 2, Int64) == 2
+   @test MLJBase._getindex(nothing, 2, Int64) == zero(Int64)
+end
+
 @testset "broadcasting mode" begin
```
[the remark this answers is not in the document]
ablaom: Nevermind, I see you did this earlier.

PR #386, opened 2020-08-10
Dev

tlienart:
Patch release with a few minor docstring adjustments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
   - JULIA_NUM_THREADS=30
 julia:
   - 1.0
-  - 1.4
+  - 1.5
   - nightly
 jobs:
   allow_failures:
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "MLJBase"
 uuid = "a7f614a8-145f-11e9-1d2a-a57a1082229d"
 authors = ["Anthony D. Blaom <anthony.blaom@gmail.com>"]
-version = "0.14.6"
+version = "0.14.7"
 
 [deps]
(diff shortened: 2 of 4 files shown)


tlienart:
tests fail on nightly, maybe not important right now... am checking on 1.5.

tlienart:
ok there's an issue on 1.5 as well; I suspect this is because 1.5 ships with modifications to the RNG and that there are a few stray tests that don't use StableRNGs; I'll check this later today if I have the time

ablaom:
Strange. I cannot reproduce the 1.5 fail locally.

ablaom:
Okay, this is more serious. StableRNGs *have* already been implented in the tests that fail, but they are not - how should I put this - stable 🤷‍♂️ 

Actually, I don't know yet if this is actually a StableRNGs issue or an issue with the Distributions.jl implementation of `rand` for `Distributions.Categorical` but the following runs on 1.5 while the assertion fails on 1.4:

```julia
using StableRNGs
using Distributions

d= Categorical([0.1, 0.2, 0.7])
rng = StableRNGs.StableRNG(600)
samples = rand(rng, d, 1000)
@assert Distributions.countmap(samples)[1] == 93
```

ablaom:
It's StableRNGs: 

https://github.com/rfourquet/StableRNGs.jl/issues/4

ablaom:
I think it's safe to merge this as the issue external and is only effecting the tests.

ablaom:
FYI problem is with Distributions after all. They are using a depredated method from Random.

ablaom:
https://github.com/JuliaStats/Distributions.jl/pull/1157

Issue #390, opened 2020-08-17
Update logic flawed in case of composite models with non-model fields


ablaom:
The discussion at https://github.com/alan-turing-institute/MLJ.jl/issues/628#issuecomment-673952999 has revealed a flaw in the logic of `update` for `Composite` models. This does not apply to models generated with `@pipeline` or `@from_network` but if one defines a composite ["by hand'](https://alan-turing-institute.github.io/MLJ.jl/dev/composing_models/#Method-II:-Finer-control-(advanced)-1) and includes fields in the model the are not models in the network, then mutating such fields is not forcing retraining of an associated machine. Indeed such mutations must force a new `fit` call to ensure a new learning network is built, for there is no way for the learning network to see such changes. 

Incidentally, there is no way in the current design to have "smart fitting" for such changes.

Also,  it's probably a good idea for the API to state explicitly that it is forbidden for a composite model `fit` method to mutate models passed as field values of the composite `struct`. Actually, this is mentioned as a general rule (applying to *all* fields of all models) but only in the `Adding new models` section of the manual.

ablaom:
Here is an example of a composite model that does not train as expected. It is a composite of KMeans clustering (for dimension reduction) with ridge regression, but with the following "coupling" of the hyper parameters: The ridge regularisation depends on the number of clusters used (less regularisation for greater number of clusters) and a user-specified "coupling" coefficient `K`. The KMeans model, ridge solver and coupling coefficient are fields of the composite. Only a change to the KMeans model is forcing a refit. 

This example is along the lines of a use-case suggested by @Leonardbcm in https://github.com/alan-turing-institute/MLJ.jl/issues/628#issuecomment-673952999 . It is not a MWE, but illustrates the kind of thing we want to support.

```julia
using MLJ

@load  KMeans pkg=Clustering
@load RidgeRegressor pkg=MLJLinearModels

mutable struct MyComposite <: DeterministicComposite
    clusterer     # the clustering model
    ridge_solver  # a ridge regression parameter we want to expose
    K::Float64    # a "coupling" coefficient
end

function MLJ.fit(composite::Composite, verbosity, X, y)

    Xs = source(X)
    ys = source(y)

    clusterer = composite.clusterer
    k = clusterer.k

    clustererM = machine(clusterer, Xs)
    Xsmall = transform(clustererM, Xs)

    # the coupling: ridge regularization depends on number of
    # clusters (and the specified coefficient `K`):
[...]

Issue #377, opened 2020-07-23
Prohibit distinct fields in composite models pointing to two models that are ===


ablaom:
This is a corner case that leads to unexpected behaviour. Introducing the prohibition is no real restriction as models can be `==`, just not `===`. For example, the following would not be allowed:

```julia
stand = Standardizer()
@pipeline stand SVC() target=stand
```

whose output is displayed as

```julia
Pipeline376(
    standardizer = Standardizer(
            features = Symbol[],
            ignore = false,
            ordered_factor = false,
            count = false),
    svc = SVC(
            kernel = LIBSVM.Kernel.RadialBasis,
            gamma = 0.0,
            weights = nothing,
            cost = 1.0,
            cachesize = 200.0,
            degree = 3,
            coef0 = 0.0,
            tolerance = 0.001,
            shrinking = true,
            probability = false),
    target = Standardizer(
            features = Symbol[],
            ignore = false,
            ordered_factor = false,
            count = false)) @045
```

but this would be fine:

```julia
@pipeline Standardizer() SVC() target=Standardizer()
```

(whose output is indistingishable)

You could also do

```julia
@pipeline Standardizer SVC target=Standardizer
```
which amounts to the same thing really. 

In the second example the two `Standardizer()` calls are creating new (mutable) instances that, while `==`, are not identical (`===`).

Actually, the first example works fine, but not if you  mutate the fields of the pipeline and do something with it. 

The issue exists because of the way learning network blueprints are turned into models and cannot be fixed without substantially disruptive changes. 

I suggest that in the composite model constructors an error is thrown explaining the remedy. Something like:

"Distinct fields of a composite model (that are themselves models) cannot have defaults that are identical (`===`), although they can be `==` (ie., corresponding nested fields can be `==`).  In particular, two components in a `@pipeline`, including a learned `target` transform, must not be identical. Consider constructing the field defaults separately or using `deepcopy`."

How's that sound?

ablaom:
To be more accurate, the required restriction is this:

"Two distinct fields of a composite model which are both associated with models in the underlying learning network cannot have identical values, although they can be `==` (corresponding nested fields are `==`). In particular, two model instances in a `@pipeline` call cannot be identical, only `==`. "

PR #391, opened 2020-08-18
Fix update logic for Composite models

ablaom:
This PR addresses #390  - **edited** and #377. To do so it is necessary to add syntax. 

**edited:** An updated proposal is in the comment  https://github.com/alan-turing-institute/MLJBase.jl/pull/391#issuecomment-677231383 further below.

---

Instead of 

```julia
MLJBase.fit(model::MyNewComposite, verbosity, X, y)
    ...
    mach = machine(surrogate, ...)
    return mach()
end
```

one needs to do

```julia
MLJBase.fit(model::MyNewComposite, verbosity, X, y)
    ...
    mach = machine(surrogate, ...)
    return!(mach, model)
end
```

to ensure `fit!` works correctly for composite models that have non-model fields (or, more obscurely, fields that are models but are not linked to models in the underlying learning network).

The old syntax is retained but gives deprecation warnings. I'm proposing a new patch release, followed immediately with a new minor version with deprecations removed. 

The PR includes pretty thorough tests.

Todo:

- [x] Add a doc-string for the new method `return!` which explains use and the logic it implies

- [ ] Update MLJ manual

- [ ] Update Data Science Tutorials (again!)


ablaom:
Mmm. Maybe calling a function `return!` is a bit perverse. For less cognitive dissonance, we could just extend current calling syntax from `mach()` to `mach(model)`??

ablaom:
The `return!` doc-string:

---



    return!(mach::Machine{<:Surrogate}, model::Model)

Equivalent to `mach(model)`, this call returns an object suitable for
returning in user-defined `Composite` model `MLJBase.fit` methods. Not
relevant when defining composite models using `@pipeline` or
`@from_network`.

### Example

The following code defines, "by hand", a new model type `MyComposite`
for composing standardization (whitening) with a deterministic
regressor:

```julia
mutable struct MyComposite <: DeterministicComposite
    regressor
end

function MLJBase.fit(model::MyComposite, verbosity, X, y)
    Xs = source(X)
    ys = source(y)

    mach1 = machine(Standardizer(), Xs)
    Xwhite = transform(mach1, Xs)

    mach2 = machine(model.regressor, Xwhite, ys)
    yhat = predict(mach2, Xwhite)

    mach = machine(Deterministic(), Xs, ys; predict=yhat)
    fit!(mach, verbosity=verbosity)

    return mach(model)  # or `return!(mach, model)`
end
```

ablaom:
Okay, after adding #377 into the mix, and for overall simplicity, I'm now proposing a breaking change that combines the `fit!(mach, verbosity=verbosity)` call with the new `return!` method, which means one less line of code. See the example at the end of this new doc-string draft. The old syntax would be allowed but deprecated. 

---

    return!(mach::Machine{<:Surrogate}, model, verbosity)

The last call in custom code defining the `MLJBase.fit` method for a
new composite model type. Here `model` is the instance of the new type
appearing in the `MLJBase.fit` signature, while `mach` is a learning
network machine constructed using `model`. Not relevant when defining
composite models using `@pipeline` or `@from_network`.

For usage, see the example given below. Specificlly, the call does the
following:

- Determines which fields of `model` point to model instances in the
  learning network wrapped by `mach`, for recording in an object
  called `cache`, for passing onto the MLJ logic that handles smart
  updating (namely, an `MLJBase.update` fallback for composite models).


- Calls `fit!(mach, verbosity=verbosity)`.

- Moves any data in sources nodes of the learning network into `cache`
  (for data-anonymization purposes).

- Records a copy of `model` in `cache`.

- Returns the cached data and outcomes of training in an appropriate
  form (specifically, `(mach.fitresult, cache, mach.report)`; see
  [Adding Models for General
  Use](https://alan-turing-institute.github.io/MLJ.jl/dev/adding_models_for_general_use/)
  for technical details.)


### Example

The following code defines, "by hand", a new model type `MyComposite`
for composing standardization (whitening) with a deterministic
regressor:

```
mutable struct MyComposite <: DeterministicComposite
    regressor
end

function MLJBase.fit(model::MyComposite, verbosity, X, y)
    Xs = source(X)
    ys = source(y)

    mach1 = machine(Standardizer(), Xs)
    Xwhite = transform(mach1, Xs)

    mach2 = machine(model.regressor, Xwhite, ys)
    yhat = predict(mach2, Xwhite)

    mach = machine(Deterministic(), Xs, ys; predict=yhat)
    return!(mach, model, verbosity)
end
```

ablaom:
The resolution of #377 involves adding some try-catch blocks. Here are some benchmarks which compare current PR with master. I'm using the stacking model `pipe` from https://github.com/ablaom/MachineLearningInJulia2020/blob/master/wow.ipynb .

**this PR**
```julia
julia> @btime fit!($mach, force=true, verbosity=-1)
  8.899 ms (17212 allocations: 2.84 MiB)
```

**master**
```julia
julia> @btime fit!($mach, force=true, verbosity=-1)
  8.426 ms (14901 allocations: 2.71 MiB)
```

About the same.

PR #392, opened 2020-08-18
Add the `:predict_joint` operation, and add composite and surrogate types for `JointProbabilistic`

DilumAluthge:
Please see: https://github.com/alan-turing-institute/MLJ.jl/issues/642

### DIFF
--- a/src/composition/abstract_types.jl
+++ b/src/composition/abstract_types.jl
@@ -1,6 +1,7 @@
 # true composite models:
 abstract type DeterministicComposite <: Deterministic end
 abstract type ProbabilisticComposite <: Probabilistic end
+abstract type JointProbabilisticComposite <: JointProbabilistic end
 abstract type IntervalComposite <: Interval end
 abstract type  UnsupervisedComposite <: Unsupervised end
 abstract type  StaticComposite <: Static end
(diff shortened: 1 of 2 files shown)


ablaom:
Thanks for this @DilumAluthge . Looks good to me. 

For completeness we should probably make corresponding additions to this file providing extras types for composite models: 

https://github.com/alan-turing-institute/MLJBase.jl/blob/8e25843bee7d705354a97bc573850f3154682cbe/src/composition/abstract_types.jl#L2

Specifically, we should add:

```julia
abstract type JointProbabilisticComposite <: JointProbabilistic end
...
struct JointProbabilisticSurrogate <: JointProbabilistic end
```

And extend the aliases for the Union types there and add a constructor `JointProbabilistic` for `JointProbabilisticSurrogate()` instances. 

(Check my spelling!)

DilumAluthge:
> Thanks for this @DilumAluthge . Looks good to me.
> 
> For completeness we should probably make corresponding additions to this file providing extras types for composite models:
> 
> https://github.com/alan-turing-institute/MLJBase.jl/blob/8e25843bee7d705354a97bc573850f3154682cbe/src/composition/abstract_types.jl#L2
> 
> Specifically, we should add:
> 
> ```julia
> abstract type JointProbabilisticComposite <: JointProbabilistic end
> ...
> struct JointProbabilisticSurrogate <: JointProbabilistic end
> ```
> 
> And extend the aliases for the Union types there and add a constructor `JointProbabilistic` for `JointProbabilisticSurrogate()` instances.
> 
> (Check my spelling!)

Sounds good, I'll add those changes!

DilumAluthge:
@ablaom Take a look and see if there is anything I missed.

DilumAluthge:
@ablaom @tlienart For `predict_joint`, we need to be able to pass keyword arguments.

The cleanest way to do this with the minimal amount of code changes was just to add kwarg support for all of the operations.

If you like, I can change it so that only `predict_joint` accepts kwargs.

ablaom:
I'm sorry but that's a big change to allow kwargs in operations. There are now kwargs anywhere in the "model" interface. I don't like the extra complication to the operation overloadings. We've managed until now without it. What's your use case?  

Maybe you need another operation instead? Or you can control the behaviour of `predict_joint` with model hyperparameters?

DilumAluthge:
I've removed the keyword arguments. Let me know if I missed one!

> (Or you can control the behaviour of predict_joint with model hyperparameters??)

I think this will be the best approach. We want to be able for users to say if they want predictions from e.g. (in the GLM example) `μ` or `y`. I think storing this preference in the model is fine.

ablaom:
Possibly we can explore the new interface point in another thread. I'm happy to merge this PR when tests pass.

DilumAluthge:
> Possibly we can explore the new interface point in another thread.

For sure. Chad and I are still experimenting in https://github.com/cscherrer/SossMLJ.jl, so there's lots of room to experiment and iterate.

That being said, it's coming together, so hopefully we'll have a full demo soon.

DilumAluthge:
> I'm happy to merge this PR when tests pass.

That would be great!

ablaom:
@DilumAluthge Thanks for making those extra changes.

PR #393, opened 2020-08-18
Make some code more concise

DilumAluthge:
This makes the `predict_mode`, `predict_mean`, and `predict_median` code more concise in the case of surrogate and composite models.

cc: @ablaom 
cc: @tlienart


DilumAluthge:
The Codecov failure is a false positive. There has not actually been a change in coverage.

ablaom (approved):
Looks good to me. Thanks @DilumAluthge !

PR #409, opened 2020-08-26
Okonsfixes

### LINKED ISSUE #405: Decouple interface points for training weights and weights used in measures.
From the current doc-string for `evaluate!`:

> Important: If mach already wraps sample weights w (as in mach = machine(model, X, y, w)) then these weights, which are used for training, are automatically passed to the measures for evaluation. However, for evaluation purposes, any weights specified as a keyword argument will take precedence over w.

This has emerged as a poor design decision. Here are two reasons:

- This is "hidden" knowledge. If you don't read the doc string, you likely don't realise this is happening. 

- It forces one to suppose that the third training argument in the construction `machine(model, X, y, w)` (corresponding to the lower-level `fit(model, verbosity, X, y, w)` call) *must* be a weight and can have no other interpretation. There are use-cases for more general signatures (see, eg, https://github.com/alan-turing-institute/MLJ.jl/issues/480). 

I propose removing the decoupling. If you want to pass weights to measures, you must explicitly specify them in the `evaluate`/`evaluate!` call. Some examples to make this clear:

```julia
# weights in training only:
evaluate(model, X, y, w, measure=rms)

# or:
mach = machine(model, X, y, w)
evaluate!(mach, measure=rms)

# weights in measures only:
evaluate(model, X, y, weights=w)

# or:
mach = machine(model, X, y)
evaluate!(mach, weights=w)

# weights in training AND measures:
evaluate(model, X, y, w, measure=rms, weights=w)

# or:
mach = machine(model, X, y, w)
evaluate!(model, X, y, measure=rms, weights)
```

OkonSamuel:
This PR 
- updates docstring of `evaluate` and `evaluate!` methods to reflect recent addition of `fitted_params_per_fold` etc.
- Decouple interface points for training weights and weights used in measures
- fixed a bug in `evaluate!` method where previously warning were display for `verbosity=0 && weights !==nothing`
closes #405

### DIFF
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -415,10 +415,6 @@ function _process_weights_measures(weights, measures, mach,
             throw(DimensionMismatch("`weights` and target "*
                                     "have different lengths. "))
         _weights = weights
-    elseif  length(mach.args) == 3
-        verbosity < 1 ||
-            @info "Passing machine sample weights to any supported measures. "
-        _weights = mach.args[3]()
     else
         _weights = weights
     end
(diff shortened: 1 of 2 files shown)


OkonSamuel:
closing in favor of #413

PR #411, opened 2020-08-26
Add logpdf method for UnivariateFinite

cscherrer:
Take 2!

This just adds a `logpdf` method and some tests. `logpdf` is needed for MCMC sampling with SossMLJ: https://github.com/cscherrer/SossMLJ.jl/issues/68

### DIFF
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -225,7 +225,7 @@ export TruePositive, TrueNegative, FalsePositive, FalseNegative,
 # re-export from Random, StatsBase, Statistics, Distributions,
 # CategoricalArrays, InvertedIndices:
 export pdf, sampler, mode, median, mean, shuffle!, categorical, shuffle,
-       levels, levels!, std, Not, support
+       levels, levels!, std, Not, support, logpdf
 
 
 # ===================================================================
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -46,7 +46,7 @@ import StatsBase
 import StatsBase: fit!, mode, countmap
 import Missings: levels
 import Distributions
-import Distributions: pdf, sampler
+import Distributions: pdf, logpdf, sampler
 const Dist = Distributions
 
 # from Standard Library:
--- a/src/univariate_finite/arrays.jl
+++ b/src/univariate_finite/arrays.jl
@@ -104,6 +104,7 @@ for func in [:pdf, :logpdf]
             return ret
         end
     end)
+end
 
 
 
--- a/src/univariate_finite/arrays.jl
+++ b/src/univariate_finite/arrays.jl
@@ -80,7 +80,8 @@ Base.vcat(us::UnivariateFiniteArray...) = cat(us..., dims=1)
 Base.hcat(us::UnivariateFiniteArray...) = cat(us..., dims=2)
 
 
-## CONVENIENCE METHOD pdf(array_of_univariate_finite, labels)
+## CONVENIENCE METHODS pdf(array_of_univariate_finite, labels)
+## AND logpdf(array_of_univariate_finite, labels)
 
 # next bit is not specific to `UnivariateFiniteArray` but is for any
 # abstract array with eltype `UnivariateFinite`.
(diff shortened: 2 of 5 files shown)


DilumAluthge:
@cscherrer You should edit line 6 of `test/univariate_finite/methods.jl`. Currently it is this:
https://github.com/alan-turing-institute/MLJBase.jl/blob/5188c318d5c663b0d1e4150bafdb6b1600c22cec/test/univariate_finite/methods.jl#L6-L6

 You should change it to this:
```
import Distributions: pdf, logpdf, support 
```

OkonSamuel:
@cscherrer  could you add `logpdf` to the list of exported functions here. https://github.com/alan-turing-institute/MLJBase.jl/blob/5188c318d5c663b0d1e4150bafdb6b1600c22cec/src/MLJBase.jl#L227

OkonSamuel:
Looks good to me

OkonSamuel (approved):

DilumAluthge:
Any chance this PR will make it into the upcoming minor release? cc: @ablaom

ablaom:
> Any chance this PR will make it into the upcoming minor release? cc: @ablaom

Yep.

ablaom (commented):
Thanks very much for this @cscherrer !

Looks good. However, I think we should also add the efficient implementation of broadcasting (over `UnivariateFiniteArray`), which we added for `pdf`. Otherwise broadcasting `pdflog` over a large `UnivariateFiniteArray` will be slow (see https://github.com/alan-turing-institute/MLJ.jl/issues/511). The analogous code is here:

https://github.com/alan-turing-institute/MLJBase.jl/blob/cc0998349c9df5ba7131191a637cc58fe4af64e7/src/univariate_finite/arrays.jl#L105

And (less crucially) you may want to add analogous overloading given here for `pdf`: https://github.com/alan-turing-institute/MLJBase.jl/blob/cc0998349c9df5ba7131191a637cc58fe4af64e7/src/univariate_finite/arrays.jl#L83

ablaom:
You could just wrap the existing code in evaluation block:

```julia
for func in [:pdf, :logpdf] 
   eval(quote .... end)
end
```

cscherrer:
> I think we should also add the efficient implementation of broadcasting (over `UnivariateFiniteArray`)

Just to make sure I understand, you're talking about `logpdf.(::UnivariateFiniteArray, ::AbstractArray)` (note the dot), and not `logpdf(::UnivariateFiniteArray, ::AbstractArray)`, right? I haven't worked with specializing broadcasting before, but this sounds sensible to me.

ablaom:
There's two separate things in my comment. The first thing (line 105) is to re-implementing the existing ` logpdf.(u::AbstractArray{<:UnivariateFinite{...}}, samples::AbstractArray{...})` syntax to make it faster when the the abstract array `u` happens to be `UnivarariateFiniteArray` (and not just `Array{<:UnivariateFinite}`) and `u` and `s` have the same shape (`size`). The return value is an array of the same shape. So typically, two vectors of length `n` in and one vector of length `n` out. 

The second (line 83) is technically some type piracy so that `pdf(yhat, classes)` works, when `yhat` is a `AbstractArray{<:UnivariateFinite}` and `classes` is any length but generally just a list of all the possible levels. In this case, if `yhat` is a vector, then a *matrix* is returned, one column per class. There was some discussion about introducing this [here](https://github.com/alan-turing-institute/MLJBase.jl/issues/137). It seems to me that having made the decision to do this for `pdf` we might as well do the same for `logpdf` while we are at it. 

Is that clearer?

cscherrer:
Does this look right?

```julia
## PERFORMANT BROADCASTING OF pdf and logpdf

# u - a UnivariateFiniteArray
# cv - a CategoricalValue
# v - a vector of CategoricalArrays

# pdf.(u, cv)

function Base.Broadcast.broadcasted(
    ::typeof(pdf),
    u::UniFinArr{S,V,R,P,N},
    cv::CategoricalValue) where {S,V,R,P,N}

    cv in classes(u) || _err_missing_class(cv)

    return get(u.prob_given_ref, int(cv), zeros(P, size(u)))
end

# logpdf.(u, cv)

function Base.Broadcast.broadcasted(
    ::typeof(logpdf),
    u::UniFinArr{S,V,R,P,N},
    cv::CategoricalValue) where {S,V,R,P,N}

    result = pdf.(u,cv)
    @simd for j in eachindex(p)
        @inbounds result[j] = log(result[j])
    end

    return result
end
```

ablaom (approved):
Super. Many thanks.

PR #413, opened 2020-08-26
Decouple training weights from evaluation weights

ablaom:
#405


src/resampling.jl
```diff
@@ -593,9 +599,12 @@ function evaluate!(mach::Machine{<:Supervised};
...
-        _process_weights_measures(weights, measure, mach,
-                                  operation, verbosity, check_measure)
+    _measures = _process_weights_measures(weights,
+                                          measure,
+                                          mach,
+                                          operation,
+                                          verbosity,
+                                          check_measure)
 
     if verbosity >= 0 && weights !== nothing
```
OkonSamuel, line 609: shouldn't `verbosity >=0` be replaced by `verbosity>0`

ablaom: No, to suppress warnings you need one lower level of verbosity than to suppress info. So verbosity = -1 will suppress warnings but verbosity = 0 will not. That's the pattern elsewhere.

ablaom: Do you think a warning is too much and that an `info` suffices here?

src/resampling.jl
```diff
@@ -344,15 +344,13 @@ function actual_rows(rows, N, verbosity)
...
-    override && (return nothing)
+function _check_measure(measure, model, y, operation)
 
     T = scitype(y)
 
     T == Unknown && (return nothing)
-    target_scitype(measure) == Unknown && (return nothing)
-    prediction_type(measure) == :unknown && (return nothing)
+    target_scitype(measure) == Unknown && (return true)
+    prediction_type(measure) == :unknown && (return true)
```
OkonSamuel, line 353:
```suggestion
    prediction_type(measure) == Unknown && (return true)
```

src/resampling.jl
OkonSamuel: A warning is fine by me

src/resampling.jl
```diff
@@ -518,8 +524,8 @@ untouched.
 - `rows` - vector of observation indices from which both train and
   test folds are constructed (default is all observations)
 
-- `weights` - per-sample weights for training and measures; see
-  important note above
+- `weights` - per-sample weights for measures (not to be confused with
+  weights used in training)
```
OkonSamuel, line 528: The summary of recently added fields (`report_per_fold` and `fitted_params_per_fold`) need to be added

OkonSamuel: something like
```julia
-`fitted_params_per_fold`: a vector containing `fitted pamarms(mach)` for each    
  individual train fold fit of `mach`. 

- `report_per_fold`: a vector containing `report(mach)` for each 
   individual train fold fit of `mach`.
```

ablaom: Good catch.

src/resampling.jl
ablaom: No, actually this is dumb but correct: https://github.com/alan-turing-institute/MLJModelInterface.jl/blob/master/src/model_traits.jl#L44

Agreed, that types rather than symbols would be better values for this trait. But this is has to start at MLJModelInterface and would break the whole stack.

src/resampling.jl
ablaom: Is that a yes?

src/resampling.jl
```diff
@@ -968,17 +971,27 @@ function MLJBase.fit(resampler::Resampler, verbosity::Int, args...)
 
     mach = machine(resampler.model, args...)
 
-    weights, measures =
-        _process_weights_measures(resampler.weights, resampler.measure,
-                                  mach, resampler.operation,
-                                  verbosity, resampler.check_measure)
+    measures =
```
OkonSamuel, line 974: The docstring at https://github.com/alan-turing-institute/MLJBase.jl/pull/413/files#diff-a577eec5e9a815ef20410dd488df700bR919 has to be updated to reflect new behaviour

src/resampling.jl
OkonSamuel: Okay

src/resampling.jl
OkonSamuel: Yes. A warning is better

src/resampling.jl
ablaom: Fixed just now.

OkonSamuel:
looks good

ablaom:
@OkonSamuel Thanks for your review and separate PR.

PR #418, opened 2020-08-28
added logpdf.(u, v) and logpdf.(u, raw) methods

OkonSamuel:
I feel that PR #411 left out some methods. I feel these methods are needed since we want `logpdf` to have similar functionality with `pdf` the difference being just the `log`.
So this PR Adds `logpdf.(u::UniFinArr{S,V,R,P,N}, v::AbstractArray{<:CategoricalValue{V,R},N})` and 
 `logpdf.(u::UniFinArr{S,V,R,P,N}, raw::Union{V,AbstractArray{V,N}})` methods.

Feel free to close this if u feel these method are not needed.

### DIFF
--- a/src/univariate_finite/arrays.jl
+++ b/src/univariate_finite/arrays.jl
@@ -107,8 +107,9 @@ for func in [:pdf, :logpdf]
 end
 
 
-
+##
 ## PERFORMANT BROADCASTING OF pdf and logpdf
+##
 
 # u - a UnivariateFiniteArray
 # cv - a CategoricalValue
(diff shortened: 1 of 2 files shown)


DilumAluthge:
Makes sense to me.

@cscherrer Can you review this?

cc: @ablaom

src/univariate_finite/arrays.jl
```diff
@@ -181,6 +164,27 @@ function Base.Broadcast.broadcasted(
...
+for typ in [:CategoricalValue, :(AbstractArray{<:CategoricalValue{V,R},N}), :(Union{V,AbstractArray{V,N}})]
+ eval(quote function Base.Broadcast.broadcasted(
+         ::typeof(logpdf),
+         u::UniFinArr{S,V,R,P,N},
+         cv::$typ) where {S,V,R,P,N}
+
+        # Start with the pdf
+        result = pdf.(u,cv)
+
+        # Take the long of each entry in-place
```
cscherrer, line 179: Typo from my original comment :)

```suggestion
        # Take the log of each entry in-place
```

OkonSamuel:
@cscherrer. I made some changes. The previous `logpdf` methods mutated the original array used in construction of `UnivariateFinite`

cscherrer:
Sorry, I don't understand. Computing the PDF allocates an array that's never used again, right? Why not reuse that memory?

cscherrer:
Oh, or does `pdf.(...)` not allocate with the broadcasting optimization?

OkonSamuel:
> Oh, or does `pdf.(...)` not allocate with the broadcasting optimization?

Yes. In some cases it doesn't allocate, while in other instances it does allocate. I just made a commit that takes both cases into consideration

ablaom:
@OkonSamuel Thanks for this PR. Looks good to me.

OkonSamuel:
@ablaom. I need to discuss some further changes i wanted to make to this PR with you before this was merged.

ablaom:
No, I would agree this is not desirable. I would welcome a PR.

Issue #416, opened 2020-08-27
Disintegration of MLJBase (discussion and tracking issue)


ablaom:
It seems there is a case to be made for further modularisation of MLJBase. These all sound reasonable to me but happy to hear other ideas:

- [ ] Separate StatisticalMeasures package 

- [ ] Separate UnivariateFinite package

- [ ] Separate MLJComposition package

The benefits of the second has come in Soss.jl integration but also for models wanting to specify `UnivariateFinite` distribution priors as hyper-parameters (originally `BayesianLDA` was this way). The third because it's so big.

ablaom:
pre-compiling MLJBase on my newish  macbook pro just too 37seconds 😢

DilumAluthge:
For easy discoverability, maybe a consistent naming scheme as such:

1. MLJComposition.jl
2. MLJDistributions.jl - would include UnivariateFinite; other distributions could be added later
3. MLJMeasures.jl - would include the statistical measures
4. MLJSerialization.jl

Of course, these packages (e.g. MLJDistributions.jl) can definitely be used outside of MLJ.

DilumAluthge:
Anyway, this sounds like a great idea.

DilumAluthge:
Of course, the hard part is keeping all the dependencies, etc. in sync. Tools like CompaHelper can help. There are probably additional tools that we can develop to make development and maintenance easier.

ablaom:
Thanks for the feedback! 

Others have criticised the adoption of the "MLJ" prefix for packages that are really meant for more general consumption. (MLJScientificTypes is a case in point - but we already had ScientificTypes for the light version.) The criticism is that developers see "MLJ" and think that means the package is only for use with MLJ packages. I think there's validity in this. 

> Of course, the hard part is keeping all the dependencies, compat entries, etc. in sync. Tools like CompaHelper can help. There are probably additional tools that we can develop to make development and maintenance easier.

Yes, these things help.

DilumAluthge:
> Others have criticised the adoption of the "MLJ" prefix for packages that are really meant for more general consumption. (MLJScientificTypes is a case in point - but we already had ScientificTypes for the light version.) The criticism is that developers see "MLJ" and think that means the package is only for use with MLJ packages. I think there's validity in this.

Yeah that's definitely a concern.

I guess we should identify which subparts of MLJBase are useful in general, and which parts are really only useful in the MLJ context.

[...]

Issue #299, opened 2020-05-11
Improvements to measures


ablaom:
The measures part of MLJBase could do with some TLC. It is not the shiniest part of the MLJ code base, written in a bit of a hurry because nothing much could go forward without *something* in place, and the existing packages came up short.

I think the API is more-or-less fine, but the way things are implemented is less that ideal, leading to:

(i) code redundancy 
(ii) less functionality: measures that could support weights or implement `reports_each_observation` don't

Recall that a measure `reports_each_observation` means `m(v1, v2)` returns a *vector* of measurements, and otherwise a single scalar is returned. So it does't really make sense for `auc`, for example, to `report_each_observation` (which it doesn't). However, `mae` should (but doesn't). 

I propose we make the following assumption that will allow us to resolve these issues for the majority of measures:

**If a measure `m(v1, v2)` implements `reports_each_observation` then it is understood that it is the sum or mean value of some scalar version `m(s1, s2)`.** 

For such measures, then, we need only implement the scalar method `m(s1, s2)` and we can generate the other methods `m(v1, v2)`, `m(v1, v2, w)` automatically. 

For other measures, such as `auc` and the `rms` family, `m(v1, v2)` (and optionally `m(v1, v2, w)`) must be explicitly implemented, as at present. 

In addition to the [docs](https://alan-turing-institute.github.io/MLJ.jl/dev/performance_measures/) there is a lot about the measure design in [this discussion](https://github.com/alan-turing-institute/MLJ.jl/issues/450).

### Details

To "automatically generate" the extra methods, we could do something like this:

```julia
# fallbacks for measures
(m::Measure)(yhat, y::AbstractVector) = _eval(Val(reports_each_observation), m, yhat, y)
(m::Measure)(yhat, y::AbstractVector, w) = _eval(Val(reports_each_observation), m, yhat, y, w)
_eval(::Val{false}, m, args...) = m(yhat, args...)
_eval(::Val{true}, m, y, yhat) = (m(yhat, y)) |> aggregation(m)
_eval(::Val{true}, m, y, yhat, w) = w .* (m(yhat, y)) |> aggregation(m)
```
@tlienart 
@azev77

ablaom:
Decision:

What do we do about measures like `mape` where we want to drop some terms where the computation is unstable? That is, what do we do if we are now reporting a value for every observation. 

One option is return a `missing` there and make sure the aggregators do `skipmissing`. There is no `skipnan`, which might be more natural.

[...]

PR #421, opened 2020-09-02
Log cosh

ven-k:
- Log-Cosh gives losses similar to MSE, but isn't much affected by occasional wild predictions. 
- It can be differentiated twice
- Keras has similar implementation.
- It's computed by 
`log_cosh = sum (log (cosh (yhat - y))) / n`


src/measures/continuous.jl
```diff
@@ -320,3 +320,33 @@ function (m::MAPE)(ŷ::Vec{<:Real}, y::Vec{<:Real})
...
+"""
+const log_cosh = LogCosh()
+
+metadata_measure(LogCosh;
+    name                     = "log_cosh",
+    target_scitype           = Union{Vec{Continuous},Vec{Count}},
+    prediction_type          = :deterministic,
+    orientation              = :loss,
+    reports_each_observation = false,
+    is_feature_dependent     = false,
```
ablaom, line 345: Suggestion (see **C** below): `reports_each_observation = true`.

src/measures/continuous.jl
```diff
@@ -320,3 +320,33 @@ function (m::MAPE)(ŷ::Vec{<:Real}, y::Vec{<:Real})
...
+    supports_weights         = false,
+    docstring                = "LogCosh; aliases: `log_cosh`.")
+
+function (log_cosh::LogCosh)(ŷ::Vec{<:Real}, y::Vec{<:Real})
+    check_dimensions(ŷ, y)
+    return sum(log.(cosh.(ŷ-y))) / length(y)
+end
```
ablaom, line 352: Suggestion (see **C** below):  return a vector of per-observation losses: `return log.(cosh.(ŷ-y))) / length(y)`

test/measures/continuous.jl
```diff
@@ -12,6 +12,7 @@ rng = StableRNG(666899)
     @test isapprox(mean(l1(yhat, y, w)), mae(yhat, y, w))
     @test isapprox(mean(l2(yhat, y)), 5)
     @test isapprox(mean(l2(yhat, y, w)), rms(yhat, y, w)^2)
+    @test isapprox(log_cosh(yhat, y), 1.3715546675)
```
ablaom, line 16: Modify test accordingly.

ablaom (commented):
@ven-k Thanks for this contribution and for taking the time to wrap your head around the API!

**C**

As the loss is just the mean of a per-observation loss, can I suggest the changes above? By default this loss will be aggregated in resampling using the mean (`aggregation(...) = Mean()`). In this way the per-observation losses will be available, for example, to hyper-parameter optimisation strategies that use this information (some Bayesian methods). 

So then the code will look basically the same as `l1` which has this behaviour:

```julia
julia> l1([1,2,3], [2,3,4])
3-element Array{Int64,1}:
 1
 1
 1
```

src/measures/continuous.jl
OkonSamuel: are we sure he needs `reports_each_observation = true` since he said the `Log-Cosh` loss is similar to `rms` and because of the definition in his comment as `log_cosh = sum (log (cosh (yhat - y))) / n`. 
To me it looks like he meant to set `reports_each_observation = false`

src/measures/continuous.jl
OkonSamuel: The `sum` is missing in this definition

src/measures/continuous.jl
ven-k: @OkonSamuel  `mean(log(cosh(yhat-y)))` returns value similar to MSE (it's what Tensorflow returns). But as @ablaom mentioned in the suggestion **C**, returning vector will help in tuning of hyperparameters, so I've modified accordingly. 
PS: The definition is `Log_Cosh(x) = log(cosh(x))`

ven-k: @OkonSamuel  `mean(log(cosh(yhat-y)))` returns value similar to MSE (it's what Tensorflow returns). But as @ablaom mentioned in the suggestion **C**, returning vector will help in tuning of hyperparameters, so I've modified accordingly. 
PS: The definition is `Log_Cosh(x) = log(cosh(x))`

OkonSamuel: Oh. I get you now

OkonSamuel (approved):

src/measures/continuous.jl
```diff
@@ -320,3 +320,33 @@ function (m::MAPE)(ŷ::Vec{<:Real}, y::Vec{<:Real})
...
 end
+
+struct LogCosh <: Measure end
+
+"""
+    LogCosh(ŷ, y)
+
+Log-Cosh loss:
+
+``\\text{Log-Cosh} = m^{-1}∑ᵢ log(cosh(ŷᵢ-yᵢ))``
```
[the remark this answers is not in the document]
ven-k: Done

ablaom (approved):
@OkonSamuel Yeah it is a bit confusing. Some measures that could be implemented as per-observation losses are not, eg, `mae` (which I guess essentially duplicates `l1`, which *is* per-observation loss). 

@ven-k  For the record, measures is overdue for a thorough review and  migration out of MLJ https://github.com/alan-turing-institute/MLJBase.jl/issues/299 .

ablaom:
@ven-k  Now live (see above)

ven-k:
Thanks! @ablaom and @OkonSamuel 

I am following #299 and #416. I am interested in StatisticalMeasures and possible visualisation package.

PR #429, opened 2020-10-01
Tweaks to the model API for supporting machine serialization

ablaom:
This PR makes tweaks to the (experimental and unpublicised) API for models with special serizialization requirements. It is not breaking for users. I have made it in response to failures of the existing API to handle the [XGBoost serialization issue](https://github.com/alan-turing-institute/MLJ.jl/issues/512).

The proposed API is outlined in the draft documentation given below, for adding to ["Adding New Models for General Use"](https://alan-turing-institute.github.io/MLJ.jl/dev/adding_models_for_general_use/#Supervised-models-1). As a POC I have implemented and locally tested the new API at a forthcoming [external glue code package](https://github.com/alan-turing-institute/MLJXGBoostInterface.jl) for XGBoost. 

Very roughly, the PR enables: (i) automatic generation of model-specific serialization formats **in addition to** a self-contained MLJ/JLSO format; (ii) a workaround for "non-persistent" representations of learned parameters, such as pointers generated by wrapped C code, provided the model-providing package already implements some kind of serialization.

@yalwan-iqvia @OkonSamuel Be great to get your feedback. 

I will leave this open for comment until Monday 11th.

---
# Draft documentation
### Serialization 

The MLJ user can serialize and deserialize a *machine*, which means
serializing/deserializing:

- the associated `Model` object (storing hyperparameters)
- the `fitresult` (learned parameters)
- the `report` generating during training

These are bundled into a single file or `IO` stream specified by the
user using the package `JLSO`. There are two scenarios in which a new
MLJ model API implementation will want to overload two additional
methods `save` and `restore` to support serialization:

1. The algorithm-providing package already has it's own serialization
  format for learned parameters and/or hyper-parameters, which users
  may want to access. In that case *the implementation overloads* `save`.
  
2. The `fitresult` is not a sufficiently persistent object; for
  example, it is a pointer passed from wrapped C code. In that case
  *the implementation overloads* `save` *and* `restore`.
  
In case 2, 1 presumably holds also, for otherwise MLJ serialization is
probably not going to be possible without changes to the
algorithm-providing package. An example is given below.

Note that in case 1, MLJ will continue to create it's own
self-contained serialization of the machine. Below `filename` refers
to the corresponding serialization file name, as specified by the
user, but with any final extension (e.g., ".jlso", ".gz") removed. If
the user has alternatively specified an `IO` object for serialization,
then `filename` is a randomly generated numeric string.

#### The save method

```julia
MMI.save(filename, model::SomeModel, fitresult; kwargs...) -> serializable_fitresult
```

Implement this method to serialize using a format specific to models
of type `SomeModel`. The `fitresult` is the first return value of
`MMI.fit` for such model types; `kwargs` is a list of keyword
arguments specified by the user and understood to relate to a some
model-specific serialization (cannot be `format=...` or
`compression=...`). The value of `serializable_fitresult` should be a
persistent representation of `fitresult`, from which a correct and
valid `fitresult` can be reconstructed using `restore` (see
below). 

The fallback of `save` performs no action and returns `fitresult`.

#### The restore method

```julia
MMI.restore(filename, model::SomeModel, serializable_fitresult) -> fitresult
```

Implement this method to reconstruct a `fitresult` (as returned by
`MMI.fit`) from a persistent representation constructed using
`MMI.save` as described above. 

The fallback of `restore` returns `serializable_fitresult`.

#### Example

Below is an example drawn from MLJ's XGBoost wrapper. In this example
the `fitresult` returned by `MMI.fit` is a tuple `(booster,
a_target_element)` where `booster` is the `XGBoost.jl` object storing
the learned parameters (essentially a pointer to some object created
by C code) and `a_target_element` is an ordinary `CategoricalValue`
used to track the target classes (a persistent object, requiring no
special treatment).

```julia
function MLJModelInterface.save(filename,
                                ::XGBoostClassifier,
                                fitresult;
                                kwargs...)
    booster, a_target_element = fitresult

    xgb_filename = string(filename, ".xgboost.model")
    XGBoost.save(booster, xgb_filename)
    persistent_booster = read(xgb_filename)
    @info "Additional XGBoost serialization file \"$xgb_filename\" generated. "
    return (persistent_booster, a_target_element)
end

function MLJModelInterface.restore(filename,
                                   ::XGBoostClassifier,
                                   serializable_fitresult)
    persistent_booster, a_target_element = serializable_fitresult

    xgb_filename = string(filename, ".tmp")
    fitresult =  open(xgb_filename, "w") do file
        write(file, persistent_booster)
    end
    booster = XGBoost.Booster(model_file=xgb_filename)
    rm(xgb_filename)
    fitresult = (booster, a_target_element)
    return fitresult
end
```


src/machines.jl
```diff
@@ -596,6 +596,17 @@ report(mach::Machine) = mach.report
 
 ## SERIALIZATION
 
+# helper:
+_filename(file::IO) = string(rand(UInt))
+function _filename(file::String) # truncates extension if present
```
yalwan-iqvia, line 601: What exactly is the objective of this function? Return just the filename without extension? without the leading path?
I'm trying to understand why  you haven't used the built in `splitext` for this (and `basename` if you want a filename without its dir)

src/interface/model_api.jl
```diff
@@ -36,16 +36,9 @@ MLJModelInterface.implemented_methods(::FI, M::Type{<:MLJType}) =
...
-MLJModelInterface.save(file, model, fitresult, report; kwargs...) =
-    JLSO.save(file,
-              :model => model,
-              :fitresult => fitresult,
-              :report => report; kwargs...)
-function MLJModelInterface.restore(file; kwargs...)
-    dict = JLSO.load(file)
-    return dict[:model], dict[:fitresult], dict[:report]
-end
+MLJModelInterface.save(filename, model, fitresult; kwargs...) = fitresult
```
yalwan-iqvia, line 39: So just to be clear, this is fallback to an asumed serialisable function and if this doesnt work the method implementer overrides this to produce a serialisable fitresult?

src/machines.jl
OkonSamuel: Yeah splittext would be better

OkonSamuel (changes requested):
Clever.
LGTM

OkonSamuel (approved):

src/machines.jl
yalwan-iqvia: It doesn't need to. we can keep the multiple dispatch structure and use splitext for a string and the existing implementation for IO

ablaom: Good suggestion, thanks.

src/interface/model_api.jl
ablaom: Yes, although I would say seriasable *object* not *function*. It may not be a function (and usually isn't).

ablaom:
@OkonSamuel @yalwan-iqvia Thanks for this feedback.

src/interface/model_api.jl
yalwan-iqvia: Of course, thanks for clarification

Issue #383, opened 2020-07-27
Add multi class precision/recall


tlienart:
(and use those in #382)

https://machinelearningmastery.com/precision-recall-and-f-measure-for-imbalanced-classification/

ablaom:
@ven-k has made a PR for this: https://github.com/alan-turing-institute/MLJBase.jl/pull/430

Issue #382, opened 2020-07-27
Add multiclass F1/Fb score


tlienart:
Allowing for a parameter to specify how the average is taken see

https://scikit-learn.org/stable/modules/generated/sklearn.metrics.f1_score.html

ablaom:
#456

PR #430, opened 2020-10-06
Multi class support to FScore, Precision, Recall and Weighted rates

ven-k:
This adds:
- Multi-class support to FScore, Precision, Recall, TP, FP, TN, FN
- Weighted TPR, TNR, FPR, FNR, FDR and NPV

Yet to add:
- Tests


ven-k:
This solves #382 and #383. 

Once the changes are reviewed, I would like to edit the test file which currently tests against few modified functions.

ablaom:
@ven-k Many thanks for this!  I've been on leave and just catching up. Just want to let you know I appreciate the PR and that it is on my radar. 

If @tlienart is not able to find time to review in the next week or so, I will take a look.

src/measures/finite.jl
```diff
@@ -539,29 +540,38 @@ metadata_measure.((FalsePositive, FalseNegative);
...
     prediction_type          = :deterministic,
     orientation              = :score,
     reports_each_observation = false,
     aggregation              = Sum(),
     is_feature_dependent     = false,
     supports_weights         = false)
 
-metadata_measure.((TruePositiveRate, TrueNegativeRate, Precision, NPV);
-    target_scitype           = Vec{<:Finite{2}},
+metadata_measure.(( TrueNegativeRate, NPV);
```
tlienart, line 559: (nit) remove the space before True

src/measures/finite.jl
```diff
@@ -539,29 +540,38 @@ metadata_measure.((FalsePositive, FalseNegative);
...
+    target_scitype           = Vec{<:Finite{N}} where N,
     prediction_type          = :deterministic,
     orientation              = :score,
     reports_each_observation = false,
     is_feature_dependent     = false,
-    supports_weights         = false)
+    supports_weights         = true)
+
+metadata_measure.((TruePositiveRate,Precision);
```
tlienart, line 567: (nit) add a space after comma

src/measures/finite.jl
```diff
@@ -607,11 +617,12 @@ MMI.docstring(::Type{<:Precision})    = "positive predictive value "*
...
 
-    true_positive(ŷ, y)
+    true_positive(ŷ, y; average="micro")
 
 Number of true positives for observations `ŷ` and ground truth
-`y`. Assigns `false` to first element of `levels(y)`. To reverse roles,
-use `TruePositive(rev=true)` instead of `true_positive`.
+`y`. In `binary` cases, assigns `false` to first element of `levels(y)`. To
+reverse roles, use `TruePositive(rev=true)` instead of `true_positive`.
+`average` takes `micro` and `macro` as arguements.
```
tlienart, line 625: (nit) `arguments` 

(not nit) could you please add a clarification as to what these arguments do (i.e. describe the kind of averaging briefly). 

Same later on. Also given it's repeated. Just define a `const DS_MICRO_MACRO` or similar and use it in the docstring `$DS_MICRO_MACRO` so that it can be maintained in only one spot.

Also I think I'd prefer it if it was symbols that were used rather than strings (`:micro`, `:macro`) this implicitly carries the information that there's only a few admitted values.

src/measures/finite.jl
```diff
@@ -820,57 +841,155 @@ const PPV = Precision
 
 ## INTERNAL FUNCTIONS ON CONFUSION MATRIX
 
-_tp(m::CM2) = m[2,2]
-_tn(m::CM2) = m[1,1]
-_fp(m::CM2) = m[2,1]
-_fn(m::CM2) = m[1,2]
+_tp(m::CM) = diag(m.mat)
+_fp(m::CM) = sum(m.mat, dims=2).-diag(m.mat)
```
tlienart, line 845: (nit) please use space around operator

src/measures/finite.jl
```diff
@@ -820,57 +841,155 @@ const PPV = Precision
...
+_fn(m::CM) = sum(m.mat, dims=1)' .- diag(m.mat)
+_tn(m::CM) = sum(m.mat) .- (_tp(m).+_fp(m).+_fn(m))
```
tlienart, line 847: same here

src/measures/finite.jl
```diff
@@ -474,8 +474,7 @@ end
 # ==========================================================================
 ## BINARY AND ORDER DEPENDENT
 
-const CM2 = ConfusionMatrix{2}
-
+const CM = ConfusionMatrix{N} where N
```
tlienart, line 477: make this a parametric type `CM{N}` or something so that you can extract the number of classes later on in computations

src/measures/finite.jl
```diff
@@ -820,57 +841,155 @@ const PPV = Precision
...
-_tn(m::CM2) = m[1,1]
-_fp(m::CM2) = m[2,1]
-_fn(m::CM2) = m[1,2]
+_tp(m::CM) = diag(m.mat)
+_fp(m::CM) = sum(m.mat, dims=2).-diag(m.mat)
+_fn(m::CM) = sum(m.mat, dims=1)' .- diag(m.mat)
+_tn(m::CM) = sum(m.mat) .- (_tp(m).+_fp(m).+_fn(m))
+
+function _tpr(m::CM; average = "micro")
+    s = size(m.mat)[1]
```
tlienart, line 850: see comment earlier, you should extract this from the type `CM{N}` 

also please call this `n_classes` (or similar) to make it clearer

src/measures/finite.jl
```diff
@@ -820,57 +841,155 @@ const PPV = Precision
...
+    s>2 && average == "macro" || return sum(tp(m)) / (sum(tp(m))+sum(fn(m)))
```
tlienart, line 851: (nit) space around operators

src/measures/finite.jl
```diff
@@ -820,57 +841,155 @@ const PPV = Precision
...
+    tpr = zeros(Float64, s)
+    tp_val, fn_val = tp(m), fn(m)
+    for i in range(1, length=s)
+        tpr[i] = tp_val[i] / (tp_val[i]+fn_val[i])
+    end
```
tlienart, line 856:
```julia
tpr = [ai/(ai+bi) for (ai, bi) in zip(tp_val, fn_val)]
```

src/measures/finite.jl
```diff
@@ -820,57 +841,155 @@ const PPV = Precision
...
+    return tpr
+end
+
+function _tpr(m::CM, w::Vec{<:Real}; average = "macro")
+    if average == "micro" throw("The mode has to be `macro`") end
```
tlienart, line 861: `average == "micro" && throw ... ` also make the error message clearer why does it have to be macro

src/measures/finite.jl
```diff
@@ -820,57 +841,155 @@ const PPV = Precision
...
+    _tpr(m, average = average) .* w
+end
+
+function _tnr(m::CM; average = "micro")
+    s = size(m.mat)[1]
+    s>2 && average == "macro" || return sum(tn(m)) / (sum(tn(m))+sum(fp(m)))
```
tlienart, line 867: similar comments to earlier.

src/measures/finite.jl
```diff
@@ -820,57 +841,155 @@ const PPV = Precision
...
+    tnr = zeros(Float64, s)
+    tn_val, fp_val = tn(m), fp(m)
+    for i in range(1, length=s)
+        tnr[i] = tn_val[i] / (tn_val[i]+fp_val[i])
+    end
```
tlienart, line 872: here too

src/measures/finite.jl
```diff
@@ -820,57 +841,155 @@ const PPV = Precision
...
+    return tnr
+end
+
+function _tnr(m::CM, w::Vec{<:Real}; average = "macro")
+    if average == "micro" throw("The mode has to be `macro`") end
```
tlienart, line 877: here too

src/measures/finite.jl
```diff
@@ -820,57 +841,155 @@ const PPV = Precision
...
+    _tnr(m, average = average) .* w
+end
+
+_fpr(m::CM; average = "micro") = 1 .- _tnr(m, average = average)
```
tlienart, line 881: for kw assignment, there's typically no space around the `=` so `_fpr(m::CM{N}; average=:micro) where N = ...`

src/measures/finite.jl
```diff
@@ -820,57 +841,155 @@ const PPV = Precision
...
+          (m::M)(ŷ, y, w::Vec{<:Real}, average = "macro") = m(confmat(ŷ, y; rev=m.rev), w; average = average)
+end
+
+#= for M in (TPR, TNR, FPR, FNR, FDR, Precision, NPV, FScore)
+          (m::M)(ŷ, y; average = "micro") = m(confmat(ŷ, y; rev=m.rev); average = average)
+          (m::M)(ŷ, y; w::Vec{<:Real}, average = "macro") = m(confmat(ŷ, y; rev=m.rev), w; average = average)
+end
+throws UndefVarError while, when individually written works just fine. For example:
+(m::TPR)(ŷ, y; average = "micro") = m(confmat(ŷ, y; rev=m.rev); average = average)=#
+
```
tlienart, line 986: do these comments still apply?

tlienart:
I did a first pass, looks pretty good I would say, thanks! Once these details have been ironed out and there are good tests, I think it'd be good if @ablaom you had a final look as it's quite a big change overall.

src/measures/finite.jl
ven-k: I've found a work-around.
```
for T in (F1, F2)
    @eval (f::$T)(a; b=value) = #do something
 end
```
does the trick. So it's resolved.

src/measures/finite.jl
ven-k: Will it be better if
 ```
average == :micro && @warn ("As `w` specifies the weights for each class, 
         promoting average to `:macro`")
```
is used instead of throwing an error?

Also, as `macro` is a keyword, should I rename `:micro` & `:macro` as `:overall` & `:foreach` (or `:for_each` or `:per_val`) ?

File: src/measures/finite.jl [hunk 16]
```diff
@@ -625,11 +636,12 @@ const truepositive  = TruePositive()
 
 $(docstring(TrueNegative()))
 
-    true_negative(ŷ, y)
+    true_negative(ŷ, y; average="micro")
```
ablaom, line 640: We never pass keyword arguments to the evaluation of a measure. The interface point for parameters is at construction of the measure. So, the user does something like:

```julia
tn = TrueNegative(average=:micro) # construction
tn(yhat, y)                                            # evaluation
```
and so `true_negative` is then just an alias for `TrueNegative() = TrueNegative(average=<default>)`. If you haven't done so, you will need to make `average` a field of the `TrueNegative` struct.

src/measures/finite.jl
ablaom: No "Macro F1" seems very standard.

ablaom: Regarding the error vs warning question. I wonder if this is the right place for this logic. This is a private method. Shouldn't this logic be in the ConfusionMatrix constructor? I vote for error. And I think we should be clear in documentation to be clear about distinguishing between *sample* weights (what a user passes when they do `tp(yhat, y, w)`) and *class* weights (what the user will specify when constructing an f1-score, confusion matrix, and so forth).

ablaom (commented):
@ven-k Okay, I've done a first pass on this PR.

I can see a lot of work has gone into this. Here are some comments:

1. I'm guessing there may be some **confusion between *sample* weights and *class* weights.** If `supports_weights` is true for a measure, this means the user can specify a real weight *for each observation* at the time the measure is evaluated. Sample weights are specified at evaluation time and are `Real` vectors with the same length as the target. So if the user does `accuracy(yhat, y, w)` then `w` has the same length as `y` and `yhat`. (Note `w` is not a keyword arg. Similarly, if a *model* supports weights, one can do `machine(model, X, y, w).`) Of course, sample weights are more general because they make sense  for regression too. So what should we do about class weights, for the special case of classification (`Finite` target)? The API does not currently specify this, but I guess it would make sense to pass class weights in the same way, namely when the measure is called (and not when it is instantiated) as you have done here. However, it would be prudent to introduce a new trait `supports_class_weights`, set to true for your generalized metrics, and to set `supports_weights` to false. How does that sound?  An important question is what representation for `class_weights` should we use? A vector of reals won't do in the unordered case, so perhaps this should be a dictionary keyed on the class ("raw" value or `CategoricalValue`). We could use `UnivariateFinite` - our distribution for finite labelled sample spaces - but that would force the weights to be normalised and introduce a code dependency we probably want to avoid later on. I vote for `AbstractDict`.  

2. **User interface point for specifying the `average` option** This should be specified at *construction* of the measure (whose struct has  `average` as a field) not at *evaluation* of the measure on data. See my comment about `TrueNegative`. So, we are handling `average` the same way we handle `\beta` in the `F_\beta` score.

3. In the generalization `true_negative(yhat, y)`, etc, becomes a *vector*, even in the binary case.  Apart from this being a breaking change, this will not be palatable to users who worry mostly about binary classification. What was your plan for dealing with this?

ablaom:
Okay, I reviewed my understanding of `F1-score` and realise that ordinary "binary F1-score" is different from "n-ary  F1-score" with n = 2. In particular, the n-ary scores are always invariant under re-labelling of the classes (in the unweighted macro case) but this is not true for binary the F1-score - which class is declared "true" matters. In the sk-learn implementation, the distinction is made by adding a ":binary" option to `average` which I find very confusing. These are really different beasts and should not be conflated, in my opinion. 

My suggestion, then, would to be to leave the existing measures as is, and introduce the "generalization" (maybe "extension" is more accurate word) by introducing new types and methods. 

@tlienart @ven-k What do you think? Or am I missing something here?

https://en.wikipedia.org/wiki/F1_score
https://arxiv.org/pdf/1911.03347.pdf

ablaom:
I also think the distinction between the case is useful at the level of types because the usual f1-score, strictly speaking, applies only to `OrderedFactor{2}`, while the other one also applies to unordered targets (`Multiclass{2}`).

ven-k:
1. I will work on `class_weights` (instead of `w`) of `AbstractDict` type. 
But `supports_class_weights = true` and `aggregator = Mean()` parameters are applicable only when `average=:macro`, which I doubt `metadata_measure` has any easy way to indicate.
So I propose to create a `metadata_measure` for `DS_MICRO_MACRO` struct and add these 2 traits in it. 
(And, if necessary have a new trait `supports_micro_macro = true` for FScore, Precision...) 

2. My understanding is redefining (even source) struct throws error. So this was a work-around I employed.

3. I wanted to know if I should allow `average=:binary` mode or another parameter, inspired from sk-learn, `label` which specifies classes that are considered to be as the true values. Ex: For classes [0,1,2], if `label = [1, 2]` then, whenever yhat = 0 for y = 1 or 2 is considered wrong prediction while yhat = 1 for y = 2 and vice-versa are ignored (or are unexpected)
However I don't think `label` is a good idea for above reason.


Because of 2, Generalization would require renaming every function (maybe with verbose names). 
I can think of two things:
- To not have a default value for `average` (again because of 2),  and also, add `:binary` mode to handle `OrderedFactor{2}`, while `micro`, `macro` would handle `Multiclass{N}`. This will cause confusion; a clear doc might alleviate.
OR
- Have a wrapper module for `OrderedFactor` and `Multiclass`. 
Ex: `of_tn = OrderedFactor.TrueNegative()` and `mc_tn = Multiclass.TrueNegative()`
(I am not quite sure how good this solution can be or if this is allowed in the API)

ablaom:
1. Good point about the different properties of the different `average` options. However, rather than introuducing even more traits, I'd rather we break up the measures. So, we have `FScoreMicro` (`Finite`) and `FScoreMacro` (`Finite`) and `FScore` keeps it's original meaning (`OrderedFactor{2}`) only). 

2. I don't the problem with users re-constructing the measure for a different option. But my suggestion in 1. makes this point moot.

3. "Generalization would require renaming every function (maybe with verbose names)" . I think this is fine. I think a long list of measures is easier for the user to navigate than combining different kinds of functionality within the same object. This also creates fewer complicated case-distinctions in other parts of MLJ.  So we have `FScoreMacro` and `FScoreMicro`. We keep `TruePositiveRate` (`OrderedFactor{2}`) and so forth and add `MulticlassTruePositiveRate`, etc (`Finite`). Probably the general  user will never use `Mutliclass*` so it's long name doesn't really matter. I may not be the best judge but I am guessing `labels` is not that useful and could be dropped. Under my suggestion of leaving the current sructs/methods alone, we don't need `:binary`.

ven-k:
Sure. I'll work accordingly.

ven-k:
Things done:

- Reverted the Scores for ordered case with N = 2.
- Added `MulticlassFScore`, `MulticlassPrecision`, `MulticlassRecall` and all related callables for N >= 2 that ordered metrics supported for N = 2.
- Added `:supports_class_weights::Bool = false` parameter to metadata_measure. And have used `class_w` for all class weights.
- average modes 
 Sklearn   | MLJBase
                              ` "micro"`     | `:micro`
                               `"macro"`   | `:macro |> Mean`
                               `"None"`     | `:macro`

test/measures/finite.jl
```diff
@@ -175,6 +175,31 @@ end
...
+                                                    [0.0; 0.5; 1.1428571428]
+
+    micro_prec = MulticlassPrecision(average=:micro)
+    micro_rec  = MulticlassRecall(average=:micro)
+    micro_f1score = MulticlassFScore{1}(average=:micro)
+    @test micro_prec(cm) == micro_prec(ŷ, y) == 0.5
+    @test micro_rec(cm) == micro_rec(ŷ, y) == 0.5
+    @test micro_f1score(cm) == micro_f1score(ŷ, y) == 0.5
+end
+
```
ablaom, line 202: Awesome.

src/measures/finite.jl
```diff
@@ -1229,73 +1231,74 @@ _mfp(m::CM{N}) where N = sum(m.mat, dims=2) .- diag(m.mat)
 _mfn(m::CM{N}) where N = sum(m.mat, dims=1)' .- diag(m.mat)
 _mtn(m::CM{N}) where N = sum(m.mat) .- (_mtp(m) .+ _mfp(m) .+ _mfn(m))
 
+@inline function _class_w(level_m::Vec{<:Real}, class_w::AbstractDict{<:Real, <:Real})
+    @assert levels(keys(class_w)) == level_m "Labels of `y` and `class_w` must match."
+    return [class_w[v] for v in levels(keys(class_w))]
```
ablaom, line 1236: Please use `if ... throw(...) ...end` or similar in place of `@assert`. 

User may not know what "class_w" and "y" mean, as they are not input as keyword arguments. Perhaps better message is "Encountered target with levels different from the keys of user-specified dictionary of class weights. " or similar.

src/measures/finite.jl
```diff
@@ -1086,7 +1088,7 @@ $(docstring(MulticlassTruePositiveRate()))
     multiclass_true_positive_rate(ŷ, y, class_w)
 
 True positive rate for multiclass observations `ŷ` and ground truth `y`.
-$DS_MICRO_MACRO `MulticlassTruePositiveRate(; average=:micro)`.
+$DS_MICRO_MACRO `MulticlassTruePositiveRate(; average=:micro)`.$CLASS_W
```
ablaom, line 1092: A single space is needed after the period in line 1091, or in the beginning of the constant $CLASS_W.

ablaom (commented):
@ven-k Thanks for this very substantial update. Awesome! 

I think this is very close to finished. 

I noticed that although `macro_f1score` (for example) is for any `Finite` target, we get warning if we use `Multiclass` data instead of `OrderedFactor`. I think this  is because `macro_f1score` ultimately calls `confmat` on the data. You can do `confmat(yhat, h, warn=false)` to suppress the warning.  This applies to multiclass precision and recall as well.

ablaom:
You have typed the class weights dictionaries to have key type `<: Real` but these keys could have any type. Consider the following error:

```julia
big = coerce([:male, :female, :binary, :male,
              :male, :binary, :female, :male], OrderedFactor);
y = big[1:4];
yhat = big[5:8];

m = Dict(:male => 1, :female => 2, :binary => 4)
# or user might use CategoricalValue's:
# m = Dict(big[1] => 1, big[2] => 2, big[3] => 3)

julia> multiclass_false_positive_rate(yhat, y, m)
ERROR: MethodError: no method matching (::MulticlassFalsePositiveRate{Symbol})(::CategoricalArrays.CategoricalArray{Symbol,1,UInt32,Symbol,CategoricalArrays.CategoricalValue{Symbol,UInt32},Union{}}, ::CategoricalArrays.CategoricalArray{Symbol,1,UInt32,Symbol,CategoricalArrays.CategoricalValue{Symbol,UInt32},Union{}}, ::Dict{Symbol,Int64})
Closest candidates are:
  Any(::Any, ::Any) at /Users/anthony/Dropbox/Julia7/MLJ/MLJBase/src/measures/finite.jl:1358
  Any(::Any, ::Any, ::AbstractDict{var"#s577",var"#s576"} where var"#s576"<:Real where var"#s577"<:Real) at /Users/anthony/Dropbox/Julia7/MLJ/MLJBase/src/measures/finite.jl:1359
Stacktrace:
 [1] top-level scope at REPL[46]:1
```

src/measures/finite.jl
```diff
@@ -871,6 +871,502 @@ end
...
+
+## INTERNAL FUCNTIONS ON MULTICLASS CONFUSION MATRIX
+
+_mtp(m::CM{N}) where N = diag(m.mat)
+_mfp(m::CM{N}) where N = sum(m.mat, dims=2) .- diag(m.mat)
+_mfn(m::CM{N}) where N = sum(m.mat, dims=1)' .- diag(m.mat)
+_mtn(m::CM{N}) where N = sum(m.mat) .- (_mtp(m) .+ _mfp(m) .+ _mfn(m))
+
+@inline function _class_w(level_m::Vec{<:String}, class_w::AbstractDict{<:Any, <:Real})
+    string.(levels(keys(class_w))) == level_m || throw(ArgumentError(W_KEY_MISMATCH))
```
OkonSamuel, line 1238:
```suggestion
   labels = string.(levels(keys(class_w)))
    labels == level_m || throw(ArgumentError(W_KEY_MISMATCH))
     return [class_w[v] for v in labels]
```

src/measures/finite.jl
OkonSamuel:
```suggestion
    return [class_w[v] for v in levels(keys(class_w))]
```

src/measures/finite.jl
```diff
@@ -871,6 +871,502 @@ end
...
+    return [class_w[v] for v in levels(keys(class_w))]
```
OkonSamuel, line 1239:
```suggestion
    return [class_w[v] for v in labels]
```

src/measures/finite.jl
OkonSamuel: avoid calling `string.(levels(keys(class_w)))` twice. see suggestion above

ven-k: Yes. I will change it to `class_w_labels` (to avoid any confusion with `labels` of confusion matrix.) 

level_m is always a `Vec{<:String}` (conf-mat property). So we don't need to check.

Additionally, checking if class_w_labels are already string came out to be same as casting on testing `@time`, so even that would be unnecessary.

src/measures/finite.jl
```diff
@@ -871,6 +871,502 @@ end
...
+"""
+const multiclass_npv                       = MulticlassNPV()
+const multiclass_negative_predictive_value = MulticlassNPV()
+const multiclass_negativepredictive_value  = MulticlassNPV()
+const MNPV                                 = MulticlassNPV
+
+## INTERNAL FUCNTIONS ON MULTICLASS CONFUSION MATRIX
+
+_mtp(m::CM{N}) where N = diag(m.mat)
+_mfp(m::CM{N}) where N = sum(m.mat, dims=2) .- diag(m.mat)
```
OkonSamuel, line 1233:
```suggestion
_mfp(m::CM{N}) where N = sum(m.mat, dims=2) .-= diag(m.mat)
```

src/measures/finite.jl
ven-k: This can't happen as string(keys) can't be used to get the value of class_w.

src/measures/finite.jl
OkonSamuel: @ven-k i just updated my suggestion.

src/measures/finite.jl
OkonSamuel: @ven-k Yeah your right. I've updated my suggestion accordingly

src/measures/finite.jl
ven-k: The idea is to sum the entire row (all predicted positives) and subtract diagonal (the true positive).

OkonSamuel: @ven-k Yeah but my suggestion does the same thing avoiding an extra allocation.

ven-k: That's really cool. I'll wait if @ablaom suggests any other changes and will update _mfp, _mfn accordingly in a single commit.

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+
+"""
+const multiclass_npv                       = MulticlassNPV()
+const multiclass_negative_predictive_value = MulticlassNPV()
+const multiclass_negativepredictive_value  = MulticlassNPV()
+const MNPV                                 = MulticlassNPV
+
+## INTERNAL FUCNTIONS ON MULTICLASS CONFUSION MATRIX
+
+_mtp(m::CM{N}) where N = diag(m.mat)
```
OkonSamuel, line 1232:
```suggestion
_mtp(m::CM) = diag(m.mat)
```
It is known in julia that using `CM{N}` instead of `CM` slightly affects performance when `N` isn't used in the function definition. Hence the above suggestion.

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+
+function _mnpv(m::CM{N}, class_w::AbstractDict{<:Any, <:Real}; average=:macro) where N
+    average == :micro && @warn W_PROMOTE_WARN
+    level_w = _class_w(m.labels, class_w)
+    return _mnpv(m, average=average) .* level_w
+end
+
+## CALLABLES ON MULTICLASS CONFUSION MATRIX
+
+(::MulticlassTruePositive)(m::CM{N}) where N  = _mtp(m)
```
OkonSamuel, line 1310:
```suggestion
(::MulticlassTruePositive)(m::CM) = _mtp(m)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+(::MulticlassTrueNegative)(m::CM{N}) where N  = _mtn(m)
```
OkonSamuel, line 1311:
```suggestion
(::MulticlassTrueNegative)(m::CM) = _mtn(m)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+(::MulticlassFalsePositive)(m::CM{N}) where N = _mfp(m)
```
OkonSamuel, line 1312:
```suggestion
(::MulticlassFalsePositive)(m::CM)= _mfp(m)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+(::MulticlassFalseNegative)(m::CM{N}) where N = _mfn(m)
```
OkonSamuel, line 1313:
```suggestion
(::MulticlassFalseNegative)(m::CM) = _mfn(m)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+
+(r::MTPR)(m::CM{N}) where N = _mtpr(m, average=r.average)
```
OkonSamuel, line 1315:
```suggestion
(r::MTPR)(m::CM)= _mtpr(m, average=r.average)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+(r::MTPR)(m::CM{N}, w::AbstractDict{<:Any, <:Real}) where N = _mtpr(m, w, average=r.average)
```
OkonSamuel, line 1316:
```suggestion
(r::MTPR)(m::CM, w::AbstractDict{<:Any, <:Real}) = _mtpr(m, w, average=r.average)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+
+(r::MTNR)(m::CM{N}) where N = _mtnr(m, average=r.average)
```
OkonSamuel, line 1318:
```suggestion
(r::MTNR)(m::CM) = _mtnr(m, average=r.average)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+(r::MTNR)(m::CM{N}, w::AbstractDict{<:Any, <:Real}) where N = _mtnr(m, w, average=r.average)
+
+(r::MFPR)(m::CM{N}) where N = _mfpr(m, average=r.average)
+(r::MFPR)(m::CM{N}, w::AbstractDict{<:Any, <:Real}) where N = _mfpr(m, w, average=r.average)
```
OkonSamuel, line 1322:
```suggestion
(r::MFPR)(m::CM, w::AbstractDict{<:Any, <:Real}) = _mfpr(m, w, average=r.average)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+
+(r::MFNR)(m::CM{N}) where N = _mfnr(m, average=r.average)
```
OkonSamuel, line 1324:
```suggestion
(r::MFNR)(m::CM)= _mfnr(m, average=r.average)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+(r::MFNR)(m::CM{N}, w::AbstractDict{<:Any, <:Real}) where N = _mfnr(m, w, average=r.average)
```
OkonSamuel, line 1325:
```suggestion
(r::MFNR)(m::CM, w::AbstractDict{<:Any, <:Real}) = _mfnr(m, w, average=r.average)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+
+(r::MFDR)(m::CM{N}) where N = _mfdr(m, average=r.average)
```
OkonSamuel, line 1327:
```suggestion
(r::MFDR)(m::CM)  = _mfdr(m, average=r.average)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+(r::MFDR)(m::CM{N}, w::AbstractDict{<:Any, <:Real}) where N = _mfdr(m, w, average=r.average)
+
+(v::MNPV)(m::CM{N}) where N = _mnpv(m, average=v.average)
+(v::MNPV)(m::CM{N}, w::AbstractDict{<:Any, <:Real}) where N = _mnpv(m, w, average=v.average)
```
OkonSamuel, line 1331: remove the `N` parameter here also.

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+
+(p::MulticlassPrecision)(m::CM{N}) where N  = 1.0 .- _mfdr(m, average=p.average)
```
OkonSamuel, line 1333:
```suggestion
(p::MulticlassPrecision)(m::CM)  = 1.0 .- _mfdr(m, average=p.average)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+function (p::MulticlassPrecision)(m::CM{N}, class_w::AbstractDict{<:Any, <:Real}) where N
```
OkonSamuel, line 1334:
```suggestion
function (p::MulticlassPrecision)(m::CM, class_w::AbstractDict{<:Any, <:Real})
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+    average == :micro && @warn W_PROMOTE_WARN
+    level_w = _class_w(m.labels, class_w)
+    return (1.0 .- _mfdr(m, average=p.average)) .* level_w
+end
+
+function (f::MulticlassFScore{β})(m::CM{N}) where β where N
```
OkonSamuel, line 1340:
```suggestion
function (f::MulticlassFScore{β})(m::CM) where β
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+    β2   = β^2
+    rec  = MulticlassRecall(; average=f.average)(m)
+    prec = MulticlassPrecision(; average=f.average)(m)
+    f.average == :micro && return (1 + β2) * (prec * rec) / (β* prec + rec)
+    return (1 + β2) .* (prec .* rec) ./ (β2 .* prec .+ rec)
+end
+
+function (f::MulticlassFScore{β})(m::CM{N}, class_w::AbstractDict{<:Any, <:Real}) where β where N
```
OkonSamuel, line 1348:
```suggestion
function (f::MulticlassFScore{β})(m::CM, class_w::AbstractDict{<:Any, <:Real}) where β
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+
+function _mnpv(m::CM{N}; average=:micro) where N
+    n_classes = N
+    mtn_val, mfn_val = mtn(m), mfn(m)
+    average == :micro && return sum(mtn_val) / (sum(mtn_val) + sum(mfn_val))
+    return mtn_val ./ (mtn_val + mfn_val)
+end
+
+
+function _mnpv(m::CM{N}, class_w::AbstractDict{<:Any, <:Real}; average=:macro) where N
```
OkonSamuel, line 1302:
```suggestion
function _mnpv(m::CM, class_w::AbstractDict{<:Any, <:Real}; average=:macro)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+end
+
+function _mfdr(m::CM{N}; average=:macro) where N
+    n_classes = N
+    mtp_val, mfp_val = mtp(m), mfp(m)
+    average == :micro && return sum(mfp_val) / (sum(mtp_val) + sum(mfp_val))
+    return mfp_val ./ (mtp_val + mfp_val)
+end
+
+function _mfdr(m::CM{N}, class_w::AbstractDict{<:Any, <:Real}; average=:macro) where N
```
OkonSamuel, line 1288:
```suggestion
function _mfdr(m::CM, class_w::AbstractDict{<:Any, <:Real}; average=:macro)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+end
+
+function _mtnr(m::CM{N}; average=:macro) where N
+    n_classes = N
+    mtn_val, mfp_val = mtn(m), mfp(m)
+    average == :micro && return sum(mtn_val) / (sum(mtn_val) + sum(mfp_val))
+    return mtn_val ./ (mtn_val + mfp_val)
+end
+
+function _mtnr(m::CM, class_w::AbstractDict{<:Any, <:Real}; average=:macro) where N
```
OkonSamuel, line 1263:
```suggestion
function _mtnr(m::CM, class_w::AbstractDict{<:Any, <:Real}; average=:macro)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+end
+
+function _mtpr(m::CM{N}; average=:macro) where N
+    n_classes = N
+    mtp_val, mfn_val = mtp(m), mfn(m)
+    average == :micro && return sum(mtp_val) / (sum(mtp_val) + sum(mfn_val))
+    return mtp_val ./ (mtp_val + mfn_val)
+end
+
+function _mtpr(m::CM{N}, class_w::AbstractDict{<:Any, <:Real}; average=:macro) where N
```
OkonSamuel, line 1250:
```suggestion
function _mtpr(m::CM, class_w::AbstractDict{<:Any, <:Real}; average=:macro)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+    return mtn_val ./ (mtn_val + mfp_val)
+end
+
+function _mtnr(m::CM, class_w::AbstractDict{<:Any, <:Real}; average=:macro) where N
+    average == :micro && @warn W_PROMOTE_WARN
+    level_w = _class_w(m.labels, class_w)
+    return _mtnr(m, average=average) .* level_w
+end
+
+_mfpr(m::CM{N}; average=:macro) where N = 1 .- _mtnr(m, average=average)
```
OkonSamuel, line 1269:
```suggestion
_mfpr(m::CM; average=:macro) = 1 .- _mtnr(m, average=average)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+function _mfpr(m::CM{N}, class_w::AbstractDict{<:Any, <:Real}; average=:macro) where N
```
OkonSamuel, line 1270:
```suggestion
function _mfpr(m::CM, class_w::AbstractDict{<:Any, <:Real}; average=:macro)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+    level_w = _class_w(m.labels, class_w)
+    return (1 .- _mtnr(m, average=average)) .* level_w
+end
+
+_mfnr(m::CM{N}; average=:macro) where N = 1 .- _mtpr(m, average=average)
```
OkonSamuel, line 1275:
```suggestion
_mfnr(m::CM; average=:macro) = 1 .- _mtpr(m, average=average)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+function _mfnr(m::CM{N}, class_w::AbstractDict{<:Any, <:Real}; average=:macro) where N
```
OkonSamuel, line 1276:
```suggestion
function _mfnr(m::CM, class_w::AbstractDict{<:Any, <:Real}; average=:macro)
```

src/measures/finite.jl
```diff
@@ -871,6 +871,503 @@ end
...
+const multiclass_negative_predictive_value = MulticlassNPV()
+const multiclass_negativepredictive_value  = MulticlassNPV()
+const MNPV                                 = MulticlassNPV
+
+## INTERNAL FUCNTIONS ON MULTICLASS CONFUSION MATRIX
+
+_mtp(m::CM{N}) where N = diag(m.mat)
+_mfp(m::CM{N}) where N = sum(m.mat, dims=2) .- diag(m.mat)
+_mfn(m::CM{N}) where N = sum(m.mat, dims=1)' .- diag(m.mat)
+_mtn(m::CM{N}) where N = sum(m.mat) .- (_mtp(m) .+ _mfp(m) .+ _mfn(m))
```
OkonSamuel, line 1235:
```suggestion
function _mtn(m::CM)
d = diag(m.mat) #diagonal of confusion matrix
sum_rows = sum(m.mat, dims=2)
sum_cols = transpose(sum(m.mat, dims=1))
# The following code has the same effect 
# as `sum(m.mat) .- (_mtp(m) .+ _mfp(m) .+ _mfn(m))` but 
# more efficient and stores the result in `sum_dim2`.
sum_dim2 .= sum(m.mat) .- (sum_rows .+ sum_cols .-  d)
```

src/measures/finite.jl
ven-k: @OkonSamuel on `@benchmark`ing
```
julia> @benchmark p(z)
BenchmarkTools.Trial: 
  memory estimate:  79.08 KiB
  allocs estimate:  3
  --------------
  minimum time:     6.817 μs (0.00% GC)
  median time:      7.660 μs (0.00% GC)
  mean time:        8.810 μs (11.57% GC)
  maximum time:     162.265 μs (87.96% GC)
  --------------
  samples:          10000
  evals/sample:     5

julia> @benchmark q(z)
BenchmarkTools.Trial: 
  memory estimate:  79.08 KiB
  allocs estimate:  3
  --------------
  minimum time:     16.470 μs (0.00% GC)
  median time:      17.570 μs (0.00% GC)
  mean time:        19.235 μs (6.30% GC)
  maximum time:     893.854 μs (96.28% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

for 
```
a = rand(100, 100)
p(a) = sum.(a) .- MLJBase.diag(a)
q(a) = sum.(a) .-= MLJBase.diag(a)
```
So, shall it remain the same?

src/measures/finite.jl
ven-k: They yield same `@benchmark` performance, I believe both are identical, as CM = ConfusionMatrix{N} where N.

I tried with  `CM_ = ConfusionMatrix`, which slightly performs worse.

```
mfp(m::CM{N}) where N = sum(m.mat, dims=2) .- diag(m.mat)
_mfp2(m::CM_) = sum(m.mat, dims=2) .- diag(m.mat)

julia> @benchmark MLJBase._mfp2(cm)
BenchmarkTools.Trial: 
  memory estimate:  368 bytes
  allocs estimate:  8
  --------------
  minimum time:     781.610 ns (0.00% GC)
  median time:      799.080 ns (0.00% GC)
  mean time:        861.941 ns (4.63% GC)
  maximum time:     52.653 μs (97.38% GC)
  --------------
  samples:          10000
  evals/sample:     100

julia> @benchmark MLJBase._mfp(cm)
BenchmarkTools.Trial: 
  memory estimate:  368 bytes
  allocs estimate:  8
  --------------
  minimum time:     778.592 ns (0.00% GC)
  median time:      794.233 ns (0.00% GC)
  mean time:        851.195 ns (4.51% GC)
  maximum time:     50.301 μs (97.59% GC)
  --------------
  samples:          10000
  evals/sample:     103
```

src/measures/finite.jl
ven-k: Cool! 
`_mtn(m::CM{N}) where N = sum(m.mat) .- (sum(m.mat, dims=2) .+ sum(m.mat, dims=1)' .+ diag(m.mat))` slashes time by half. (I've slightly modified your suggestion to compact it) I'll update in the next commit.

src/measures/finite.jl
OkonSamuel: yeah. You can leave it as it is?

src/measures/finite.jl
OkonSamuel: You can stick to using ` CM = ConfusionMatrix{N} where N`

src/measures/finite.jl
ven-k: Yes it turns out for `N<31`,  `.-=` performs much better. I've added a ternary operator to choose between `.-=` and `-=` 
(PS: The operator didn't cost anything additionl on @benchmark)

src/measures/finite.jl
ven-k: However, I did drop `where N` wherever the code could be made compact, later when noticed.

src/measures/finite.jl
```diff
@@ -871,6 +871,505 @@ end
...
+const multiclass_npv                       = MulticlassNPV()
+const multiclass_negative_predictive_value = MulticlassNPV()
+const multiclass_negativepredictive_value  = MulticlassNPV()
+const MNPV                                 = MulticlassNPV
+
+## INTERNAL FUCNTIONS ON MULTICLASS CONFUSION MATRIX
+
+_mtp(m::CM) = diag(m.mat)
+_mtn(m::CM)= sum(m.mat) .- (sum(m.mat, dims=2) .+ sum(m.mat, dims=1)' .+ diag(m.mat))
+_mfp(m::CM{N}) where N = N < 31 ? sum(m.mat, dims=2) .-= diag(m.mat) :
```
OkonSamuel, line 1234: To avoid error on julia v 1.0 try
```suggestion
_mfp(m::CM{N}) where N = (col_sums = sum(m.mat, dims=2); N < 31 ? col_sums.-= diag(m.mat) :
                                                                                                                  col_sums .- diag(m.mat) )    
```

src/measures/finite.jl
```diff
@@ -871,6 +871,505 @@ end
...
+                                  sum(m.mat, dims=2) .- diag(m.mat)
+_mfn(m::CM{N}) where N = N < 31 ? sum(m.mat, dims=1)' .-= diag(m.mat) :
```
OkonSamuel, line 1236: same applies here to avoid build error on julia v1.0

src/measures/finite.jl
OkonSamuel: I forgot to mention. Your not bench-marking properly.
try.
```julia
@benchmark p($z)
@benchmark q($z)
```
with the same definitions of `p` and `q`.

src/measures/finite.jl
OkonSamuel: Again the code for bench-marking should be
```julia
julia> @benchmark MLJBase._mfp2($cm)
julia> @benchmark MLJBase._mfp($cm)
```

src/measures/finite.jl
ven-k: Yeah, I figured from logs; was trying to see if I could keep it 1 lined even in Julia 1.0 somehow. (Had to download it)

But, thank you for so quickly going through it and suggesting!

OkonSamuel: @ven-k could you try bench-marking again with my suggestion [here](https://github.com/alan-turing-institute/MLJBase.jl/pull/430#discussion_r514157641) to see if we could drop the conditional statement?

src/measures/finite.jl
```diff
@@ -871,6 +871,511 @@ end
...
+"""
+const multiclass_npv                       = MulticlassNPV()
+const multiclass_negative_predictive_value = MulticlassNPV()
+const multiclass_negativepredictive_value  = MulticlassNPV()
+const MNPV                                 = MulticlassNPV
+
+## INTERNAL FUCNTIONS ON MULTICLASS CONFUSION MATRIX
+
+_mtp(m::CM) = diag(m.mat)
+_mtn(m::CM) = sum(m.mat) .- (sum(m.mat, dims=2) .+ sum(m.mat, dims=1)' .+ diag(m.mat))
```
OkonSamuel, line 1233: Your code still has unnecessary allocations
```suggestion
function _mtn(m::CM)
    col_sum = sum(m.mat, dims=2)
    col_sum .= sum(m.mat) .- ( col_sum .+ sum(m.mat, dims=1)' .+ diag(m.mat))
```
The way julia broadcasting work is that it first allocates a suitable container for the output then fuses several operations and stores the result into the output container. 
Already `sum(m.mat, dims=2)` and `sum(m.mat, dims=1)` allocates output container which can hold the final result.

ablaom:
Okay, I have a new suggestion. I apologise for not raising this earlier. It concerns the rates `mutliclass_true_positive_rates`, etc. As I understand it, they currently return a *vector* for each observation, ordered according to `levels(y)`, which forces us, for the sake of unambiguous interpretation, to restrict to `OrderedFactor` targets, even though they make sense more generally (any `Finite`) (unlike the non-multiclass version).  A similar issue came up in [Fairness metrics](https://github.com/ashryaagr/Fairness.jl) and in the end we went for dictionaries keyed on class instead. I know dictionaries are less performant (`LittleDict`s are slightly better for small numbers of classes - the main use case) but I think usability/unambiguity trumps performance in this case. (Zygote-based AD would (currently) be an issue, and for all the other measures predicting UnivariateFinite, but hopefully these kind of limitations of Zygote are ultimately addressed or we go for a different AD approach.) We could predict tuples, but that could be weird for certain class label types (even integers are a problem). 

@ven-k @tlienart @ashryaagr @OkonSamuel Do you see any issues with returning dictionaries?

ven-k:
If the idea is to use the same `confmat` and return `LittleDict(m.labels, multiclass_vector))` it does hurt the performance by atleast 0.1 to 0.2 μs. (We can however introduce a `mode=:dict` for all multiclass functions which also takes `:vec` to determine the format user wants (It doesn't cost a lot than what the required format usually takes; approx 10 ns). 

But I do want to know if `OrderedFactor{N}` where N>2 is encountered frequently, as we might have to consider renaming these functions as `FiniteFScore` and so on if it's high. (Or should I rename it anyways to eliminate any ambiguity?)

ablaom:
> If the idea is to use the same confmat and return LittleDict(m.labels, multiclass_vector)) it does hurt the performance by atleast 0.1 to 0.2 μs. (We can however introduce a mode=:dict for all multiclass functions which also takes :vec to determine the format user wants (It doesn't cost a lot than what the required format usually takes; approx 10 ns).

@ven-k That sounds like a good compromise, assuming you're happy to implement it. Maybe we can think of a more informative name for the new field; instead of `mode` maybe `return_type`? With possible values `Vector` and `LittleDict` to make dispatch a little less cumbersome? 

By the way, if we have dictionaries, we will need to overload the calling for `Mean` and `Sum` (see https://github.com/alan-turing-institute/MLJBase.jl/blob/c58bcfc9853a18e8168d54d471228cdf21b341cf/src/measures/measures.jl#L37) so that aggregation works for dictionary-valued measures.  (In the case of confusion matrices we instead overloaded Base.+ and Base.round, but that would be piracy in our case.) @ashryaagr Did you already do this in Fairness.jl?

> But I do want to know if OrderedFactor{N} where N>2 is encountered frequently, as we might have to consider renaming these functions as FiniteFScore and so on if it's high. (Or should I rename it anyways to eliminate any ambiguity?)

Good point about the name! I also thought about this, but worried that "Finite" might be a bit cryptic for new users. And, as you say, the Multiclass case is probably the more important use case. What's better: less confusion - or better discoverability?Probably the latter but you go ahead and make the call, unless someone else has a strong opinion.

src/measures/finite.jl
ven-k: Yes. The benchmarks go neck to neck for large numbers.

But, on analyzing locally, I found `.-=` is not compatible with Julia 1.0

src/measures/finite.jl
```diff
@@ -1230,16 +1230,11 @@ const MNPV                                 = MulticlassNPV
...
-
-function _mfp(m::CM{N}) where N
-    col_sum, diag_val = sum(m.mat, dims=2), diag(m.mat)
-    return N < 31 ? col_sum .-= diag_val : col_sum .- diag_val
-end
-
-function _mfn(m::CM{N}) where N
-    row_sum_t, diag_val = sum(m.mat, dims=1)', diag(m.mat)
-    return N < 31 ? row_sum_t .-= diag_val : row_sum_t .- diag_val
+_mfp(m::CM) = sum(m.mat, dims=2) .-= diag(m.mat)
```
OkonSamuel, line 1233: To avoid error on julia 1.0 you could use
```julia
_mfp(m::CM) = (col_sum = sum(m.mat, dims=2); col_sum .-= diag(m.mat))
```
 ```

src/measures/finite.jl
```diff
@@ -1230,16 +1230,11 @@ const MNPV                                 = MulticlassNPV
...
+_mfn(m::CM) = sum(m.mat, dims=1)' .-= diag(m.mat)
```
OkonSamuel, line 1234: To avoid error on julia 1.0 you could use
```julia
_mfn(m::CM) = (row_sum_t = transpose(sum(m.mat, dims=2)); row_sum_t .-= diag(m.mat))
```

src/measures/finite.jl
OkonSamuel: Yeah. I made a suggestion [here](https://github.com/alan-turing-institute/MLJBase.jl/pull/430#discussion_r515977488) that you could use to avoid julia v1.0 errors

OkonSamuel:
> Okay, I have a new suggestion. I apologise for not raising this earlier. It concerns the rates `mutliclass_true_positive_rates`, etc. As I understand it, they currently return a _vector_ for each observation, ordered according to `levels(y)`, which forces us, for the sake of unambiguous interpretation, to restrict to `OrderedFactor` targets, even though they make sense more generally (any `Finite`) (unlike the non-multiclass version). A similar issue came up in [Fairness metrics](https://github.com/ashryaagr/Fairness.jl) and in the end we went for dictionaries keyed on class instead. I know dictionaries are less performant (`LittleDict`s are slightly better for small numbers of classes - the main use case) but I think usability/unambiguity trumps performance in this case. (Zygote-based AD would (currently) be an issue, and for all the other measures predicting UnivariateFinite, but hopefully these kind of limitations of Zygote are ultimately addressed or we go for a different AD approach.) We could predict tuples, but that could be weird for certain class label types (even integers are a problem).
> 
> @ven-k @tlienart @ashryaagr @OkonSamuel Do you see any issues with returning dictionaries?

Assuming that potential Zygote AD issues would be resolve in the future then am happy with the extra 0.1 - 0.2μs for the same of ease of use by the user.( I guess we would have to make this a standard. To have a consistent interface).

  

> We can however introduce a `mode=:dict` for all multiclass functions which also takes `:vec` to determine the format user wants (It doesn't cost a lot than what the required format usually takes; approx 10 ns

Allowing the user choose is a nice idea. But we should set the default `return_type` option to `dict`. Instead of using `mode=:dict` which would introduce type instabilities . Why not create an `abstract type ModeOption end`. Then make `DictOption <: ModeOption` etc.

ablaom:
> Allowing the user choose is a nice idea. But we should set the default return_type option to dict. Instead of using mode=:dict which would introduce type instabilities (you could check by using @code_warntype . Why not create an abstract type ModeOption end. Then make DictOption <: ModeOption and so forth.

I think we achieve the same by making `return_type` a type-parameter with values `Vector` or `LittleDict` as possible values. Then no need to introduce a new type for user to remember.

ablaom:
@ven-k You still expecting to have some time to finish this off? Let me know if there's anything more you need from me.

Project.toml
```diff
@@ -55,6 +55,7 @@ Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 MultivariateStats = "6f286f6a-111f-5878-ab1e-185364afe411"
 NearestNeighbors = "b8a86587-4115-5ab1-83bc-aa920d37bbce"
+OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
```
ablaom, line 59: I think we can drop OrderedCollections in extras and targets as it is in [deps] (and needed there, no?)

src/measures/finite.jl
```diff
@@ -875,9 +875,19 @@ Base.precision(ŷ, y)   = confmat(ŷ, y) |> Precision()
...
+struct MacroAvg <: MulticlassAvg end
+struct MicroAvg <: MulticlassAvg end
 
-MulticlassFScore{β}(; average=:macro) where β = MulticlassFScore{β}(average)
+const macro_avg = MacroAvg()
+const micro_avg = MicroAvg()
+
+struct MulticlassFScore{β, M<:MulticlassAvg} <:Measure
+    average::M
+    return_type::Union{Type{Vec}, Type{LittleDict}}
```
ablaom, line 887: Is there a reason we do not make `return_type` a type parameter here, to avoid abstract field types?

src/measures/finite.jl
```diff
@@ -875,9 +875,19 @@ Base.precision(ŷ, y)   = confmat(ŷ, y) |> Precision()
...
+end
+
+MulticlassFScore{β}(; average=macro_avg, return_type=LittleDict) where β = MulticlassFScore{β, typeof(average)}(average, return_type)
```
ablaom, line 890: It's weird that I cannot do `Multiclass(beta=...)` and I know you have just kept this for consistency with  vanilla `FScore` (which has the same issue). Can we *add* a constructor 

`MulticlassFScore(; beta=1.0, average=macro_avg, ...)`

and, while your here, do the same for FScore?

test/measures/finite.jl
```diff
@@ -181,20 +181,23 @@ end
...
 
-    @test multiclass_precision(cm) ≈ multiclass_precision(ŷ, y) ≈ [0.4; 0.5; 0.666666667]
-    @test multiclass_recall(cm) ≈ multiclass_recall(ŷ, y) ≈ [0.5; 0.5; 0.5]
-    @test macro_f1score(cm) ≈ macro_f1score(ŷ, y) ≈ [0.444444444445; 0.5; 0.5714285714]
+    @test multiclass_precision(cm) == multiclass_precision(ŷ, y) ==
+            LittleDict("0"=>0.4, "1"=>0.5, "2"=>0.6666666666666667)
+    @test multiclass_recall(cm) == multiclass_recall(ŷ, y) ==
+            LittleDict("0"=>0.5, "1"=>0.5, "2"=>0.5)
+    @test macro_f1score(cm) == macro_f1score(ŷ, y) ==
+            LittleDict("0"=>0.4444444444444445, "1"=>0.5, "2"=>0.5714285714285715)
```
ablaom, line 189: Mmm. Comparing floats with `==`. Note that LittleDict's are actually ordered (unlike regular docs) so you can probably just compare `values` of both sides using `approx`.

test/measures/finite.jl
```diff
@@ -181,20 +181,23 @@ end
...
-    @test multiclass_recall(cm, class_w) ≈ multiclass_recall(ŷ, y, class_w) ≈
-                                                    [0.0; 0.5; 1.0]
-    @test macro_f1score(cm, class_w) ≈ multiclass_f1score(ŷ, y, class_w) ≈
-                                                    [0.0; 0.5; 1.1428571428]
+    @test multiclass_precision(cm, class_w) == multiclass_precision(ŷ, y, class_w) ==
+                        LittleDict(["0", "1", "2"], [0.0; 0.5; 1.3333333333333335])
+    @test multiclass_recall(cm, class_w) == multiclass_recall(ŷ, y, class_w) ==
+                        LittleDict(["0", "1", "2"], [0.0; 0.5; 1.0])
+    @test macro_f1score(cm, class_w) == multiclass_f1score(ŷ, y, class_w) ==
+                        LittleDict(["0", "1", "2"], [0.0; 0.5; 1.142857142857143])
```
ablaom, line 196: It's more helpful (and more convincing) if you write `8/7` instead of `1.142857...` etc. Then I know you actually worked this out by hand  😄

test/measures/finite.jl
ablaom, line 189: Could we have a `micro_f1score` test too?

ablaom (commented):
Wow! What a massive bunch of commits. Many thanks for taking so many suggestions on board.

Apart from addressing the code annotations - pretty minor, I think - we just need doc strings for  `MulticlassFScore` and `MulticlassPrecision` and `MulticlassRecall`, which can be based on the binary versions.

Project.toml
ven-k: The Travis threw an error and logs pointed to this. I pushed another commit with the line in the [extras].

test/measures/finite.jl
ven-k: I wanted to test both key and value so floating point number up-to last dp and `==`. I will test for `values` and repeat one for keys, maybe?

Project.toml
ablaom: Yeah - if it's in [targets] it needs to be in [extras] but I think we get rid of both as it's already in [deps]

src/measures/finite.jl
ven-k: `struct MulticlassFScore{β, M<:MulticlassAvg, U<:Union{Vec, LittleDict}} <:Measure` this threw error "can't convert type Type{LittleDict} to type LittleDict" while constructing functor with ` = MulticlassFScore{β, typeof(average), return_type}(average, return_type)` . So I did resort this.

ablaom: Okay, first, let's get rid of `beta` as a type parameter. I know vanilla FScore has it, but it really serves no purpose that I can see; it can be a regular field. (We can fix FScore in a later clean-up). Then the following works:

```julia
struct MulticlassFScore{M<:MulticlassAvg, U<:Union{Vec, LittleDict}} <:Measure
    beta::Float64
    average::M
    return_type::Type{U}
end

MulticlassFScore(beta,
                 average::M,
                 return_type::Type{U}) where {M,U<:Union{Vec,LittleDict}} =
    Multiclass{M,U}(beta, average, return_type)

MulticlassFScore(; beta=1.0, average=macro_avg, return_type=LittleDict) =
    MulticlassFScore(beta, average, return_type)

MulticlassFScore(1.0, macro_avg, LittleDict)
```

src/measures/finite.jl
```diff
@@ -882,12 +882,28 @@ struct MicroAvg <: MulticlassAvg end
 const macro_avg = MacroAvg()
 const micro_avg = MicroAvg()
 
+"""
+    MulticlassFScore{β}()
+
```
ablaom, line 887: Under suggestion below, this would become

```
"""
    MulticlassFScore(; beta=1.0, average=macro_avg, return_type=LittleDict)

```

src/measures/finite.jl
```diff
@@ -882,12 +882,28 @@ struct MicroAvg <: MulticlassAvg end
...
+One-parameter generalization, ``F_β``, of the F-measure or balanced F-score for
+multiclass observations.
+
+    MulticlassFScore()(ŷ, y)
```
ablaom, line 891: and 891 could be replaced with

```
    MulticlassFScore()(ŷ, y)
    MulticlassFScore()(ŷ, y, w)

Evaluate the default score on mutliclass .... Here `w` is an optional dictionary of weights, keyed on `levels(y)`. 

etc 
```

src/measures/finite.jl
```diff
@@ -882,12 +882,28 @@ struct MicroAvg <: MulticlassAvg end
...
+
```
ablaom, line 892: I think you meant to interpolate `DS_AVG_RET` somewhere in this docstring.

ablaom: You should also mention `MulticlassFScore()(ŷ, y, w)` and say "Here `w` is an optional dictionary of weights, keyed on `levels(y)`."

ablaom (approved):
@ven-k Awesome contribution. One of the largest volunteer contributions to MLJ 🎉 

One final note: With all this work, you are well versed to give feedback on this part of the code. Anything you have to suggest at https://github.com/alan-turing-institute/MLJBase.jl/issues/299 will be seriously considered and much appreciated. You can also comment on slack.

Again many thanks!

ablaom:
@ven-k Oops. I believe I missed some issues in my final review.

**1**

Can you look at the following code and fix the broken tests with a new PR onto dev. You may as well add these tests to the code:

```julia

using MLJBase
using Test

table = reshape(collect("aabbbccccddbabccbacccd"), 11, 2)
table = coerce(table, Multiclass);
yhat = table[:,1] # ['a', 'a', 'b', 'b', 'b', 'c', 'c', 'c', 'c', 'd', 'd']
y    = table[:,2] # ['b', 'a', 'b', 'c', 'c', 'b', 'a', 'c', 'c', 'c', 'd']

# class | TP | FP | TP + FP | precision | FN | TP + FN | recall
# ------|----|----|------------------------------------|-------
# a     | 1  | 1  | 2       | 1/2       | 1  | 2       | 1/2
# b     | 1  | 2  | 3       | 1/3       | 2  | 3       | 1/3
# c     | 2  | 2  | 4       | 1/2       | 3  | 5       | 2/5 
# d     | 1  | 1  | 2       | 1/2       | 0  | 1       | 1

# helper:
inverse(x) = 1/x
harmonic_mean(x...) = inverse(mean(inverse.(x)))

# precision:
p_macro = mean([1/2, 1/3, 1/2, 1/2])
# NOT RETURNING A NUMBER; need mean of dict values here
@test_broken MulticlassPrecision(average=macro_avg)(yhat, y) ≈ p_macro
p_micro = (1 + 1 + 2 + 1)/(2 + 3 + 4 + 2)
@test MulticlassPrecision(average=micro_avg)(yhat, y) ≈ p_micro

# recall:
r_macro = mean([1/2, 1/3, 2/5, 1])
# NOT RETURNING A NUMBER: need mean of dict values here
@test_broken MulticlassRecall(average=macro_avg)(yhat, y) ≈ r_macro
r_micro = (1 + 1 + 2 + 1)/(2 + 3 + 5 + 1)   
@test MulticlassPrecision(average=micro_avg)(yhat, y) ≈ r_micro

# fscore:
f1_macro = mean([harmonic_mean(1/2, 1/2),
                harmonic_mean(1/3, 1/3),
                harmonic_mean(1/2, 2/5),
                 harmonic_mean(1/2, 1)]) # mean of class-wise f1-scores
# NOT RETURNING A NUMBER: need mean of dict values here
@test_broken MulticlassFScore(average=macro_avg)(yhat, y) ≈ f1_macro
f1_micro = harmonic_mean(p_micro, r_micro) 
@test MulticlassFScore(average=micro_avg)(yhat, y) ≈ f1_micro
```

**2**

Can you explain why `average` is a field of `MulticlassTruePositive` (and similar)? I understand why this is so for *rates* but not the counts.

**3**

I believe the counts should return dictionaries by default, but they appear to be returning vectors still:

```julia
julia> MulticlassTruePositive(return_type=LittleDict)(yhat, y)
4-element Array{Int64,1}:
 1
 1
 2
 1
```

**4**

In retrospect, the `return_type` field only makes sense for F-score and rates (including precision and recall) if we provide an `average=nothing` option (meaning a dictionary or vector is returned), which we have not implemented.  Let's keep the field for now and add the implementation of `average=nothing` in a later PR, unless you have time to do it quickly now.

ablaom:
@ven-k  Okay, after merging this, I discovered some confusing things I missed before:

**confusion 1**

My understanding is that FScore should return a single number, irrespective of `average`, but this is not so:

```julia
import Random
Random.seed!(123)

using MLJBase

y    = coerce(rand("abc", 100), Multiclass);
yhat = coerce(rand("abc", 100), Multiclass);

# A single number, as expected:
julia> MulticlassFScore(average=micro_avg)(yhat, y)
0.42

# A dictionary, not expected:
julia> MulticlassFScore(average=macro_avg)(yhat, y)
OrderedCollections.LittleDict{String,Float64,Array{String,1},Array{Float64,1}} with 3 entries:
  "a" => 0.42623
  "b" => 0.363636
  "c" => 0.465753
```

**confusion 2**

```julia
# A dictionary, as expected:
julia> MulticlassTruePositiveRate()(yhat, y)
OrderedCollections.LittleDict{String,Float64,Array{String,1},Array{Float64,1}} with 3 entries:
  "a" => 0.464286
  "b" => 0.307692
  "c" => 0.515152

# Expected dictionary here but did not get one:
julia> MulticlassTruePositive()(yhat, y)
3-element Array{Int64,1}:
 13
 12
 17

```

**confusion 3**

I don't understand why `average` is a field in any of the new measures, apart from `MulticlassFScore`.

ablaom:
Also,

**confusion 4**

Specifying weights still does not work for case if classes are not `CategoricalValue{<:Real}`, it seems:

```julia
class_w = Dict('a'=>1, 'b'=>1, 'c'=>1, 'd'=>1)
julia> MulticlassFScore()(yhat, y, class_w)
ERROR: MethodError: no method matching _fs_helper(::MLJBase.ConfusionMatrix{4}, ::Dict{Char,Int64}, ::Float64, ::MLJBase.MacroAvg, ::Type{LittleDict})
Closest candidates are:
  _fs_helper(::MLJBase.ConfusionMatrix{N} where N, ::AbstractDict{var"#s577",var"#s576"} where var"#s576"<:Real where var"#s577"<:Real, ::Real, ::MLJBase.MacroAvg, ::Type{LittleDict}) at /Users/anthony/Dropbox/Julia7/MLJ/MLJBase/src/measures/finite.jl:1498
  _fs_helper(::MLJBase.ConfusionMatrix{N} where N, ::AbstractDict{var"#s577",var"#s576"} where var"#s576"<:Real where var"#s577"<:Real, ::Real, ::MLJBase.MicroAvg, ::Type{U}) where U at /Users/anthony/Dropbox/Julia7/MLJ/MLJBase/src/measures/finite.jl:1492
  _fs_helper(::MLJBase.ConfusionMatrix{N} where N, ::AbstractDict{var"#s577",var"#s576"} where var"#s576"<:Real where var"#s577"<:Real, ::Real, ::MLJBase.MacroAvg, ::Type{AbstractArray{T,1} where T}) at /Users/anthony/Dropbox/Julia7/MLJ/MLJBase/src/measures/finite.jl:1504
  ...
Stacktrace:
 [1] (::MulticlassFScore{MLJBase.MacroAvg,LittleDict})(::MLJBase.ConfusionMatrix{4}, ::Dict{Char,Int64}) at /Users/anthony/Dropbox/Julia7/MLJ/MLJBase/src/measures/finite.jl:1511
 [2] (::MulticlassFScore{MLJBase.MacroAvg,LittleDict})(::CategoricalArrays.CategoricalArray{Char,1,UInt32,Char,CategoricalArrays.CategoricalValue{Char,UInt32},Union{}}, ::CategoricalArrays.CategoricalArray{Char,1,UInt32,Char,CategoricalArrays.CategoricalValue{Char,UInt32},Union{}}, ::Dict{Char,Int64}) at /Users/anthony/Dropbox/Julia7/MLJ/MLJBase/src/measures/finite.jl:1530
 [3] top-level scope at REPL[231]:1
````
(continuation of code in previous comment)

ven-k:
Thanks for the all the suggestions @ablaom, @tlienart and @OkonSamuel. I've learnt quite a lot of Julia tricks. 

(Additionally, 'have realized that instead of having a large PR with too many moving parts, breaking it to couple of PRs is much necessary😃)

Resolving, 
Confusion 1: As MLJBase returns per-observation values for most loss functions, these functions return per-class values. Hence running `aggregation` for `average=macro_avg` is a necessity for MulticlassFScore and all the rates (and even `MulticlassTruePositive` and simliar once `LittleDict` dispatch is added). I could export `Mean` and `Sum` for Dict to make it crispier to use. 
Please suggest if you think I should, by default, return `Mean`. 

Confusion 2 and 3: Oops. I will add a dispatch for `LittleDict` and also get rid of `average` option.
Confusion 4: This too I missed. Will rectify.
(Sorry, somehow I didn't notice these three as things evolved)

(I'll PR ASA confusion 1 is clear)

About 4:
A. The idea was both `return_type`s always make sense for `macro_avg` (aka works like suggested `average=nothing`).
B.  And for `micro_avg` it has no effect at all for non-weighted classes. But it's helpful when class weights are passed. The average is upgraded to `macro_avg` and the specified `return_type` is returned. 

So the user will not have any unintended results in any of the cases. Do you still think `nothing` mode would be required? (It would make current `macro_avg` to `nothing` mode and `nothing, return_type=Vec` as the new `macro_avg`.)

ablaom:
@ven-k Very sorry, it seems I double-posted a comment (one was a draft, which I have since deleted). Can you just confirm you have read through https://github.com/alan-turing-institute/MLJBase.jl/pull/430#issuecomment-730085747 and this is what you are responding to?

ablaom:
@ven-k I think, if you wouldn't mind, we'd best have a call to sort things out a little quicker. (It would also give me great pleasure to meet you "in person"). Can you take a call between 6.30am and 8.30pm Auckland time, any day?

In the meantime:

> Confusion 1: As MLJBase returns per-observation values for most loss functions, these functions return per-class values. Hence running aggregation for average=macro_avg is a necessity for MulticlassFScore and all the rates (and even MulticlassTruePositive and simliar once LittleDict dispatch is added). 

Not quite sure I follow this - perhaps because you are responding to my deleted comment (😳 )?? Let me try to respond as best I can. Sorry if I go on a tangent. 

All of the new measures have `reports_each_observation = false`, yes? I'm happy with this. So, "aggregation" of the observations is all done internally and has nothing to do with the `aggregation` trait. What's the use of this trait? It tells other parts of MLJ how to aggregate this measure - say in resampling, such as cross-validation. If I have 5 F1-scores (from 5-fold cross-validation) - that is 5 *numbers* - how should I aggregate these to report an overall estimate of performance? Because of non-linearities, there is no "right" answer (in particular if the folds have different sizes), but computing the mean is a reasonable compromise, so `aggregation=Mean()` for `FScore()`. For a count, such as `TruePositive`, it makes more sense (assuming the test folds are non-overlapping and exhaustive) to use `aggregate=Sum` instead.

> I could export Mean and Sum for Dict to make it crispier to use.
Please suggest if you think I should, by default, return Mean.

I'm really lost about what this suggestion is resolving. Probably no action required here on  your part. 


My main confusion is that MulticlassFScore, MulticlassPrecision, MulticlassRecall currently returning one kind of object for micro averaging (a single number, just like sk-learn) but a Dictionary in the case of macro averaging (unlike sk-learn). This makes no sense to me at all. In the case of sk-learn, for example, a number is returned in both cases, and a vector of per-class values is only returned if `average=None`. From [sk-learn docs](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.f1_score.html):

Screen Shot 2020-11-20 at 10 01 30 AM

(I agree - I think you mentioned this before - that we throw an error if  `class_w` is passed and `average` is different from `macro_avg`. Our interface point for class weights is different from sk-learn.)

ablaom:
P.S. My Julia slack handle is the same as my github one and our channel is #mlj

ven-k:
#455 addresses all the issues we discussed.

PR #432, opened 2020-10-08
Support Distributions 0.24

devmotion:
The changes in Distributions 0.24 (more precisely, https://github.com/JuliaStats/Distributions.jl/pull/1171) require some additional dispatches to avoid method ambiguity errors (as seen e.g. in https://github.com/alan-turing-institute/MLJBase.jl/pull/431).

This PR supersedes https://github.com/alan-turing-institute/MLJBase.jl/pull/431.

### DIFF
--- a/src/univariate_finite/methods.jl
+++ b/src/univariate_finite/methods.jl
@@ -198,6 +198,14 @@ function _pdf(d::UnivariateFinite{S,V,R,P}, ref) where {S,V,R,P}
     return get(d.prob_given_ref, ref, zero(P))
 end
 
+function _pdf(d::UnivariateFinite{S,V}, c::V) where {S,V}
+    _classes = classes(d)
+    c in _classes || throw(DomainError("Value $c not in pool. "))
+    pool = CategoricalArrays.pool(_classes)
+    class = pool[get(pool, c)]
+    return pdf(d, class)
+end
+
 """
     Distributions.pdf(d::UnivariateFinite, x)
 
@@ -227,20 +235,12 @@ One can also do weighted fits:
 See also `classes`, `support`.
 """
 Distributions.pdf(d::UnivariateFinite, cv::CategoricalValue) = _pdf(d, int(cv))
-
-function Distributions.pdf(
-    d::UnivariateFinite{S,V,R,P}, c::V) where {S,V,R,P}
-
-    _classes = classes(d)
-    c in _classes || throw(DomainError("Value $c not in pool. "))
-    pool = CategoricalArrays.pool(_classes)
-    class = pool[get(pool, c)]
-    return pdf(d, class)
-end
+Distributions.pdf(d::UnivariateFinite{S,V}, c::V) where {S,V} = _pdf(d, c)
+Distributions.pdf(d::UnivariateFinite{S,V}, c::V) where {S,V<:Real} = _pdf(d, c)
 
 Distributions.logpdf(d::UnivariateFinite, cv::CategoricalValue) = log(pdf(d,cv))
-
-Distributions.logpdf(d::UnivariateFinite{S,V,R,P}, c::V) where {S,V,R,P} = log(pdf(d,c))
+Distributions.logpdf(d::UnivariateFinite{S,V}, c::V) where {S,V} = log(pdf(d, c))
+Distributions.logpdf(d::UnivariateFinite{S,V}, c::V) where {S,V<:Real} = log(pdf(d, c))
 
 function Distributions.mode(d::UnivariateFinite)
     dic = d.prob_given_ref
--- a/src/univariate_finite/methods.jl
+++ b/src/univariate_finite/methods.jl
@@ -193,19 +193,6 @@ function average(dvec::AbstractVector{UnivariateFinite{S,V,R,P}};
     return UnivariateFinite(sample_scitype(d1), d1.decoder, prob_given_ref)
 end
 
-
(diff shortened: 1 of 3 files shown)


src/univariate_finite/methods.jl
```diff
@@ -198,6 +198,14 @@ function _pdf(d::UnivariateFinite{S,V,R,P}, ref) where {S,V,R,P}
     return get(d.prob_given_ref, ref, zero(P))
 end
 
+function _pdf(d::UnivariateFinite{S,V}, c::V) where {S,V}
```
OkonSamuel, line 201:
```suggestion
function pdf_(d::UnivariateFinite{S,V}, c::V) where {S,V}
```
rename `_pdf` internal method to `pdf_` to avoid wrongly dispatching on the existing definition.
```julia
function _pdf(d::UnivariateFinite{S,V,R,P}, ref) where {S,V,R,P}
    return get(d.prob_given_ref, ref, zero(P))
end
```
This is shown by the following example.
```julia
julia> using MLJBase, CategoricalArrays
[ Info: Precompiling MLJBase [a7f614a8-145f-11e9-1d2a-a57a1082229d]

julia> y = UInt32[  0   0   0   1   0   1   1   0] |> vec |> categorical
8-element CategoricalArray{UInt32,1,UInt32}:
 0x00000000
 0x00000000
 0x00000000
 0x00000001
 0x00000000
 0x00000001
 0x00000001
 0x00000000

julia> d = UnivariateFinite(classes(y), [0.4, 0.6])
UnivariateFinite{Multiclass{2}}(0=>0.4, 1=>0.6)

julia> MLJBase._pdf(d, UInt32(1))
ERROR: DomainError with Value 2 not in pool. :

Stacktrace:
 [1] _pdf(::UnivariateFinite{Multiclass{2},UInt32,UInt32,Float64}, ::UInt32) at /home/okonsamuel/.julia/packages/MLJBase/BrLfz/src/univariate_finite/methods.jl:203
 [2] pdf(::UnivariateFinite{Multiclass{2},UInt32,UInt32,Float64}, ::CategoricalValue{UInt32,UInt32}) at /home/okonsamuel/.julia/packages/MLJBase/BrLfz/src/univariate_finite/methods.jl:237
 [3] _pdf(::UnivariateFinite{Multiclass{2},UInt32,UInt32,Float64}, ::UInt32) at /home/okonsamuel/.julia/packages/MLJBase/BrLfz/src/univariate_finite/methods.jl:206
 [4] top-level scope at REPL[4]:1
```

src/univariate_finite/methods.jl
```diff
@@ -227,20 +235,12 @@ One can also do weighted fits:
...
-    d::UnivariateFinite{S,V,R,P}, c::V) where {S,V,R,P}
-
-    _classes = classes(d)
-    c in _classes || throw(DomainError("Value $c not in pool. "))
-    pool = CategoricalArrays.pool(_classes)
-    class = pool[get(pool, c)]
-    return pdf(d, class)
-end
+Distributions.pdf(d::UnivariateFinite{S,V}, c::V) where {S,V} = _pdf(d, c)
+Distributions.pdf(d::UnivariateFinite{S,V}, c::V) where {S,V<:Real} = _pdf(d, c)
```
OkonSamuel, line 239: Sorry to be such a pain. Please could add comments here. To help us remember why this method was added

src/univariate_finite/methods.jl
```diff
@@ -227,20 +235,12 @@ One can also do weighted fits:
...
 
 Distributions.logpdf(d::UnivariateFinite, cv::CategoricalValue) = log(pdf(d,cv))
-
-Distributions.logpdf(d::UnivariateFinite{S,V,R,P}, c::V) where {S,V,R,P} = log(pdf(d,c))
+Distributions.logpdf(d::UnivariateFinite{S,V}, c::V) where {S,V} = log(pdf(d, c))
+Distributions.logpdf(d::UnivariateFinite{S,V}, c::V) where {S,V<:Real} = log(pdf(d, c))
```
OkonSamuel, line 243: And here also.

OkonSamuel:
@devmotion Thanks.

src/univariate_finite/methods.jl
OkonSamuel: Could you please add this as a test at https://github.com/alan-turing-institute/MLJBase.jl/blob/master/test/univariate_finite/methods.jl#L176 with a comment. Thanks

devmotion:
I addressed the comments and added a test. However, since I fixed the dispatch problem differently by removing the existing `_pdf` definition and moving it into the function body of `pdf` (it was only used there and it is not exported), I don't think the test is particularly meaningful. So maybe it should be removed again?

OkonSamuel:
> So maybe it should be removed again?

The test is fine by me

OkonSamuel (approved):

OkonSamuel:
@ablaom you might want to take a look before i merge. 
LGTM

ablaom:
@devmotion Thanks so much for this contribution! 

FYI In case you make more PRs, these should be onto `dev` instead of `master`. https://github.com/alan-turing-institute/MLJ.jl/blob/master/CONTRIBUTING.md .

Only question I have is: Are we sure the changes don't break the older versions of Distributions (^0.22, ^0.23). If not, let's just bump instead of extend.

ablaom:
Okay, I tested "backwards compatibility" locally and am satisfied.

PR #438, opened 2020-11-02
Avoids `cosh` overflow and closes #437

### LINKED ISSUE #437: log_cosh distance overflow
The current implementation of the log_cosh measure overflows quite soon:

```julia
julia> using MLJ, MLJBase

julia> log_cosh([0], [712])
1-element Array{Float64,1}:
 Inf
```

Somewhere I found this implementation, which circumvents the computation of the `cosh` responsible for the overflow:

```
softplus(x::Real) = ifelse(x > 0, x + log1p(exp(-x)), log1p(exp(x)))
logcosh(x::T) where T = x + softplus(-2x) - log(convert(T, 2))
```

See also https://github.com/JuliaML/LossFunctions.jl/issues/136#issue-678609059

ven-k:
Uses the code suggestions by @pfarndt in #437.

### DIFF
--- a/src/measures/continuous.jl
+++ b/src/measures/continuous.jl
@@ -346,10 +346,10 @@ metadata_measure(LogCosh;
     supports_weights         = false,
     docstring                = "log cosh loss; aliases: `log_cosh`.")
 
-softplus(x::Vec{<:Real}) = ifelse(x > [0], x .+ log1p.(exp.(-x)), log1p.(exp.(x)))
+_softplus(x::T) where T<:Real = x > 0 ? x + log1p(exp(-x)) : log1p(exp(x))
+_log_cosh(x::T) where T<:Real = x + _softplus(-2x) - log(convert(T, 2))
 
 function (log_cosh::LogCosh)(ŷ::Vec{<:T}, y::Vec{<:T}) where T <:Real
     check_dimensions(ŷ, y)
-    diff = ŷ - y
-    return diff + softplus(-2diff) .- log(convert(T, 2))
+    return _log_cosh.(ŷ - y)
 end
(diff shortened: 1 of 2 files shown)


src/measures/continuous.jl
```diff
@@ -346,7 +346,10 @@ metadata_measure(LogCosh;
     supports_weights         = false,
     docstring                = "log cosh loss; aliases: `log_cosh`.")
 
-function (log_cosh::LogCosh)(ŷ::Vec{<:Real}, y::Vec{<:Real})
+softplus(x::Vec{<:Real}) = ifelse(x > [0], x .+ log1p.(exp.(-x)), log1p.(exp.(x)))
```
OkonSamuel, line 349: avoid the `ifelse` statement reason being that it executes all args before checking if the condition is true or false.
Sorry if am being a pain but for performance reasons i would take advantage of julia's efficient broadcast machinery and write internal scalar functions `_softplus(x::Real) = x > 0 ? x + log1p(exp(-x)) : log1p(exp(x)) ` and `_logcosh(x::T) where T <: Real = x + _softplus(-2x) - log(convert(T, 2))` 
Then define 
```julia

function (log_cosh::LogCosh)(ŷ::Vec{<:Real}, y::Vec{<:Real})
    check_dimensions(ŷ, y)
    return _logcosh.(ŷ-y)
end
```

ven-k: I expected `ifelse` to be highly performant (despite being a function) as it is supposed to work as a broadcasted ternary operator.

OkonSamuel (approved):
@ablaom Looks good to me.

src/measures/continuous.jl
ablaom: Still, I think this comment still applies, no?

> avoid the ifelse statement reason being that it executes all args before checking if the condition is true or false.

ablaom: Okay, I see this is addressed.

ablaom:
@ven-k Thanks for this!!

PR #441, opened 2020-11-04
Fix the TagBot configuration

DilumAluthge:
This needs to be merged into master ASAP to make sure that we continue getting tags and releases.

@ablaom @tlienart

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -1,11 +1,15 @@
 name: TagBot
 on:
-  schedule:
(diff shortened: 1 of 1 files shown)


ablaom:
@DilumAluthge Thanks for that! I guess this is needed across the MLJ Universe?

DilumAluthge:
> @DilumAluthge Thanks for that! I guess this is needed across the MLJ Universe?

Yep. You can use MassInstallAction.jl to quickly make this change across multiple Julia packages.

PR #446, opened 2020-11-06
minor tweak

OkonSamuel:


src/hyperparam/one_dimensional_ranges.jl
```diff
@@ -174,7 +174,7 @@ nominal_range(T, field, ::Nothing) = throw(ArgumentError(
 "the example, "*
 "`range(Int, :dummy, lower=1, upper=10)`. "  ))
 
-function nominal_range(::Type{T}, field, values::AbstractVector{T}) where T
+function nominal_range(::Type{T}, field, values::AbstractVector{<:T}) where T
     return NominalRange{T,length(values)}(field, Tuple(values))
 end
```
ablaom, line 180: Curious. Why should this make any difference?

OkonSamuel: @ablaom . This allows me to construct values of any abstract type. 
This following code would fail on #master but pass on this pr
```julia
using Distances
r3 = range(Distances.Metric, :metric, [Euclidean(0.0), Minkowski(0.0)])
```

ablaom: Oh I see. You have `T` appearing twice in the signature (first argument and last). Thanks!

PR #459, opened 2020-11-26
Cleanup the measures (aka loss functions and scores)

### LINKED ISSUE #450: name, aliases cleanup for measures
**tl;dr:** see proposals A, B, C below. 

I realize that measures is due for a thorough review and expansion (see, in particular https://github.com/alan-turing-institute/MLJBase.jl/issues/299) but this is a moving target (see eg, the very substantial https://github.com/alan-turing-institute/MLJBase.jl/pull/430) and I would like to fix a couple of minor but annoying things before things run away from us:

The  main issue is a sad lack of consistency about the `name` attribute of a measure and a confusion about what it precisely represents. This makes programatic selection of measures buggy.

```julia
(name = area_under_curve, ...)
 (name = accuracy, ...)
 (name = balanced_accuracy, ...)
 (name = cross_entropy, ...)
 (name = FScore, ...)
 (name = false_discovery_rate, ...)
 (name = false_negative, ...)
 (name = false_negative_rate, ...)
 (name = false_positive, ...)
 (name = false_positive_rate, ...)
 (name = l1, ...)
 (name = l2, ...)
 (name = log_cosh, ...)
 (name = mae, ...)
 (name = mape, ...)
 (name = matthews_correlation, ...)
 (name = misclassification_rate, ...)
 (name = negative_predictive_value, ...)
 ⋮
 (name = L1HingeLoss(), ...)
 (name = L2HingeLoss(), ...)
 (name = L2MarginLoss(), ...)
 (name = LogitMarginLoss(), ...)
 (name = ModifiedHuberLoss(), ...)
 (name = PerceptronLoss(), ...)
 (name = SigmoidLoss(), ...)
 (name = SmoothedL1HingeLoss(), ...)
 (name = ZeroOneLoss(), ...)
 (name = HuberLoss(), ...)
 (name = L1EpsilonInsLoss(), ...)
 (name = L2EpsilonInsLoss(), ...)
 (name = LPDistLoss(), ...)
 (name = LogitDistLoss(), ...)
 (name = PeriodicLoss(), ...)
 (name = QuantileLoss(), ...)
 (name = confusion_matrix, ...)
```

Note we already have the `docstring` trait, which has (mostly) included a list of aliases for common instances of a measure. Recall that a measure is a *type* and that some measures have fields; for example `CrossEntropy` has `eps`, the cutoff threshold to prevent `NaN`. 

In MLJ, models also have a name attribute and this is  the string version of the model type; it is not literally the model type because model code is loaded on demand only. So far, we load all measures into scope. However, this might change in the future; think of measures for probabilistic predictions created by MCMC. My first proposal is:

- [x] **A**. Rather than assigning the `name` attribute for each measure, this should just be a "derived" trait that just returns the string version of the *type* name (as for models), as in `"AUC"`, `"FScore"`, `"CrossEntropy"` (better would be `"LogLoss"`;  see (https://github.com/alan-turing-institute/MLJBase.jl/issues/299#issuecomment-721520032), `"ZeroOneLoss"`, etc.

Next, rather than throwing aliases for default instances into the doc-string, we should:

- [x] **B.**  Add an new trait called `aliases` or `instance_aliases`; this returns a vector of strings, whose corresponding `Symbol` expressions can be evaluated when relevant code has been loaded. For example `TruePositiveRate` gets the value `["true_positive_rate", "truepositive_rate", "tpr", "sensitivity", "recall", "hit_rate"]`. This trait would be listed second in the "registry" (what you print out with `measures()`) to maximise discoverability.

The final clean-up item would be:

- [x] **C**. Ensure that every measure has a default keyword constructor with defaults for all arguments. So `FScore()` should just work; it currently doesn't.

These changes are technically breaking and would require a new release, but I can't imagine they would not be that disruptive.

Thoughts anyone?

**edit**

And I forgot:

- [x] **D**. Make sure all current measure type names are exported, as well as their aliases.

ablaom:
This PR resolves #450:

- [x] (**enhancement**) Thorough review of measure doc-strings, including addition of automatic doc-string generation to simplify new contributions (`@create_docs` macro)

- [x] Add a new trait for measures called `human_name` (mainly for use in documentation); for example its value for `L1EpsilonInsLoss` is `l1  ϵ-insensitive loss`. There is a fallback that guesses a human name from the type name. 

- [x] (**breaking**) Redefine `name` trait on all measures to return string version of type name, eg, for `L1EpsilonInsLoss` type (or its instances) the value is "L1EpsilonLoss". This is for consistency with `name` on models. Type names have changed to make them maximally descriptive (eg, `RMS` is now `RootMeanSquaredError`) but the old names are kept as aliases.  

- [x] (**enhancement**) Add a new trait for measures called `instances` (distinct from `Base.instances` and not exported) which lists string versions of built in instances, mostly aliases for the default instance. For example, the value on `Recall` is `["true_positive_rate", "truepositive_rate", "tpr", "sensitivity", "recall", "hit_rate"]`. These instances appear second on calls to `measures()` so user to can easily lookup a short method to call in common use cases (see below)

- [x] (**enhancement**) Ensure every measure has a default keyword constructor, so, for example `FScore()` now works. 

- [x] (**breaking**) `FScore` are `BrierScore` have been simplified to avoid unnecessary type parameters; this may break some constructors, like `FScore{1.0}()`. You should never need to use `{}` to construct a measure instance. 

- [x] (**enhancement**) All types from LossFunctions.jl are exported and they now have default key-word constructors, just like the built-in measures, and their instances can be called the same way too. (In LossFunctions.jl the instances are not callable.)

- [x]  Add `BierLoss`. It is just negative of `BrierScore`. 

- [x] (**breaking**) Dissallow keyword arguments in calls to `confmat`, which is now an instance of a measure type (`confmat = ConfusionMatrix()`), rather than a function hacked to look like a measure. So, instead of `confmat(yhat, y, perm=[2, 3, 1])` use `ConfusionMatrix(perm=[2, 3, 1])(yhat, y)`.  What used to be called  `ConfusionMatrix` is now `ConfusionMatrixObject`. The ordinary user will never need to construct one directly.  

- [x] (**mildlly breaking**) The field name `eps` in `LogLoss=CrossEntropy` (for clamping) is renamed to `tol` for consistency with other measures and elsewhere in julia. However `CrossEntropy(eps=...)` still works. 

```julia
julia> measures()
59-element Array{NamedTuple{(:name, :instances, :human_name, :target_scitype, :supports_weights, :prediction_type, :orientation, :reports_each_observation, :aggregation, :is_feature_dependent, :docstring, :distribution_type, :supports_class_weights),T} where T<:Tuple,1}:
 (name = Accuracy, instances = [accuracy], ...)
 (name = AreaUnderCurve, instances = [area_under_curve, auc], ...)
 (name = BalancedAccuracy, instances = [balanced_accuracy, bacc, bac], ...)
 (name = BrierLoss, instances = [brier_loss], ...)
 (name = BrierScore, instances = [brier_score], ...)
 (name = ConfusionMatrix, instances = [confusion_matrix, confmat], ...)
 (name = FScore, instances = [f1score], ...)
 (name = FalseDiscoveryRate, instances = [false_discovery_rate, falsediscovery_rate, fdr], ...)

...

 (name = L2MarginLoss, instances = [l2_margin_loss], ...)
 (name = LogitMarginLoss, instances = [logit_margin_loss], ...)
 (name = ModifiedHuberLoss, instances = [modified_huber_loss], ...)
 (name = PerceptronLoss, instances = [perceptron_loss], ...)
 (name = SigmoidLoss, instances = [sigmoid_loss], ...)
 (name = SmoothedL1HingeLoss, instances = [smoothed_l1_hinge_loss], ...)
 (name = ZeroOneLoss, instances = [zero_one_loss], ...)
 (name = HuberLoss, instances = [huber_loss], ...)
 (name = L1EpsilonInsLoss, instances = [l1_epsilon_ins_loss], ...)
 (name = L2EpsilonInsLoss, instances = [l2_epsilon_ins_loss], ...)
 (name = LPDistLoss, instances = [lp_dist_loss], ...)
 (name = LogitDistLoss, instances = [logit_dist_loss], ...)
 (name = PeriodicLoss, instances = [periodic_loss], ...)
 (name = QuantileLoss, instances = [quantile_loss], ...)

julia> log_loss
LogLoss(
    tol = 2.220446049250313e-16) @422
```

I will be looking to merge this in about 13 hours time so that the PR for CategorialArrays [compat] bump (also breaking) can follow soon after and a minor release can go out soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ repository provides core functionality for MLJ, including:
   for **performance measures** (losses and scores), enabling the
   integration of the
   [LossFunctions.jl](https://github.com/JuliaML/LossFunctions.jl)
-  library, user-defined measures, as well as about two dozen natively
+  library, user-defined measures, as well as about forty natively
   defined measures.
 
 - integration with [OpenML](https://www.openml.org)
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -210,32 +210,34 @@ export cross_entropy, BrierScore, brier_score,
     BrierLoss, brier_loss,
     LogLoss, log_loss,
     misclassification_rate, mcr, accuracy,
-    balanced_accuracy, bacc, bac,
+    balanced_accuracy, bacc, bac, BalancedAccuracy,
     matthews_correlation, mcc, MCC, AUC, AreaUnderCurve,
-    MisclassificationRate, Accuracy, MCR, BACC, BAC
+    MisclassificationRate, Accuracy, MCR, BACC, BAC,
+    MatthewsCorrelation
 
 # measures/finite.jl -- Multiclass{2} (order independent):
 export auc, area_under_curve, roc_curve, roc
 
 # measures/finite.jl -- OrderedFactor{2} (order dependent):
 export TruePositive, TrueNegative, FalsePositive, FalseNegative,
-       TruePositiveRate, TrueNegativeRate, FalsePositiveRate,
-       FalseNegativeRate, FalseDiscoveryRate, Precision, NPV, FScore,
-       # standard synonyms
-       TPR, TNR, FPR, FNR, FDR, PPV,
-       Recall, Specificity, BACC,
-       # instances and their synonyms
-       truepositive, truenegative, falsepositive, falsenegative,
-       true_positive, true_negative, false_positive, false_negative,
-       truepositive_rate, truenegative_rate, falsepositive_rate,
-       true_positive_rate, true_negative_rate, false_positive_rate,
-       falsenegative_rate, negativepredictive_value,
-       false_negative_rate, negative_predictive_value,
-       positivepredictive_value, positive_predictive_value,
-       tpr, tnr, fpr, fnr,
-       falsediscovery_rate, false_discovery_rate, fdr, npv, ppv,
-       recall, sensitivity, hit_rate, miss_rate,
-       specificity, selectivity, f1score, fallout
+    TruePositiveRate, TrueNegativeRate, FalsePositiveRate,
+    FalseNegativeRate, FalseDiscoveryRate, Precision, NPV, FScore,
+    NegativePredictiveValue,
+    # standard synonyms
+    TPR, TNR, FPR, FNR, FDR, PPV,
+    Recall, Specificity, BACC,
+    # instances and their synonyms
+    truepositive, truenegative, falsepositive, falsenegative,
+    true_positive, true_negative, false_positive, false_negative,
+    truepositive_rate, truenegative_rate, falsepositive_rate,
+    true_positive_rate, true_negative_rate, false_positive_rate,
+    falsenegative_rate, negativepredictive_value,
+    false_negative_rate, negative_predictive_value,
+    positivepredictive_value, positive_predictive_value,
+    tpr, tnr, fpr, fnr,
+    falsediscovery_rate, false_discovery_rate, fdr, npv, ppv,
+    recall, sensitivity, hit_rate, miss_rate,
+    specificity, selectivity, f1score, fallout
 
 # measures/finite.jl -- Finite{N} - multiclass generalizations of
 # above OrderedFactor{2} measures (but order independent):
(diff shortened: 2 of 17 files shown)


ablaom:
Looks like the travis fail is unrelated (connection to OpenML busy)

PR #462, opened 2020-11-27
Address a few things uncovered by the VS Code linter

CameronBieganek:
This PR addresses a few things that the VS Code linter uncovered.

The linter also recommends using `x === nothing` rather than `x == nothing`. (Actually, I prefer to use `isnothing`.)

Also, there appear to be a few whitespace changes. I have my editor set up to strip trailing whitespace characters when I save a file.

### DIFF
--- a/src/measures/measures.jl
+++ b/src/measures/measures.jl
@@ -33,7 +33,7 @@ orientation(::Type) = :loss  # other options are :score, :other
 reports_each_observation(::Type) = false
 aggregation(::Type) = Mean()  # other option is Sum() or callable object
 is_feature_dependent(::Type) = false
-human_name(::Type) = snakecase(name(M), delim=' ')
+human_name(M::Type) = snakecase(name(M), delim=' ')
 instances(::Type) = String[]
 
 # specific to `Finite` measures:
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -445,7 +445,7 @@ function _process_accel_settings(accel::CPUThreads)
       throw(ArgumentError("`n`used in `acceleration = CPUThreads(n)`must" *
                         "be an instance of type `T<:Signed`"))
       accel.settings > 0 ||
-            throw(error("Can't create $(acceleration.settings) tasks)"))
+            throw(error("Can't create $(accel.settings) tasks)"))
       _accel = accel
     end
     return _accel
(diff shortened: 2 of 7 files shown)


DilumAluthge:
> The linter also recommends using `x === nothing` rather than `x == nothing`. (Actually, I prefer to use `isnothing`.)

From the compiler's point of view, `x === nothing` is much better.
- `==` can be overloaded
- `isnothing` can be overloaded
- `===` cannot be overloaded

CameronBieganek:
Interesting! But shouldn't the compiler be able to inline `isnothing` most of the time?

At any rate, I'll probably still stick with `isnothing` in my own codebase, since I find it easier to read.

src/resampling.jl
```diff
@@ -445,7 +445,7 @@ function _process_accel_settings(accel::CPUThreads)
       throw(ArgumentError("`n`used in `acceleration = CPUThreads(n)`must" *
                         "be an instance of type `T<:Signed`"))
       accel.settings > 0 ||
-            throw(error("Can't create $(acceleration.settings) tasks)"))
+            throw(error("Can't create $(accel.settings) tasks)"))
```
ablaom, line 448: Good catch.

src/measures/measures.jl
```diff
@@ -33,7 +33,7 @@ orientation(::Type) = :loss  # other options are :score, :other
 reports_each_observation(::Type) = false
 aggregation(::Type) = Mean()  # other option is Sum() or callable object
 is_feature_dependent(::Type) = false
-human_name(::Type) = snakecase(name(M), delim=' ')
+human_name(M::Type) = snakecase(name(M), delim=' ')
```
ablaom, line 36: Actually did catch this one 😄

src/univariate_finite/arrays.jl
```diff
@@ -149,14 +149,14 @@ function Base.Broadcast.broadcasted(
         "Arrays could not be broadcast to a common size; "*
         "got a dimension with lengths $(length(u)) and $(length(v))"))
     for cv in v
-        cv in classes(u) || _err_missing_class(c)
+        cv in classes(u) || _err_missing_class(cv)
```
ablaom, line 152: Another good catch!

src/utilities.jl
```diff
@@ -112,7 +112,7 @@ function reduce_nested_field(ex)
     tail = ex.args[2]
     tail isa QuoteNode || throw(ArgumentError)
     field = tail.value
-    field isa Symbol || throw(ArgmentError)
+    field isa Symbol || throw(ArgumentError)
```
ablaom, line 115: Oops!

ablaom (approved):
@CameronBieganek 

Many thanks for this! I do have the linter working again in my ODE. Hopefully I pay attention when it suggests corrections. 

Final note in passing. I seem to remember `isnothing` does not work in in julia 1.0. I expect the compiler would be happy with `x isa Nothing` too?

Issue #457, opened 2020-11-23
In resampling: Allow specification of class weights, for passing to classification metrics that support them 


ablaom:
#456 implements measures, such as `MulticlassFScore`, which allow the user to pass class weights as in:

```julia
y = coerce(categorical([1, 2, 0, 2, 1, 0, 0, 1, 2, 2, 2, 1, 2,
                            2, 1, 0, 1, 1, 1, 2, 1, 2, 2, 1, 2, 1,
                            2, 2, 2]), Multiclass)
ŷ = coerce(categorical([2, 0, 2, 2, 2, 0, 1, 2, 1, 2, 0, 1, 2,
                            1, 1, 1, 2, 0, 1, 2, 1, 2, 2, 2, 1, 2,
                            1, 2, 2]), Multiclass)
class_w = Dict(0=>1, 1=>2, 2=>3) 

julia> MulticlassFScore()(ŷ, y, class_w)
0.8890257252326217
```

In analogy with the current interface point in for specifying *per-sample* weights, we should add an interface 
point for specifying class weights. So, for example, something like this should make sense:

```julia
X, y = @load_iris
model = @load DecisionTreeClassifier pkg=DecisionTreeClassifier
mach = machine(model, X, y)
evaluate!(mach, 
    resampling=CV(),
    class_weights=Dict("setosa"=>1, "versicolor"=>1.5, "virginica"=>1.9),
    measure=MulticlassFScore())
```

Should be allow both per-sample `weights=...` *and* per-class `weights` to be specified? Probably, as we may have a number of measures, some that support one kind and some that support another. 

cc: @ven-k

(Note currently no *model* supports class weights for training, although the model API does not preclude this, and some scikitLearn-learn models [already provide this possibility](https://github.com/alan-turing-institute/MLJ.jl/issues/327). The user interface point is in the machine constructor, as in `mach(model, X, y, class_w)`, just as for per-sample weights, where one does `mach(model, X, y, w)`. )

PR #464, opened 2020-11-28
`class_weights` for `evaluate!`

ven-k:
-- Solves initial half of #457 
-- This adds `class_weights::Union{Nothing, AbstractDict{<:Any, <:Real}}` for 
- `evaluate!` functions
- `mutable struct Resampler` and `resampler` functions

-- Warns that it is being ignored when it isn't applicable.
-- Does not _process_class_weights in **resampling.jl** as that is taken care by the implementation of relevant measures.
-- This considers that both `weights` and `class_weights` aren't supported by any measure at the same time.


src/resampling.jl
```diff
@@ -937,6 +967,8 @@ end
 MLJBase.is_wrapper(::Type{<:Resampler}) = true
 MLJBase.supports_weights(::Type{<:Resampler{<:Any,M}}) where M =
     supports_weights(M)
+supports_class_weights(::Type{<:Resampler{<:Any,M}}) where M =
```
ven-k, line 970: Here the problem seems that MLJModelInterface doesn't have `supports_class_weights` in `model_traits`. Should I add it to match it with the rest?

test/resampling.jl
```diff
@@ -4,6 +4,7 @@ using Distributed
 import ComputationalResources: CPU1, CPUProcesses, CPUThreads
 using .TestUtilities
 using ProgressMeter
+using MLJScikitLearnInterface: RidgeCVClassifier
```
ven-k, line 7: All native models in `Model` returns 0s as the output for this test; while `MLJScikitLearnInterface` seems to introduce some version incompatibility.

ablaom:
@ven-k I assume you will ping me when you are ready for a review. Thanks.

test/resampling.jl
```diff
@@ -323,11 +324,46 @@ end
...
+    @test round(e, digits=3) ≈ 
+        round(mean([0.333, 0.333, 0.624, 0.639, 0.913, 0.864]), digits=3)
+
+    # if class weights in `evaluate!` isn't specified:
+    e = evaluate!(mach, resampling=cv, measure=MulticlassFScore(return_type=Vector),
+                  verbosity=verb, acceleration=accel).measurement[1]
+    @test round(e, digits=3) ≈ 
+        round(mean([0.333, 0.333, 0.312, 0.319, 0.304, 0.288]), digits=3)
+end
+
```
ven-k, line 345: @ablaom this has been a tricky error. The `machine(::SVC ...)` fails to construct in `test` environment, however `evaluate!` works fine when invoked in `] develop MLJBase` env in repl.

However for `DeterministicConstantClassifier` it works always. (See `class weights in evaluation (native)`). Output is a zero vector, hence not a good test.

test/resampling.jl
```diff
@@ -475,6 +512,27 @@ end
                      acceleration=accel, verbosity=verb).measurement[1]
 
     @test e1 ≈ e2
+
+    model = @load SVC
```
ven-k, line 516: Here, I had to redefine `resampler`. So I couldn't test in `] develop` mode in repl. (Is there a way to do it without renaming for test purposes)

Even here the `resampler` doesn't construct for SVC model. (It does for DeterministicConstantClassifier, which isn't a good test)

Is it because it's an imported model?

OkonSamuel:
> his considers that both `weights` and `class_weights` aren't supported by any measure at the same time.

Aren't we being too streamlined here. Is it possible in the future to have a measure that supports both of these traits.

src/measures/measures.jl
```diff
@@ -110,7 +110,8 @@ aggregate(x::MeasureValue, measure) = x
 
 function value(measure, yhat, X, y, w)
     vfdep     = Val(is_feature_dependent(measure))
-    vsweights = Val(supports_weights(measure))
+    vsweights = Val(supports_weights(measure) ||
+                    supports_class_weights(measure))
     return value(measure, yhat, X, y, w, vfdep, vsweights)
```
OkonSamuel, line 115: @ven-k and @ablaom. can't this be 
`value(measure, yhat, X, y, w, vfdep, sample_weights, class_weights)` just in case a measure supports both.

src/resampling.jl
ablaom: Yes, to be consistent should add `supports_class_weights` to MLJModelInterface. However, doing so will break the whole stack, and in any case I don't like this coupling between MLJModelInterface and the measures API, which is eventually to become a stand-alone package (without MLJModelInterface as a dependency). Sorry to say this, but I think this issue should probably be sorted before proceeding, and I have just posted this: https://github.com/alan-turing-institute/MLJBase.jl/issues/495

test/resampling.jl
ablaom: In MLJBase testing `@load` does not work for arbitrary models. It only works for limited models in test/_models/ . You should attempt to limit your testing to those models and not introduce further test dependencies, unless this is really necessary. The available models are these:

``` 
 SimpleDeterministicCompositeModel
 DecisionTreeRegressor
 DeterministicConstantClassifier
 DeterministicConstantRegressor
 FooBarRegressor
 KNNRegressor
 RidgeRegressor
 ConstantClassifier
 ConstantRegressor
 DecisionTreeClassifier
 KNNClassifier
 FeatureSelector
 OneHotEncoder
 PCA
 Standardizer
 UnivariateBoxCoxTransformer
 UnivariateDiscretizer
 UnivariateStandardizer
```
These become available in a test module when you do `using .Models`.

test/resampling.jl
ablaom: Perhaps my previous comment addresses this also?

src/measures/measures.jl
ablaom: We could, but this would break the current API which has the signature of `value` fixed to 7 arguments. Let me think about this.

ven-k: @ablaom, PR is ready for review for things I planned with this PR so far; based on what you decide about the above change and `supports_class_weights` I will move forward.

ven-k:
> Aren't we being too streamlined here. Is it possible in the future to have a measure that supports both of these traits.

@OkonSamuel, I didn't find any measure with both weights on my primary search. However I don't see why it can _never_ be used. So based on what gets decided about `value(...` from above I will make the changes.

ablaom:
Am revisiting this today, in light of the resolution of some issues preventing progress earlier:

- issue, now resolved,  with `selectrows(::Dict, rows)`: https://github.com/alan-turing-institute/MLJModelInterface.jl/pull/84
- `supports_class_weights` not available to models implementing MLJModelInterface.jl, resolved in MLJModelInterface.jl 0.4 (wasn't actually a show-stopper for this PR, but good to have)
- bunch of changes to `evaluate!` to support new data front-end (#501), likely to cause merge conflicts 

To do:

- [ ] check/ resolve conflicts in this PR (@ablaom)

ablaom:
@ven-k I have resolved the merge conflicts here: #515 .

Suggest you close this PR in favour of #515 or a new PR continuing #515.

Let me know when you are ready for a final review and I will provide one promptly.

ven-k:
Thanks @ablaom for all the new changes in the API. I'll update the PR #515 soon

PR #483, opened 2020-12-01
Fix the docs deployment

DilumAluthge:
@ablaom


DilumAluthge:
@ablaom This should fix it.

PR #485, opened 2020-12-01
Only build docs on PRs from dev to master, pushes to master, and pushes to tags

DilumAluthge:


ablaom (approved):
Looks good to me, thanks!

Build fail unrelated.

PR #486, opened 2020-12-02
Test doc ci again - should be no doc generation in this PR

ablaom:


ablaom:
Good. Working as expected.

Issue #309, opened 2020-05-14
taking performance issue more seriously


OkonSamuel:
I know MLJ will definitely have some overhead since it's wraps other code. But i believe this overhead can be reduced below the current level with care design considerations.  Avoidable overhead which are neglected may come to hunt us when the code is scaled. There are a few things i discovered which are important.
1. As pointed out in issue #151 `selectrows`. This causes overhead when used in `evaluate` method which called these method a lot of times depending on the `repeats` parameter.
2. In the `MLJBase.fit` method, the `matrix` method (which to my understanding copies data)  is called on a given table `X` of course this isn't bad if i call this method only once using the same data changing a couple of model parameters. This becomes important in the `evaluate` method which calls `fit` a couple of times depending on `repeats` .(If X is larger this isn't nice). (I don't think `update` method does enough justice). Also other methods call these methods repeatedly copying X in each case
3. The return type to `MLJBase.predict` for probabilistic Classifiers. I can't find the issue here.

This are just some of the point. I believe there are other things which affects scalability. It is better we start taking this issues. Imagine a case where One wants to embark on a kaggle competition on a large dataset only to find out that the overhead is just unbearable in this case). You may correct me if perhaps i'm missing something. 
@ablaom , @tlienart

tlienart:
re (2) I raised that a while back but I seem to remember that MatrixTable can be formed without copies but I'm not sure... what's for sure is that in some case  (when you have to materialise the transpose for e.g. glm stuff) then there's definitely a copy.

OkonSamuel:
@tlienart. What do you think?
Since we are still going to call `matrix(X)` in the inner methods. Why don't we don't in some cases call `matrix(X)` earlier so we can use `Xmatrix[inds, :]` many times instead of having to call both `selectrows` and `matrix` many times. Also `Xmatrix[inds, :]` should be faster than `selectrows`.
It just a suggestion though??

OkonSamuel:
In addition i noticed that internally all tables are converted to dense matrices. Are there not cases where we would like to have a sparse matrix.(Maybe if something like a sparse table exists??).

tlienart:
> In addition i noticed that internally all tables are converted to dense matrices. 

what code are you referring to here?

I think Anthony will have to chip in to give more clarity here but in terms of the sparse case I believe I tried that when we were working on LightGBM, you can have a Table wrapping around a sparse matrix and recover said sparse matrix (but I'm not certain).

[...]

PR #492, opened 2020-12-03
Performance improvements obtained by caching model-specific data representations and resamplings thereof

ablaom:
Depends on: https://github.com/alan-turing-institute/MLJModelInterface.jl/pull/76 which introduces `reformat` and `selectrows(::Model, I, args...)` to the MLJ model API. 

### Goals

Addresses some issues raised in #309, and in particular:

- Avoid the unnecessarily reconversion of data into a form specific to the model algorithm (eg, table -> matrix). Currently this happens every time model hyperparameter changes, or when training rows change.

- Allow resampling of data to happen at level of model-specific representation. Currently resampling is applied to the user-representation (eg, table). This would also allow implementation to use matrix views, for example,  avoiding copying, if this is appropriate. 

- Avoid unecessary resamplings of the data. Currently data is resampled on every `fit!` after a model hyperparameter change, even if the training rows are unchanged.

This PR is a proof of concept achieving these goals by:

- Adding new fields `data` and `resampled_data` to machines to cache model-specific representations of the data, and resampled versions respectively

- Change the logic in `fit_only!` to cache these representations, making use of the two methods `reformat` and  `selectrows(::Model)` in such a way that the fallbacks given in MLJModelInterface result in essentially the old behaviour. 

### Main features proposed solution

- **Buying into the improvements is optional** for existing (or new) implementations of the extended model API and so can be phased in gradually.

- Changes suggested here are only breaking to users who access current "low-level" methods `fit` and `update`, and only for those models that choose to buy into the performance improvements by implementing `reformat`, `selectrows(::Model)` and altering their `fit/predict` (see "Buying into the extended API" below). Users who interact using `fit!` should not encounter breaking behaviour. 

- Tested locally with the MLJModelInterface PR

- The current PR is limited to achieving the goals above on the training side, not on the predict/transform side

### Buying into the extended API

Here's an example of how to buy into the the new API, which is tested in detail [here](https://github.com/alan-turing-institute/MLJBase.jl/blob/b260019cd5f38ed7899b28109f003e3973cbac94/test/machines.jl#L176)

#### Existing implementation

```julia
function MLJModelInterface.fit(model::SomeModel,
                               verbosity,
                               X,
                               y)

    A = MLJModelInterface.matrix(X)
    fitresult = ...

    return fitresult, cache, report
end

function MLJModelInterface.predict(::SomeModel, fitresult, Xnew)
    Anew = MLJModelInterface.matrix(Xnew)
    return Anew*fitresult
end
```

#### Additions and changes

The signature of `fit` is changed to accept matrix `A` in place of table `X`, and the conversion to matrix is removed, with that role being played by the overloaded `reformat` method:
```julia

function MLJModelInterface.fit(model::SomeModel,
                               verbosity,
                               A,
                               y)
    fitresult = ...

    return fitresult, cache, report
end

MLJModelInterface.reformat(model::SomeModel, X, y) = (MLJModelInterface.matrix(X), y)
```

If `update` is also overloaded, then a similar change is made there. 

Finally, we must specify how to resample the model-specific representation of the data:

```julia
MLJModelInterface.selectrows(model::SomeModel, I, A, y) = (view(A, I, :), view(y, I))
```

That's it.


ablaom:
Thoughts anyone?

@OkonSamuel @tlienart @yalwan-iqvia @jeremiedb

yalwan-iqvia:
So cursory glance this seems good to me ( I didn't touch the code changes yet )

One of my main remarks would be that this seems geared towards implementors buying into the performance enhancements, however, I think a level of control needs to be afforded to the user on a case by case basis as to whether the implementation *should* buy in -- why? 

Well for some modelling/system setups, modelling might be getting near to memory limits in system. In this case caching resamplings can make the difference between running/not running.

In this case, I'd imagine user wishes to opt out in order to have it running with degraded performance, vs not running at all.

Does current implementation provide any mechanism for this?

yalwan-iqvia:
P.s. Thanks for this

In LightGBM the system current has to construct LGBM specific data-sets for each resampling of data prior to fitting. This can actually be computationally non-trivial. By providing a mechanism by which we can hold ontoi our own form of the data and then producing our own resamples, this cost can almost entirely be negated. In fact, in my tests, the subsampling of the specific data structure is much faster than even just vanilla matrix slicing for large data sets (multi GBs +)

ablaom:
@yalwan-iqvia Thanks for the feedback!

> In this case, I'd imagine user wishes to opt out in order to have it running with degraded performance, vs not running at all.

Maybe worth mentioning here that there are two kinds of caching - caching the model-specific representation of the whole training data set (before resampling) - and caching a particular subsample. Whether each adds to memory requirements depends both on the model implementation and what form the user-data takes. For example, if the "table" presented by the user is just a wrapped matrix (a `MatrixTable`) and the model algorithm is just unwrapping it, then there is no extra memory cost to "caching" the data. We are just pointing to stuff inside the user-specified data. Similarly, if subsampling of a matrix is just constructing a `view`, then again no extra memory cost. 

But, yes an "optimize_memory" option makes sense and can be added. However a particular model may want to also provide options - such as whether to construct a `view` of data or make a copy when restricting to specified rows of a matrix.

Can one construct a "view" of the model-specific representation in LightGBM (DMatrices?) (a view that can be consumed by `fit`). Or does indexing necessarily create copies? 

Be **great** if you could have a stab at implementing the extended API for LightGBM in a fork and do some benchmarking. The two branches you need - from MLJModelInterface - and from MLJBase for testing - are both called "mlj2". 

FYI I am on leave for the next week.

yalwan-iqvia:
> Can one construct a "view" of the model-specific representation in LightGBM (DMatrices?) (a view that can be consumed by fit). Or does indexing necessarily create copies?

I *think* so

Basically you can create a Dataset object and then on top of this there is a getsubset API which returns a new dataset reference, but when I tested this with whomper data and then obtained a subset dataset I didnt see the memory usage change -- I believe (without looking at the C++ source code) that it just just a new dataset object which references the old one and additionally stores the subset elements request for use during training steps. Since this is still a dataset object it is consumable by lightgbm for subsequent fit calls.

We actually put a PR on the mainline code to allow this specialisation and performance improvement just recently:

https://github.com/IQVIA-ML/LightGBM.jl/pull/80/files

Idea was rather than requiring matrices, refactor out the dataset construction element and allow user to provide own dataset(s). This would then make it straightforward for us to support the paradigm within the MLJ wrapper (as the "user").

I can't promise I can get around to trying out the extended API over the next week but I'll certainly see what I can do

jeremiedb:
Sorry being late to the show. Being able to circumvent the overhead was definitely a concern on my side, very nice to see an answer to it! I'll be testing the proposed buy-in from the `mlj2` branches with EvoTrees and hopefully there won't be any noticeable difference with the native API. I'll get back to you shortly. Let me know if there were any other aspect you'd like me to check.

jeremiedb:
So the outcomes look positive from using the `mlj2` branches. 

For the model setup and initial iterations, the timing remains similar, but with a good reduction in the memory footprint: 

Old: 
``` julia
julia> @time fit!(tree, rows=train, verbosity=1, force=false)
[ Info: Training Machine{EvoTreeRegressor{Float32,…}} @230.
  3.089232 seconds (22.55 k allocations: 2.899 GiB, 18.94% gc time)
```

MLJ2:
```julia
julia> @time fit!(tree, rows=train, verbosity=1, force=false)
[ Info: Training Machine{EvoTreeRegressor{Float32,…}} @882.
  2.823589 seconds (21.94 k allocations: 1.701 GiB, 4.42% gc time)
```

Where things are interesting, is that while the backen `update` function was already performing well, the issue was coming from the call to `fit!` it situations of simple model update. On the new `mlj2` branches, that `fit!` function now performs pretty as fast as the underlying update. 

Old: 
```julia
tree.model.nrounds += 1
julia> @time update(tree.model, 0, tree.fitresult, tree.cache, X, Y);
  0.068055 seconds (1.94 k allocations: 41.474 MiB)

julia> @time fit!(tree, rows=train, verbosity=1)
[ Info: Updating Machine{EvoTreeRegressor{Float32,…}} @230.
  0.684355 seconds (2.78 k allocations: 658.003 MiB, 4.51% gc time)
```

New: 
```julia
tree.model.nrounds += 1
julia> @time update(tree.model, 0, tree.fitresult, tree.cache, X, Y);
  0.074739 seconds (1.95 k allocations: 41.473 MiB)

julia> @time fit!(tree, rows=train, verbosity=1)
[ Info: Updating Machine{EvoTreeRegressor{Float32,…}} @882.
  0.085758 seconds (2.22 k allocations: 41.507 MiB)
```

Which is good to see! 

Some small notes about implementation. 
To restrain the dependency to `MLJModelInterface`, the MLJBase.matrix is changed for: 
```julia
MLJModelInterface.reformat(model::EvoTypes, X, y) = (MLJModelInterface.matrix(X), y)
```` 

For the `metadata_model`, I've comment out `input=MLJModelInterface.Table(MLJModelInterface.Continuous)`.  What would be the proper way to specify the union of Table and Matrix?

Finally, this is a bit tangent, but has there been any adopted design for supporting the particular iterative training nature of boosted models? I'm referring to the difficulty from my perspective from using MLJ for such models given that the number of iterations isn't treated as other hyper-parameters, but rather more as an upper bound while the model is trained till an early stop criteria is met. It is over such sequence of iterations that one would want to perform an hyper search for GBT, but it doesn't appear trivial to integrate a tracking metric of an eval dataset in MLJ for that purpose, maybe I've overlooked this? That early stop over an eval data is something commonly embedded in GBT libraries, whether it's xgboost, lightgbm or evotrees.

**Comment by ablaom** [comment]
@jeremiedb Thanks for that very helpful and encouraging feedback!

> Some small notes about implementation.
> To restrain the dependency to MLJModelInterface, the MLJBase.matrix is changed for:

Good catch, thanks. Updated my original post above.

> For the metadata_model, I've comment out input=MLJModelInterface.Table(MLJModelInterface.Continuous). What would be the proper way to specify the union of Table and Matrix?

Are you saying you needed to comment this out to get the mlj2 branches to work? If so, please post the error. 

Or you want to allow both forms of input? Yes; you can do this:

```julia
input=Union{MLJModelInterface.Table(MLJModelInterface.Continuous), AbstractMatrix{MLJModelInterface.Continuous)}
```

But this means the `X` received by `fit` in the old MLJ could be either type (and you have to process appropriately) or, in mlj2, you have two `reformat` methods. For example, something like:

```julia
_reformat(::Type{<:Table}, X) = MLJModelInterface.matrix(X)
_reformat(::Type{<:AbstractMatrix}, X) = X
MLJModelInterface.reformat(model::SomeModel, X, y) = (_reformat(scitype(X), X), y) 
```

> Finally, this is a bit tangent, but has there been any adopted design for supporting the particular iterative training nature of boosted models? I'm referring to the difficulty from my perspective from using MLJ for such models given that the number of iterations isn't treated as other hyper-parameters, but rather more as an upper bound while the model is trained till an early stop criteria is met. It is over such sequence of iterations that one would want to perform an hyper search for GBT, but it doesn't appear trivial to integrate a tracking metric of an eval dataset in MLJ for that purpose, maybe I've overlooked this? That early stop over an eval data is something commonly embedded in GBT libraries, whether it's xgboost, lightgbm or evotrees.

Yes, this is a very high priority (https://github.com/alan-turing-institute/MLJ.jl/issues/139). Appreciate your patience.

jeremiedb:
> Are you saying you needed to comment this out to get the mlj2 branches to work? If so, please post the error.

No, it was just a confusion on my side where I assumed MLJ2 would be expected to directly support support Matrix. Actually, the Union for Tables and Matrix is right what I wanted, so both Tables and Matrix goes smoothly, even without adding the double `_reformat` trick. 

No worry about the evaluation metric tracking, it's just that I recently had some reflections on whether it whether we should consider transfer our ML pipeline from R to Julia, and to what extent MLJ could be piggy-backed. 

Do you have any rough timeline as to when you expect to have those MLJ2 released?

ablaom:
@jeremiedb I've now a pretty good idea how the improvements can be carried over to the prediction side.  After Christmas, I'll add this and we should be good to go. So, hopefully early January for mlj2.

ablaom:
Update:

I have now added:

- Option for a machine to turn off the data caching. This is implemented with a new type parameter `C` for machines, now instances of `Machine{M<:Model, C}` where `C` is (`true` or `false`). To instantiate a machine with no caching user does `machine(model, data...; cache=false)`. 
 
- Data caching on the `predict` side (and for any other operation, except `inverse_transform`). This requires the implementer to: (i)  further overload `reformat(model, args...)` and `selectrows(model, I, data...)` for the case that `length(args)=1` and `length(data)=1` (unless model is `Unsupervised`, where this would have been done already on the training side); and (ii) refactor every operation (such as `predict`, but except for `inverse_transform`) to expect its data arguments to be `reformat`ed versions of their original ones. 

Details are given in a proposed [documentation PR](https://github.com/alan-turing-institute/MLJ.jl/blob/mlj2/docs/src/adding_models_for_general_use.md#implementing-a-data-front-end) at MLJ.

ablaom:
Update:

I have now added:

- Option for a machine to turn off the data caching. This is implemented with a new type parameter `C` for machines, now instances of `Machine{M<:Model, C}` where `C` is (`true` or `false`). To instantiate a machine with no caching user does `machine(model, data...; cache=false)`. 

- Interface point in `Resample` objects for specifying what `cache` parameter to use in machines constructed in resampling. This is tested in the following [MLJTuning PR](https://github.com/alan-turing-institute/MLJTuning.jl/pull/107), which passes locally.
 
- Data caching on the `predict` side (and for any other operation, except `inverse_transform`). This requires the implementer to: (i)  further overload `reformat(model, args...)` and `selectrows(model, I, data...)` for the case that `length(args)=1` and `length(data)=1` (unless model is `Unsupervised`, where this would have been done already on the training side); and (ii) refactor every operation (such as `predict`, but except for `inverse_transform`) to expect its data arguments to be `reformat`ed versions of their original ones. This only avoids an extra `reformat` in the case of a call like `predict(mach, rows=...)`, and resampling has been refactored to call `predict(mach, rows=...)` instead of `predict(mach, Xtrain)` to make use of this.

Details are given in a proposed [documentation PR](https://github.com/alan-turing-institute/MLJ.jl/blob/mlj2/docs/src/adding_models_for_general_use.md#implementing-a-data-front-end) at MLJ.

ablaom:
Re-opening as error appears to be OpenML connection related.

ablaom:
closing in favour of #501

PR #501, opened 2021-01-04
Realize performance improvements for models implementing new data front-end

ablaom:
Continuation of #492 with squashed commits.

Addresses key parts of #309.

Builds on the [MLJModelInterface PR 76](https://github.com/alan-turing-institute/MLJModelInterface.jl/pull/76) (now part of MLJModelInterface 0.3.7) which provides a way for the implementer of MLJ's model interface to articulate: (i)  how user-data (eg, a table) is transformed into some model-specific representation needed by the core algorithm (eg, a matrix) -  the implementer overloads `reformat`; and (ii) how one can construct samples of the model-specific representation given a vector of observation indices - the implementer overloads `selectrows`.  Detailed responsibilities of this optional "data front-end" implementation are given in [this proposed doc PR](https://github.com/alan-turing-institute/MLJ.jl/blob/mlj2/docs/src/adding_models_for_general_use.md#implementing-a-data-front-end).

This PR introduces optional caching of the model-specific data representations for machines and for subsamples of these representations, which could speed up computations (at the cost of larger memory allocations). Machines have a new type parameter `C` which determines whether or not they buy into data caching, with `C=true` the default. A user opts-out during machine construction, as in `mach = machine(model, X, y, cache=false)`. 

Some key implementation details to help with a review:

- machines have a new type parameter `C` (see above) defaulting to `true`, except for composite models, where caching provides no obvious advantage.  (Machines *within* in a learning network defining the composite may still cache with some benefit.)
- machines have new fields `data` and `resampled_data` for caching model-specific representations for the what is returned by `[arg() for arg in args()]` (recall `args` is a tuple of `Source` or `Node` objects). We will refer to the latter as *user-supplied data*.
- the logic of `fit_only!` at https://github.com/alan-turing-institute/MLJBase.jl/blob/2eda522877451fdd26838faf279ac74c5a5a837c/src/machines.jl#L417 is adapted to apply `reformat` to the user-supplied data when `fit!` is first called, or the user-supplied data has changed because of upstream changes in a learning network. This data is cached for future calls to `fit!`; `data` is resampled according to the `rows` specified in the `fit!` call and also cached. Future calls to `fit!` only perform `reformat` and `selectrows` as needed. All this assumes `C=true` and otherwise `reformat` and `selectrows` are called every time and nothing is cached. 
- when an operation, `predict`, say, is called on a machine with the `rows=....` syntax, then, assuming the machine has caching enabled, `selectrows` is applied to the `data` cached, avoiding a call to `reformat`. If new data `Xnew` is supplied instead, as in `predict(mach, Xnew)`, then naturally `reformat` is then necessary. The relevant changes are in src/operations.jl
- in resampling, instead of calling `predict(mach, Xtest)` as currently, we call `predict(mach, rows=test)`. This avoids one call to `reformat`. However, there is one case distinction: the uncommon case where one of the measures has `is_feature_dependent=true` (for measures with `X` in the signature), in which case `Xtest` must be still be created, using the slow `selectrows(X, test)` fallback.  
- Add interface points for specifying `cache=...` in machine constructors, and the `Resampled` objects. An interface point for `cache` in `TunedModels` is also needed and addressed in [this MLJTuning PR](https://github.com/alan-turing-institute/MLJTuning.jl/pull/107) which passes locally (Julia 1.5).


ablaom:
cc @OkonSamuel @tlienart @yalwan-iqvia @jeremiedb

OkonSamuel (approved):
LGTM

PR #508, opened 2021-01-26
For a 0.17.0 release

### LINKED ISSUE #495: Decouple model traits from measure traits?
Presently there are some measure traits which are overloaded model traits. Specifically these are:
`name`, `target_scitype`, `supports_weights`, `prediction_type`, and `docstring`. To this list, [this PR](https://github.com/alan-turing-institute/MLJBase.jl/pull/464) would require us to add `supports_sample_weights` (currently a measure trait but needs to be a model trait as well).

**edit** Another method that is used by both models and traits is `info`

The problem with this state of affairs is that model traits are defined in MLJModelInterface, but we plan to have StatisticalMeasures as a separate packages (see #416). Although MLJModelInterface is very lightweight, to have it as a dependency of StatisticalMeausures.jl is like to jeopardise the latter's adoption as a general use package, independent of the MLJ stack. 

Here are two possible resolutions:

1.  Introduce a tiny ultra-lightweight package `MachineLearningMethodStubs.jl` (or whatever) to contain only stubs for traits used throughout the MLJ stack. It becomes a common dependency of MLJModelInterface and the eventual StatisticalMeasures.jl package. 

2. We move the current measures code into a module (or make it a package already) and keep its traits private. That is, their use will need to be qualified in, say, MLJBase and elsewhere.

Probably option 1 is the least disruptive; we could do this and still postpone the creation of `StatisticalMeasures`, which I expect to be a little painful. 

Thoughts or other ideas anyone?

cc @DilumAluthge @ven-k @tlienart

ablaom:
Some code re-organization:

- Import all model and measure traits from new base package StatisticalTraits.jl, which are programatically re-exported. This resolves #495
- Get some utilities into StatisticalTraits.jl, previously defined locally
- (**breaking for users**) `measures(matching(...))` facility removed; private `info_dic` method and `matching` method moved to MLJModels

Also added optimisations for models buying into a new data front-end #501 . This may increase the default memory footprint of MLJBase operations.


ablaom:
@JuliaRegistrator register

PR #509, opened 2021-01-29
Bump [compat] MLJModelInterface = "^0.4"; rm redundant exports

ablaom:
Turns out the more liberal compat causes issues.


ablaom:
OpenML currently slow and causing a build fail 😢

PR #526, opened 2021-03-24
Remove code now moved to MLJSerialization and MLJOpenML

giordano:
MLJBase v0.17.7:
```julia
julia> @time using MLJBase
  4.388609 seconds (7.48 M allocations: 476.401 MiB, 5.80% gc time, 0.12% compilation time)
```
with this PR:
```julia
julia> @time using MLJBase
  2.368572 seconds (4.08 M allocations: 264.086 MiB, 3.20% gc time, 0.17% compilation time)
```

Ref: #416


ablaom:
Before merging this PR:

- [ ] Register MLJOpenML

- [ ] Register MLJSerialization

ablaom:
On second thoughts, it looks like we need this minor release first. And then bump the [compat] requirements of the new packages before registering.  For currently, I get

```
julia> using MLJSerialization
[ Info: Precompiling MLJSerialization [17bed46d-0ab5-4cd4-b792-a5c4b8547c6d]
WARNING: Method definition save(Union{IO, String}, MLJBase.Machine{M, C} where C where M<:MLJModelInterface.Model) in module MLJBase at /Users/anthony/Dropbox/Julia7/MLJ/MLJBase/src/machines.jl:743 overwritten in module MLJSerialization at /Users/anthony/Dropbox/Julia7/MLJ/MLJSerialization/src/machines.jl:79.
  ** incremental compilation may be fatally broken for this module **
```

PR #540, opened 2021-04-22
Run CI on Julia nightly

DilumAluthge:
This will be useful for CompatHelper. On Julia nightly, CompatHelper workflows will now fail if they are not able to use the new version of the dependency.

This feature requires Julia 1.7, and thus is currently only available in Julia nightly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
         version:
           - '1.0'
           - '1' # automatically expands to the latest stable 1.x release of Julia.
+          - 'nightly'
         os:
           - ubuntu-latest
         arch:


DilumAluthge:
This will make sure that if CompatHelper opens a PR, but the new version of the dependency is being held back by one thing else, then the CI job will fail on Julia nightly.

ablaom:
How will I know that is the reason, and not some other problem with nightly itself? Will I have to dig into the log?

DilumAluthge:
Right now, you will need to look at the log unfortunately.

DilumAluthge:
Of course, this change will be available in Julia 1.7. So once Julia 1.7 is released, if you see that CI is failing on Julia 1.7, you'll know that there is an actual problem (since Julia 1.7 will be stable, and thus not susceptible to spurious problems).

PR #561, opened 2021-05-28
Add method `partition(::AbstractMatrix, ...)`

### LINKED ISSUE #361: Generalize `partition` to split any table or matrix or vector, not just vector of integers
So one could do, for example

```julia
Xtrain, Xtest = partition(X, 0.7)
ytrain, ytest = partition(y, 0.7)
```

if so inclined.

A quick look at the code suggest this wouldn't be too hard to fix. 

cc: @mlonig

giordano:
Also, relax signature of `partition(::AbstractVector, ...)` to allow any
vectors, not just vectors of integers.

Fix #361.

### DIFF
--- a/src/data/data.jl
+++ b/src/data/data.jl
@@ -81,14 +81,17 @@ function _partition(rows, fractions, raw_stratify::AbstractVector)
 end
 
 """
-    partition(X::Union{AbstractVector,AbstractMatrix}, fractions...;
+    partition(X, fractions...;
               shuffle=nothing, rng=Random.GLOBAL_RNG, stratify=nothing)
 
 Splits the vector or matrix `X` into a tuple of vectors or matrices whose
 numbers of rows are given by the corresponding `fractions` of
 `length(nrows(X))`, where valid fractions are in (0,1) and sum up to less than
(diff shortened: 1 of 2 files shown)


ablaom:
@giordano Thanks for this.

It remains to relax the signature altogether to include the tables case. As tables are defined by a trait, we need trait dispatch here. Have a look at the implementation of `MLJModelInterface.matrix` for a similar case: https://github.com/alan-turing-institute/MLJModelInterface.jl/blob/275beda8e90e7946cd408ab84ca59c8efc34b407/src/data_utils.jl#L26 You can use `Tables.istable` instead of `vtrait` (MLJModelInterface does not have `Tables` as a dep).

giordano:
Ok, thanks, I relaxed the signature, so now the argument is `X::Any` and we check at runtime whether `X` is a matrix/table.

I don't think having a separate method would speed up anything: `partition` itself for a matrix isn't superfast (for a small matrix it takes time of the order of units of microseconds), which dwarfs the time to look up a specific method, and the `X isa AbstractMatrix` check should hopefully make it as good as having a dedicated method.

ablaom (approved):
Looks good to me, thanks!

PR #564, opened 2021-06-02
Stacking

olivierlabayle:
Following on [this issue](https://github.com/alan-turing-institute/MLJ.jl/issues/76), here is a suggested implementation of the stack.

### Implementation

I used the learning network API and provide 2 types `DeterministicStack` and `ProbabilisticStack`. The implementation mostly follows the issue's suggestions. 

### Documentation

I have added a docstring to the `Stack` constructor with a working example (provided the required dependencies are installed) to illustrate.

### Testing

The basic functionalities should be tested.

I provide a performance test in the continuous deterministic case that shows that when the true model is not part of the library, the stack will improve the results. I've also looked at [vecstack in python](https://github.com/vecxoz/vecstack) and [sl3 in R](https://github.com/tlverse/sl3) to find new theoretical testing possibilities but couldn't easily find anything.

Let me know if you have ideas I'll be happy to enrich the tests. I think some more models (from the `test/_models` suite), like a logistic regression, would probably be needed to improve the current tests though.

### Enhancements

Some future ideas:

- DeterministicClassifier
- Pass the pre_judge_transform as a hyperparameter
- Optional Internal evaluation of the models in the library
- Option to include inputs `X` as columns of the metalearner input (in addition to output of  the base learners), when these are continuous (**added by** @ablaom)

Overall I'm happy to discuss this and to provide any required material may it be tests, doc and code correction.


test/composition/models/stacking.jl
```diff
@@ -0,0 +1,141 @@
...
+using Test
+using MLJBase
+using MLJModelInterface
+using ..Models
+using Random
+
+import Distributions
+
+
+rng = Random.seed!(1234)
```
ablaom, line 12: Very minor point: Sadly, Julia's global RNG is not uniform across versions. To address this, StableRNGs packages was introduced for package testing.  So instead of this line, do

```julia
using StableRNGs
rng = StableRNGs.StableRNG(1234)
```

test/composition/models/stacking.jl
```diff
@@ -0,0 +1,141 @@
...
+        mach = machine(model, X, y)
+        ev = evaluate!(mach; resampling=cv, measure=measure, check_measure=false)
+        push!(results, ev.measurement[1])
+    end
+    results
+end
+
+
+@testset "Testing Stack on Continuous target" begin
+    X, y = make_regression(500, 5)
```
ablaom, line 28:
```julia
X, y = make_regression(500, 5, rng=rng)
```

src/composition/models/stacking.jl
```diff
@@ -0,0 +1,276 @@
...
+
+    target_scitype(stack.metalearner) <: Union{AbstractArray{<:Continuous}, AbstractArray{<:Finite}} ||
+        throw(ArgumentError("The metalearner should have target_scitype: 
+                $(Union{AbstractArray{<:Continuous}, AbstractArray{<:Finite}})"))
+
+    return message
+end
+
+
+Base.propertynames(::Stack{modelnames, <:Any, <:Any}) where modelnames = tuple(:cv_strategy, :metalearner, :models, modelnames...)
```
ablaom, line 154: I found it a little confusing that `:models ` is included as property. So, for example, on construction of first stack in test, the REPL displays it thus:

```julia
julia> mystack = Stack(;metalearner=FooBarRegressor(),
       cv_strategy=CV(;nfolds=3),
       models...)

DeterministicStack(
    cv_strategy = CV(
            nfolds = 3,
            shuffle = false,
            rng = Random._GLOBAL_RNG()),
    metalearner = FooBarRegressor(
            lambda = 0.0),
    models = (DeterministicConstantRegressor @659, DecisionTreeRegressor @604, FooBarRegressor @931, FooBarRegressor @570),                                                
    constant = DeterministicConstantRegressor(),
    decisiontree = DecisionTreeRegressor(
            pruning_purity_threshold = 0.0,
            max_depth = -1,
            min_samples_leaf = 5,
            min_samples_split = 2,
            min_purity_increase = 0.0,
            n_subfeatures = 0,
            post_prune = false),
    ridge_lambda = FooBarRegressor(
            lambda = 0.1),
    ridge = FooBarRegressor(
            lambda = 0.0)) @079
```

So, for example, `DeterministicConstantRegressor @659` appears twice in this display. 

Would it be better to keep `models` private? One can still access it  internally using the syntax `getfield(mystack, :models)`.

src/composition/models/stacking.jl
```diff
@@ -0,0 +1,276 @@
...
+        end
+    end
+    return Unknown
+end
+
+
+mutable struct DeterministicStack{modelnames, input_scitype, target_scitype} <: DeterministicComposite
+   models::NTuple{<:Any, Supervised}
+   metalearner::Deterministic
+   cv_strategy::Union{CV, StratifiedCV}
```
ablaom, line 28: For consistency with `TunedModel`, `IteratedModel`, `evaluate!` and elsewhere, I suggest we rename `cv_strategy`, `resampling`. 

Also, while `CV()` is a good default,  do you think we need to restrict to `Union{CV, Stratified}`?  For example, a user may want to specify [explicit traint/test pairs](https://alan-turing-institute.github.io/MLJ.jl/dev/evaluating_model_performance/#User-specified-train/test-sets) or a [custom resampling strategy](https://alan-turing-institute.github.io/MLJ.jl/dev/evaluating_model_performance/#Custom-resampling-strategies).  

Technical note: Yes, performance is better if you annotate each field with a concrete type or small Union, or introduce an associated type parameter. However, in this use case, and many other MLJ meta-algorithms, the performance gains will rarely be noticed and you sacrifice genericity or, in the case of introducing a type parameter, mutability).

src/composition/models/stacking.jl
```diff
@@ -0,0 +1,276 @@
...
+                library...)
+
+mach = machine(stack, X, y)
+evaluate!(mach; resampling=Holdout(), measure=rmse)
+
+```
+
+"""
+function Stack(;metalearner=nothing, cv_strategy=CV(), named_models...)
+    metalearner === nothing && throw(ArgumentError("metalearner argument should be overrided"))
```
ablaom, line 122: For consistency with similar messages, I suggest "You need to specify `metalearner=...`"as the message.

src/composition/models/stacking.jl
```diff
@@ -0,0 +1,276 @@
...
+    if metalearner isa Deterministic
+        stack =  DeterministicStack(modelnames, models, metalearner, cv_strategy)
+    elseif metalearner isa Probabilistic
+        stack = ProbabilisticStack(modelnames, models, metalearner, cv_strategy)
+    else
+        throw(ArgumentError("The metalearner should be a subtype 
+                    of $(Union{Deterministic, Probabilistic})"))
+    end
+    MMI.clean!(stack)
+
```
ablaom, line 137: For future proofing, please make this

```julia
message = MMI.clean!(stack)
isempty(message) || @warn message
```

test/composition/models/stacking.jl
```diff
@@ -0,0 +1,141 @@
...
+    @test nrows(getfield(fp, :constant)) == 4
+    @test nrows(getfield(fp, :decisiontree)) == 4
+    @test nrows(getfield(fp, :ridge)) == 4
+    @test nrows(getfield(fp, :ridge_lambda)) == 4
+
+    # Testing prediction is Deterministic
+    @test predict(mach) isa Vector{Float64}
+    end
+
+    @testset "Testing ProbabilisticStack" begin
```
ablaom, line 84: minor organization point: Perhaps you meant this sub `@testset` block to appear instead as a stand-alone block beginning at the end of the surrounding block? That is, did you really intend a nested block here?

test/composition/models/stacking.jl
```diff
@@ -0,0 +1,141 @@
...
+        mach = machine(mystack, X, y)
+        fit!(mach)
+        @test predict(mach) isa Vector{Distributions.Cauchy{Float64}}
+
+    end
+
+end
+
+@testset "Testing ProbabilisticStack on Finite target" begin
+    X, y = make_blobs()
```
ablaom, line 108: `X, y = make_blobs(rng=rng)` See earlier comment.

ablaom:
Post-merge to-do list (not assigned):

- [ ] add page to MLJ manual, to appear after [this subsection](https://alan-turing-institute.github.io/MLJ.jl/dev/composing_models/#Homogeneous-Ensembles)
- [ ] update [this](https://alan-turing-institute.github.io/DataScienceTutorials.jl/getting-started/stacking/) DataScienceTutorials tutorial. 
- [ ] update wow.jl/wow.md/wow.ipynb in [2020 Workshop](https://github.com/ablaom/MachineLearningInJulia2020)
- [ ] Make a Julia Discourse [ANN] announcement (together with MLJIterations?)

src/composition/models/stacking.jl
ablaom: shortcut: You can do `Stack{modelnames}` instead of `Stack{modelnames, <:Any, <:Any}`.

src/composition/models/stacking.jl
```diff
@@ -0,0 +1,276 @@
...
+   cv_strategy::Union{CV, StratifiedCV} 
+   function DeterministicStack(modelnames, models, metalearner, cv_strategy)
+        target_scitype = MMI.target_scitype(metalearner)
+        input_scitype = glb(models)
+        return new{modelnames, input_scitype, target_scitype}(models, metalearner, cv_strategy)
+   end
+end
+
+mutable struct ProbabilisticStack{modelnames, input_scitype, target_scitype} <: ProbabilisticComposite
+    models::NTuple{<:Any, Supervised}
```
ablaom, line 37: I suggest we allow replacement of any model in the "library" `models`. This might be useful in "architecture search" and there will be neglible performance loss for use cases I can imagine. Currently, this is not possible: 

```julia
julia> mystack.constant
ConstantClassifier(
    testing = false,
    bogus = 0) @272

julia> mystack.constant = ConstantRegressor()
ERROR: type ProbabilisticStack has no field constant
Stacktrace:
 [1] setproperty!(m::MLJBase.ProbabilisticStack{(:constant, :decisiontree, :knn), Table{var"#s46"} where var"#s46"<:(AbstractVector{var"#s9"} where var"#s9"<:Continuous), AbstractVector{var"#s213"} where var"#s213"<:Finite}, s::Symbol, value::ConstantRegressor{Distributions.Normal})                                                                     
   @ Main ./REPL[74]:1
 [2] top-level scope
   @ REPL[76]:1
```

To fix this requires two steps: 
- make `models` a `Vector` instead of tuple; and
- overload `Base.setproperty!`, something like this:

```julia
function Base.setproperty!(stack::Stack{modelnames},
                                        _name::Symbol,
                                        val) where modelnames
    _name === :metalearner && return setfield!(stack, :metalearner, val)
    _name === :cv_strategy && return setfield!(stack, :cv_strategy, val)
    idx = findfirst(==(_name), modelnames)
    idx isa Nothing || return getfield(stack, :models)[idx] = val
    error("type Stack has no field $name")
end
```

ablaom (changes requested):
@olivierlabayle Thanks for this exciting PR and all the work I know went into it!

I've started reviewing the code today. So far the comments are pretty minor. I'll let you know when I am done but feel free to address these comments in the meantime.

test/composition/models/stacking.jl
olivierlabayle: Ok I guess I have to substitute the `rng=rng` everywhere then, including the `CV()` objects?

src/composition/models/stacking.jl
olivierlabayle: Completely agree, I have been thinking about this this morning, my future use cases will probably depend on providing a custom resampling strategy. Thanks for the note!

src/composition/models/stacking.jl
olivierlabayle: Allright, my current understanding of properties is that they just represent fields with potential extra functionalities so I thought it was expected behavior to have it for all fields. It's also because think there is a basic fallback that returns the field.

test/composition/models/stacking.jl
olivierlabayle: On this specific synthetic dataset I've noticed that I also need to pass `shuffle=false`. Else there is a call to `Random.seed!()` [here](https://github.com/alan-turing-institute/MLJBase.jl/blob/fb183dc07d1b1c10bfc3aef23f7b61eae53abfc7/src/utilities.jl#L133).

Is that intended?

test/composition/models/stacking.jl
olivierlabayle: In fact I had intended of having the following organization levels:

- Continuous target
   - DeterministicStack
   - ProbabilisticStack
   
- Finite Target
   - DeterministicStack (Which is not implemented yet)
   - ProbabilisticStack

because each sub testset uses the same dataset. I can still pull it out if you prefer and declare the dataset at the main level.

src/composition/models/stacking.jl
olivierlabayle: Great, and Thanks for the code snippet which works "as is"!

olivierlabayle:
> @olivierlabayle Thanks for this exciting PR and all the work I know went into it!
> 
> I've started reviewing the code today. So far the comments are pretty minor. I'll let you know when I am done but feel free to address these comments in the meantime.

Thanks for these first nice and relevant comments. I don't know how you like to go about reviews so I just added a 👍 where I think I've addressed your comments and added some replies along the way. I let you the liberty to resolve the conversations.

test/composition/models/stacking.jl
ablaom: No, no. All good.

src/composition/models/stacking.jl
ablaom: Your new alternative may have an advantage. However, my suggestion is along the lines of usage elsewhere and is probably going to be less confusing. Do you mind changing it?

test/composition/models/stacking.jl
ablaom: Good catch! That's a [bug](https://github.com/alan-turing-institute/MLJBase.jl/issues/567). 🙏🏾

test/composition/models/stacking.jl
ablaom: Good catch! That's a [bug](https://github.com/alan-turing-institute/MLJBase.jl/issues/567) 🙏🏾

src/composition/models/stacking.jl
```diff
@@ -0,0 +1,286 @@
...
+end
+
+
+function testrows(X::AbstractNode, folds::AbstractNode, nfold)
+    node((XX, ff) -> selectrows(XX, ff[nfold][2]), X, folds)
+end
+
+
+pre_judge_transform(ŷ::Node, ::Type{<:Probabilistic}, ::Type{<:AbstractArray{<:Finite}}) = 
+    node(ŷ->pdf.(ŷ, levels.(ŷ)), ŷ)
```
ablaom, line 214: This is absolutely correct, but will implement a fallback for pdf broadcasting that is slow. I suggest 

```julia
node(yhat -> pdf(yhat, levels(first(yhat))), yhat)
```
However, you will need to pull in the latest dev branch to get the test model to work as it should (https://github.com/alan-turing-institute/MLJBase.jl/pull/566). Or, you can make the single line change yourself: https://github.com/alan-turing-institute/MLJBase.jl/pull/566/commits/ba2b0021ce2eb47b54e1e554d3363122cb036740

ablaom:
Nightly fails look unrelated. I think you can ignore.

src/composition/models/stacking.jl
```diff
@@ -0,0 +1,286 @@
...
+
+
+```julia
+using MLJ
+using MLJDecisionTreeInterface
+using MLJLinearModels
+using NearestNeighborModels
+using EvoTrees
+using MLJXGBoostInterface
+
```
ablaom, line 99: In a standard workflow MLJ automatically imports the needed packages. So instead of those `using` statements, I suggest

```julia
DecisionTreeRegressor = @load DecisionTreeRegressor pkg=DecisionTree
EvoTreeRegressor = @load EvoTreeRegressor
XGBoostRegressor = @load XGBoostRegressor
KNNRegressor = @load KNNRegressor pkg=NearestNeighborModels
LinearRegressor = @load LinearRegressor pkg=MLJLinearModels
````

**Review by ablaom — COMMENTED** [review]
The doc string is extremely good. My own inclination would be simplify the example, skipping the pipeline, and perhaps calling the constructor without the intermediate `library = ...` step. Then port the existing example into the MLJ manual entry, with some demonstration of `fitted_params` access.

However, given the fine work here, I think you should make the call. Ie, if you prefer, leave as is.

File: src/composition/models/stacking.jl [hunk 12]
```diff
@@ -0,0 +1,286 @@
...
+        stack =  DeterministicStack(modelnames, models, metalearner, resampling)
+    elseif metalearner isa Probabilistic
+        stack = ProbabilisticStack(modelnames, models, metalearner, resampling)
+    else
+        throw(ArgumentError("The metalearner should be a subtype 
+                    of $(Union{Deterministic, Probabilistic})"))
+    end
+    message = MMI.clean!(stack)
+    isempty(message) || @warn message
+
```
ablaom, line 139: To be thorough we should probably add a check here that prohibits user from using `:resampling` or `:metalearner` as names for their base learners. That is, `modelnames` should not include either of these.

test/composition/models/stacking.jl
```diff
@@ -0,0 +1,165 @@
...
+    @test mystack.decisiontree isa DecisionTreeRegressor
+    
+    @test target_scitype(mystack) == target_scitype(FooBarRegressor())
+    @test input_scitype(mystack) == input_scitype(FooBarRegressor())
+
+    # Testing fitted_params results are easily accessible for each
+    # submodel. They are in order of the cross validation procedure.
+    # Here 3-folds then 3 machines + the final fit
+    mach = machine(mystack, X, y)
+    fit!(mach)
```
ablaom, line 73: To reduce noise in test logging, make this and similar `fit!(mach, verbosity=0)`

src/composition/models/stacking.jl
```diff
@@ -0,0 +1,286 @@
...
+#######################################
+################# Fit ################# 
+#######################################
+"""
+    fit(m::Stack, verbosity::Int, X, y)
+"""
+function fit(m::Stack, verbosity::Int, X, y)
+    n = nrows(y)
+
+    X = source(X)
```
ablaom, line 231: You probably copied this from somewhere else, but I think using the same variable name for objects of different types can lead to performance issues (`X = source(X)`). Maybe the compiler is now smart enough - but I'm not sure. So maybe `Xs = source(X)`, etc is better.

ablaom:
Okay, been working on a new test and found what I think is a bug:

```julia
x = Float64[1, 1, 2, 2, 3, 3]
X = (x = x,)
y = coerce(['a', 'b', 'b', 'c', 'a', 'a'], Multiclass)

model1 = KNNClassifier(K=2)
model2 = ConstantClassifier()
judge = KNNClassifier(K=3)

stack = Stack(metalearner=judge,
       model1=model1,
       model2=model2,
       resampling=CV(nfolds=3))

julia> fit!(mach, verbosity=0)
┌ Error: Problem fitting the machine Machine{KNNClassifier,…} @484. 
└ @ MLJBase ~/Dropbox/Julia7/MLJ/MLJBase/src/machines.jl:484
[ Info: Running type checks... 
┌ Warning: The scitype of `X`, in `machine(model, X, ...)` is incompatible with `model=KNNClassifier @161`:                                                                        
│ scitype(X) = Table{AbstractVector{AbstractVector{Continuous}}}
│ input_scitype(model) = Table{var"#s46"} where var"#s46"<:(AbstractVector{var"#s9"} where var"#s9"<:Continuous).
└ @ MLJBase ~/Dropbox/Julia7/MLJ/MLJBase/src/machines.jl:103
[ Info: It seems an upstream node in a learning network is providing data of incompatible scitype. See above. 
┌ Error: Problem fitting Machine{KNNClassifier,…} @484
└ @ MLJBase ~/Dropbox/Julia7/MLJ/MLJBase/src/machines.jl:574
┌ Error: Problem fitting the machine Machine{ProbabilisticStack{,…},…} @642. 
└ @ MLJBase ~/Dropbox/Julia7/MLJ/MLJBase/src/machines.jl:484
[ Info: Running type checks... 
[ Info: Type checks okay. 
ERROR: TaskFailedException
...
```

I went and built the learning network (extracted the code in `fit`) and found that `fit!(metamach)` fails. Looking closer, I believe the problems is with `Zval`, which ought to be delivering a table with six float columns (three probabilities per base model). However, I get something weirder with just two "columns" which aren't actually columns:

```julia
julia> Zval().x1
6-element view(::Matrix{Vector{Float64}}, :, 1) with eltype Vector{Float64}:
 [0.0, 0.5, 0.5]
 [0.0, 0.5, 0.5]
 [1.0, 0.0, 0.0]
 [1.0, 0.0, 0.0]
 [0.0, 0.5, 0.5]
 [0.0, 0.5, 0.5]

julia> Zval().x2
6-element view(::Matrix{Vector{Float64}}, :, 2) with eltype Vector{Float64}:
 [0.5, 0.25, 0.25]
 [0.5, 0.25, 0.25]
 [0.75, 0.25, 0.0]
 [0.75, 0.25, 0.0]
 [0.25, 0.5, 0.25]
 [0.25, 0.5, 0.25]
```

src/composition/models/stacking.jl
olivierlabayle: Out of curiosity, what is the benefit of `@load` compared to a single `using NearestNeighborModels:KNNRegressor` for instance?

src/composition/models/stacking.jl
olivierlabayle: Sorry I had not understood what you initially meant, I hope this solves it.

src/composition/models/stacking.jl
olivierlabayle: Agreed for simplifying the example, it's probably too heavy for a docstring.

Regarding the later note, I'm not sure I understand how one could build a stack with `:resampling` or `:metalearner` as a modelname given the current external constructor.  I can see some issues if one tries to use the internal constructor like this:
```
  modelnames = (:resampling, :metalearner)
  models = [DeterministicConstantRegressor(), FooBarRegressor(;lambda=0)]
  metalearner = DeterministicConstantRegressor()
  resampling = CV()

  MLJBase.DeterministicStack(modelnames, models, metalearner, resampling)
```

It would then also require to move the `clean!` statement to the internal constructors I guess. Is that what you have in mind?

src/composition/models/stacking.jl
olivierlabayle: Thanks for the tip, I addressed the change to be completely sure. 

I didn't delve in it too much but I think I understood from [this conversation](https://discourse.julialang.org/t/type-unstable-function-because-same-variable-name-used-twice/58810) that renaming variables is  probably "fine". I also tried `@code_warntype` illustrated [in the doc](https://docs.julialang.org/en/v1/manual/performance-tips/#man-code-warntype). It seems that the current `fit(m::Stack,...)` is type stable, the only union types I've seen are simply coming from the  `models...` argument of the stack which can theoretically be `nothing` from the compiler's perspective. Of course there is not much we can do and, again from the doc,  this specific usecase seems to be fine.

olivierlabayle:
> Okay, been working on a new test and found what I think is a bug:
> 
> ```julia
> x = Float64[1, 1, 2, 2, 3, 3]
> X = (x = x,)
> y = coerce(['a', 'b', 'b', 'c', 'a', 'a'], Multiclass)
> 
> model1 = KNNClassifier(K=2)
> model2 = ConstantClassifier()
> judge = KNNClassifier(K=3)
> 
> stack = Stack(metalearner=judge,
>        model1=model1,
>        model2=model2,
>        resampling=CV(nfolds=3))
> 
> julia> fit!(mach, verbosity=0)
> ┌ Error: Problem fitting the machine Machine{KNNClassifier,…} @484. 
> └ @ MLJBase ~/Dropbox/Julia7/MLJ/MLJBase/src/machines.jl:484
> [ Info: Running type checks... 
> ┌ Warning: The scitype of `X`, in `machine(model, X, ...)` is incompatible with `model=KNNClassifier @161`:                                                                        
> │ scitype(X) = Table{AbstractVector{AbstractVector{Continuous}}}
> │ input_scitype(model) = Table{var"#s46"} where var"#s46"<:(AbstractVector{var"#s9"} where var"#s9"<:Continuous).
> └ @ MLJBase ~/Dropbox/Julia7/MLJ/MLJBase/src/machines.jl:103
> [ Info: It seems an upstream node in a learning network is providing data of incompatible scitype. See above. 
> ┌ Error: Problem fitting Machine{KNNClassifier,…} @484
> └ @ MLJBase ~/Dropbox/Julia7/MLJ/MLJBase/src/machines.jl:574
> ┌ Error: Problem fitting the machine Machine{ProbabilisticStack{,…},…} @642. 
> └ @ MLJBase ~/Dropbox/Julia7/MLJ/MLJBase/src/machines.jl:484
> [ Info: Running type checks... 
> [ Info: Type checks okay. 
> ERROR: TaskFailedException
> ...
> ```
> 
> I went and built the learning network (extracted the code in `fit`) and found that `fit!(metamach)` fails. Looking closer, I believe the problems is with `Zval`, which ought to be delivering a table with six float columns (three probabilities per base model). However, I get something weirder with just two "columns" which aren't actually columns:
> 
> ```julia
> julia> Zval().x1
> 6-element view(::Matrix{Vector{Float64}}, :, 1) with eltype Vector{Float64}:
>  [0.0, 0.5, 0.5]
>  [0.0, 0.5, 0.5]
>  [1.0, 0.0, 0.0]
>  [1.0, 0.0, 0.0]
>  [0.0, 0.5, 0.5]
>  [0.0, 0.5, 0.5]
> 
> julia> Zval().x2
> 6-element view(::Matrix{Vector{Float64}}, :, 2) with eltype Vector{Float64}:
>  [0.5, 0.25, 0.25]
>  [0.5, 0.25, 0.25]
>  [0.75, 0.25, 0.0]
>  [0.75, 0.25, 0.0]
>  [0.25, 0.5, 0.25]
>  [0.25, 0.5, 0.25]
> ```

Thank you for that. 

Indeed, by inspection of the commit history, the origin of the problem was `node(ŷ->pdf.(ŷ, levels.(ŷ)), ŷ)` which seems to be not fine. I'm not sure if it is due to a bad interpretation from my part or if this needs to be solved upstream though. Of course this is solved by your previous comment however this made me realize the `fit!` function is probably too big. I extracted a [function](https://github.com/olivierlabayle/MLJBase.jl/blob/cfc85b7cece3a9080101fdb4ab4c600bfdaf9e73/src/composition/models/stacking.jl#L212) out of it and built on your test case to test it [here](https://github.com/olivierlabayle/MLJBase.jl/blob/cfc85b7cece3a9080101fdb4ab4c600bfdaf9e73/test/composition/models/stacking.jl#L177). Hopefully this also increases the test coverage, please let me know your thoughts.

src/composition/models/stacking.jl
ablaom: > Regarding the later note, I'm not sure I understand how one could build a stack with :resampling or :metalearner as a modelname given the current external constructor. 

No you are right! Nevermind.

src/composition/models/stacking.jl
ablaom: The only benefit is one of convenience. The API allows interface  code to live anywhere. Sometimes this is in the same package providing the core algorithm, sometimes in a separate "interface only" package, sometimes it's in a submodule, and so forth. The general user needn't worry about such details. They query the model registry as described in [the manual](https://alan-turing-institute.github.io/MLJ.jl/dev/model_search/) and then simple load the model by name, specifying the core package to resolve name ambiguities. 

However, if you know exactly where the model struct is defined, there's nothing wrong with your approach. 

See also https://alan-turing-institute.github.io/MLJ.jl/dev/loading_model_code/

ablaom:
> made me realize the fit! function is probably too big. I extracted a function out of it and built on your test case to test it here. Hopefully this also increases the test coverage, please let me know your thoughts.

I think that was a great move, thanks.

src/composition/models/stacking.jl
```diff
@@ -0,0 +1,282 @@
...
+    return Unknown
+end
+
+
+mutable struct DeterministicStack{modelnames, input_scitype, target_scitype} <: DeterministicComposite
+   models::Vector{Supervised}
+   metalearner::Deterministic
+   resampling
+   function DeterministicStack(modelnames, models, metalearner, resampling)
+        target_scitype = MMI.target_scitype(metalearner)
```
ablaom, line 30: I think I was a bit sloppy in the design brief. It seems to me we should probably make the stack `target_scitype` the greatest lower bound of the target scitypes of the base learners and metalearner, no? At present there is nothing stopping the user specifying a regressor as a base model when the metalearner is a classifier, which doesn't make sense. 

To this end, you may want to refactor `glb(models)` to new signature `glb(types...)`.

ablaom:
I'd like to add the following integration test. It is currently failing, although I'm less confident in this case that the problem is with the PR. Maybe I'm doing something wrong here? @olivierlabayle Be great if you could have a look. 

Ultimately it might make sense to increase `N` and swap out the `ConstantRegressor` for something else, but I've kept it simple for now. 

```julia
@testset "an integration test for stacked classification" begin

    # We train a stack by hand and compare with the canned version
    # `Stack(...)`. There are two base learners, with 3-fold
    # cross-validation used to construct the out-of-sample base model
    # predictions. 

    rng = StableRNGs.StableRNG(123)

    probs(y) = pdf(y, levels(first(y)))

    # data:
    N = 2
    X = (x = rand(rng, 3N), )
    y = coerce(rand("abc", 3N), Multiclass)

    # row splits:
    test1 = 1:N
    test2 = (N + 1):2N
    test3 = (2N + 1):3N
    train1 = (N + 1):3N
    train2 = vcat(1:N, (2N + 1):3N)
    train3 = 1:2N

    # base `model1`:
    model1 = KNNClassifier(K=2)
    mach1 = machine(model1, X, y)
    fit!(mach1, rows=train1, verbosity=0)
    y11 = predict(mach1, rows=test1) |> probs
    fit!(mach1, rows=train2, verbosity=0)
    y12 = predict(mach1, rows=test2) |> probs
    fit!(mach1, rows=train3, verbosity=0)
    y13 = predict(mach1, rows=test3) |> probs
    y1_oos = vcat(y11, y12, y13)
    mach1_full = machine(model1, X, y)
    fit!(mach1_full, verbosity=0)
    y1 = predict(mach1_full, X) |> probs

    # base `model2`:
    model2 = ConstantClassifier()
    mach2 = machine(model2, X, y)
    fit!(mach2, rows=train1, verbosity=0)
    y21 = predict(mach2, rows=test1) |> probs
    fit!(mach2, rows=train2, verbosity=0)
    y22 = predict(mach2, rows=test2) |> probs
    fit!(mach2, rows=train3, verbosity=0)
    y23 = predict(mach2, rows=test3) |> probs
    y2_oos = vcat(y21, y22, y23)
    mach2_full = machine(model2, X, y)
    fit!(mach2_full, verbosity=0)
    y2 = predict(mach1_full, X) |> probs

    # metalearner (`judge`):
    X_oos = MLJBase.table(hcat(y1_oos, y2_oos))
    judge = KNNClassifier(K=3)
    m_judge = machine(judge, X_oos, y)
    fit!(m_judge, verbosity=0)
    X_judge = MLJBase.table(hcat(y1, y2))
    yhat_matrix = predict(m_judge, X_judge) |> probs

    # alternatively, use stack:
    stack = Stack(metalearner=judge,
                  model1=model1,
                  model2=model2,
                  resampling=CV(nfolds=3))
    mach = machine(stack, X, y)
    fit!(mach, verbosity=0)
    yhat_matrix_stack = predict(mach, X) |> probs

    # compare:
    @test yhat_matrix ≈ yhat_matrix_stack
end
```

olivierlabayle:
> I'd like to add the following integration test. It is currently failing, although I'm less confident in this case that the problem is with the PR. Maybe I'm doing something wrong here? @olivierlabayle Be great if you could have a look.
> 
> Ultimately it might make sense to increase `N` and swap out the `ConstantRegressor` for something else, but I've kept it simple for now.
> 
> ```julia
> @testset "an integration test for stacked classification" begin
> 
>     # We train a stack by hand and compare with the canned version
>     # `Stack(...)`. There are two base learners, with 3-fold
>     # cross-validation used to construct the out-of-sample base model
>     # predictions. 
> 
>     rng = StableRNGs.StableRNG(123)
> 
>     probs(y) = pdf(y, levels(first(y)))
> 
>     # data:
>     N = 2
>     X = (x = rand(rng, 3N), )
>     y = coerce(rand("abc", 3N), Multiclass)
> 
>     # row splits:
>     test1 = 1:N
>     test2 = (N + 1):2N
>     test3 = (2N + 1):3N
>     train1 = (N + 1):3N
>     train2 = vcat(1:N, (2N + 1):3N)
>     train3 = 1:2N
> 
>     # base `model1`:
>     model1 = KNNClassifier(K=2)
>     mach1 = machine(model1, X, y)
>     fit!(mach1, rows=train1, verbosity=0)
>     y11 = predict(mach1, rows=test1) |> probs
>     fit!(mach1, rows=train2, verbosity=0)
>     y12 = predict(mach1, rows=test2) |> probs
>     fit!(mach1, rows=train3, verbosity=0)
>     y13 = predict(mach1, rows=test3) |> probs
>     y1_oos = vcat(y11, y12, y13)
>     mach1_full = machine(model1, X, y)
>     fit!(mach1_full, verbosity=0)
>     y1 = predict(mach1_full, X) |> probs
> 
>     # base `model2`:
>     model2 = ConstantClassifier()
>     mach2 = machine(model2, X, y)
>     fit!(mach2, rows=train1, verbosity=0)
>     y21 = predict(mach2, rows=test1) |> probs
>     fit!(mach2, rows=train2, verbosity=0)
>     y22 = predict(mach2, rows=test2) |> probs
>     fit!(mach2, rows=train3, verbosity=0)
>     y23 = predict(mach2, rows=test3) |> probs
>     y2_oos = vcat(y21, y22, y23)
>     mach2_full = machine(model2, X, y)
>     fit!(mach2_full, verbosity=0)
>     y2 = predict(mach1_full, X) |> probs
> 
>     # metalearner (`judge`):
>     X_oos = MLJBase.table(hcat(y1_oos, y2_oos))
>     judge = KNNClassifier(K=3)
>     m_judge = machine(judge, X_oos, y)
>     fit!(m_judge, verbosity=0)
>     X_judge = MLJBase.table(hcat(y1, y2))
>     yhat_matrix = predict(m_judge, X_judge) |> probs
> 
>     # alternatively, use stack:
>     stack = Stack(metalearner=judge,
>                   model1=model1,
>                   model2=model2,
>                   resampling=CV(nfolds=3))
>     mach = machine(stack, X, y)
>     fit!(mach, verbosity=0)
>     yhat_matrix_stack = predict(mach, X) |> probs
> 
>     # compare:
>     @test yhat_matrix ≈ yhat_matrix_stack
> end
> ```

Typos can be the worst things to debug 😄  1-->2

```
mach2_full = machine(model2, X, y)
fit!(mach2_full, verbosity=0)
y2 = predict(mach1_full, X) |> probs
```

I've added your test and replaced N = 200 and changed the constant classifier for a decision tree.

src/composition/models/stacking.jl
olivierlabayle: I think it makes sense to define the `target_scitype` as a glb of both models in the library and the metalearner. The question is, should we then raise if the `target_scitype` is then Unknown?  

This makes me realize that changing a model or the metalearner by `setproperty!` should then change the `input_scitype` or `target_scitype` of the stack. Eventually I guess it should change the parametric type of the stack too? Which I think is not what we want to do (not even sure if this is possible?). Maybe we should prevent from using `setproperty!` on the models and the metalearner? It should still be possible to change hyperparameters of the models themselves.

src/composition/models/stacking.jl
olivierlabayle: Allright, thx!

src/composition/models/stacking.jl
```diff
@@ -2,52 +2,55 @@
...
-end
 
-function glb(models)
-    for model in models
-        potential_glb = input_scitype(model)
-        if is_glb(potential_glb, models)
-            return potential_glb
-        end
+function glb(types...)
+    # If a lower bound is in the models then it is greatest
```
ablaom, line 7: # If a lower bound is in ~~the models~~ `types`  then it is greatest

test/composition/models/stacking.jl
```diff
@@ -234,7 +251,7 @@ end
     probs(y) = pdf(y, levels(first(y)))
 
     # data:
-    N = 200
+    N = 2000
```
ablaom, line 254: Probably `N = 200` is plenty big.

ablaom:
@olivierlabayle Ahh! Thank you so much for catching the typo. 😌

src/composition/models/stacking.jl
ablaom: No a type parameter cannot be changed, which I don't think is too big a restriction here. We should however allow mutations of component models that are consistent with the original stack scitypes (I suppose  "more permissive" model mutations could be allowed, but this is probably too sneaky). `clean!` should check consistency and throw an error otherwise. I'd throw an error in that case. Note that 99% of the time computing scitypes is not expensive, so recomputing every `fit!` call is no big deal, in my opinion.

If the target scitype or input scitype *computed at construction* is `Unknown`, then I'd issue `@warn`.

ablaom:
For completeness, please also add these trait declarations:

```julia
MLJBase.is_wrapper(::Type{<:Stack}) = true
MLJBase.load_path(::Type{<:ProbabilisticStack}) =
    "MLJTuning.ProbabilisticStack"
MLJBase.load_path(::Type{<:DeterministicStack}) =
    "MLJTuning.DeterministicStack"
MLJBase.package_name(::Type{<:Stack}) = "MLJBase"
MLJBase.package_uuid(::Type{<:Stack}) =
    "a7f614a8-145f-11e9-1d2a-a57a1082229d"
MLJBase.package_url(::Type{<:Stack}) =
    "https://github.com/alan-turing-institute/MLJBase.jl"
MLJBase.package_license(::Type{<:Stack}) = "MIT"
```

src/composition/models/stacking.jl
olivierlabayle: Ok, I think I've addressed the issue

src/composition/models/stacking.jl
```diff
@@ -133,17 +133,35 @@ function Stack(;metalearner=nothing, resampling=CV(), named_models...)
...
     return stack
 end
 
 
 function MMI.clean!(stack::Stack)
     # We only carry checks and don't try to correct the arguments here
     message = ""
-
+    # Checking target_scitype and input_scitype have not been changed from the original stack
+    params = typeof(stack).parameters
```
ablaom, line 153: Okay, I'm not sure accessing parameters this way is necessarily stable. The sanctioned way to do this is to change your signature:

```julia
function MMI.clean!(stack::Stack{modelnames, inp_scitype, tg_scitype}) where {modelnames,inp_scitype,tg_scitype})
```

src/composition/models/stacking.jl
```diff
@@ -133,17 +133,35 @@ function Stack(;metalearner=nothing, resampling=CV(), named_models...)
...
+    inp_scitype, tg_scitype = input_target_scitypes(getfield(stack, :models), stack.metalearner)
+    inp_scitype == params[end-1] || 
+            throw(DomainError(inp_scitype, "The newly inferred input_scitype of the stack doesn't 
+            match its original one. You have probably changed one of the model or the metalearner
```
ablaom, line 157: model -> base models

src/composition/models/stacking.jl
```diff
@@ -133,17 +133,35 @@ function Stack(;metalearner=nothing, resampling=CV(), named_models...)
...
     # We only carry checks and don't try to correct the arguments here
     message = ""
-
+    # Checking target_scitype and input_scitype have not been changed from the original stack
+    params = typeof(stack).parameters
+    inp_scitype, tg_scitype = input_target_scitypes(getfield(stack, :models), stack.metalearner)
+    inp_scitype == params[end-1] || 
+            throw(DomainError(inp_scitype, "The newly inferred input_scitype of the stack doesn't 
+            match its original one. You have probably changed one of the model or the metalearner 
+            to a non compatible type."))
```
ablaom, line 158: non compatible -> incompatible

and below as well

src/composition/models/stacking.jl
olivierlabayle: Ok thanks, I had also the feeling that the syntax wasn't particularly "Julian"

olivierlabayle:
Thanks for the time invested in this review and for helping me to adapt the code to the requirements of the package. Let me know if you need anything else.  I'm happy for you to merge.

ablaom (approved):
So good to finally have this stacking, which we promised for some time. 🙏🏾

Issue #575, opened 2021-06-09
New patch release of LossFunctions is triggering build fail


ablaom:
Here's the fail: https://github.com/alan-turing-institute/MLJBase.jl/pull/574/checks?check_run_id=2787740796

The problem, I'm pretty sure, is this new line in LossFunctions.jl

```
(loss::SupervisedLoss)(target::AbstractArray, output::AbstractArray) = value(loss, target, output)
```

In MLJBase, we also overload this, which I suppose is type piracy.

@juliohm

ablaom:
https://github.com/alan-turing-institute/MLJBase.jl/blob/8f30a1b67b281237fac0819aabf2537ed5fe1915/src/measures/loss_functions_interface.jl#L118

juliohm:
Yes, this is type piracy, unfortunately. I added the convenience method in LossFunctions.jl because I realized that it was missing. We had the functionality documented in the docs but I forgot to re-add the method after a major refactoring of the source code. Without the method users are forced to import `LossFunctions.value` to evaluate a loss object.

Perhaps you can just reuse the method in MLJBase.jl now or create a wrapper type to make minor adjustments if necessary. I remember there were differences in the order of yhat and y arguments? We should probably change the order in LossFunctions.jl in the future to be consistent with MLJ and Flux if they are following the same convention already.

ablaom:
@juliohm Thanks for chipping in.

Yes, you are right, we use the order `yhat, y` which is the opposite. However, the signature is also different, as the MLJ versions of the probabilistic loss functions expect distribution objects rather than just numbers. 

I was going to just to define local versions of all the types and just use LossFunctions methods. I think a wrapper type is probably not worth the extra trouble, in terms making parameter specification and access the same as it is now.

ablaom:
Mmm. Maybe a wrapper is better after all. Otherwise, we have to construct a LossFunctions.jl loss instance for every evaluation.

ablaom:
There is a residual fail after above PR due to [LossFunctions bug](https://github.com/JuliaML/LossFunctions.jl/issues/142). When that is resolved, this issue can be closed.

PR #577, opened 2021-06-10
Remove LossFunctions type piracy 

ablaom:
#575


ablaom:
The fail on julia 1.0 is due to this bug: https://github.com/JuliaML/LossFunctions.jl/issues/142 . MLJBase will not load under 1.0 until this is fixed.

Issue #576, opened 2021-06-10
`yhat` and `y` transposed in measures integrated from LossFunctions.jl


ablaom:
In MLJBase:

https://github.com/alan-turing-institute/MLJBase.jl/blob/8f30a1b67b281237fac0819aabf2537ed5fe1915/src/measures/loss_functions_interface.jl#L32

However, the latest LossFunctions docs:

<img width="477" alt="Screen Shot 2021-06-10 at 12 21 36 PM" src="https://user-images.githubusercontent.com/30517088/121445968-a56d5900-c9e6-11eb-831e-1b55158fbf63.png">

I'm presuming this was not picked up in tests because tests were restricted to symmetric cases?

ablaom:
On further examination, it appears all the loss functions in LossFunctions.jl are symmetric in their arguments `yhat` and `y`. 

Still, we should fix this for future proofing.

PR #578, opened 2021-06-10
Fix transposed `y` and `yhat` in measures from LossFunctions

ablaom:
Addresses #576


ablaom:
The fail on julia 1.0 is due to this bug: JuliaML/LossFunctions.jl#142 .

Even the current MLJBase release will not load under 1.0 until this is fixed, unless LossFunctions is pinned.

So I don't see any advantage in delaying the current fix.

PR #583, opened 2021-06-18
Fix some formatting in stack doc-string and simplify a bit

ablaom:
No changes to code here.

@olivierlabayle


ablaom:
Fail is unrelated and tracked.

PR #587, opened 2021-06-27
Update doc deployment url

ablaom:


ablaom:
Cause of julia 1.0 build is unrelated and tracked. 

Note: CI for this PR skips doc generation, so this  still needs testing.

PR #591, opened 2021-07-03
improve performance of `MMI.nrows`

OkonSamuel:
Improve performance of `MMI.nrow` for high-dimensional tables.


ablaom (approved):
Excellent! Many thanks!

PR #593, opened 2021-07-05
add docstring fix

olivierlabayle:
While reading the MLJ doc I realized I had missed to change the docstring relative to the resampling argument.


ablaom (approved):
Good catch! Thanks

PR #595, opened 2021-07-22
Nested cross-validation

rikhuijzer:
This PR implements nested cross-validation for different models.

For anyone also searching for nested cross-validation in MLJ, I just learned from @ablaom that it is possible to wrap a model in a `TunedModel` and `evaluate` that nested model with cross-validation.

EDIT: The reason that I think this is a good idea is because I want to compare multiple models in a paper. To people who say that you shouldn’t use nested cross-validation for model selection, I would argue that it’s better than picking the model by hand. But, now that I think of it, doing cross-validation for each model and comparing those scores is very similar except that it gives a more biased estimate for the out of sample accuracy.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -49,6 +49,7 @@ julia = "1"
 DecisionTree = "7806a523-6efd-50cb-b5f6-3fa6f1930dbb"
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
+MLJModels = "d491faf4-2d78-11e9-2867-c94bc002c0b7"
 MultivariateStats = "6f286f6a-111f-5878-ab1e-185364afe411"
 NearestNeighbors = "b8a86587-4115-5ab1-83bc-aa920d37bbce"
 StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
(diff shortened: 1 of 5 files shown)


ablaom:
Okay, I’ve had a think about this and believe a more MLJ idiomatic solution *almost* exists already. One of the unpublicised tuning strategies, called `Explicit`, enables you to simply choose, out of the models listed in the `range` the one with the best out-of-sample loss, according to specified `resampling` strategy, which could be `CV` or whatever. Then to get the functionality you want, you would do something like this:

```julia
models = [DecisionTreeClassifier(), KNNClassifier()]
blend = TunedModel(strategy=Explicit(), resamling=CV(nfolds=3), range=models, measure=log_loss, selection_heuristic=...)
evaluate(blend, X, y, resampling=CV(nfolds=6), measure=log_loss)
```

I say "almost" because at present the list of models must be models all of the same concrete type, which is useless (except for the testing purposes it was introduced for). 

I've opened an issue to fix this (with some suggested syntactic improvements) [here](https://github.com/JuliaAI/MLJTuning.jl/issues/141).

One of the reasons this might be a better solution is that whatever future `selection_heuristic`s we add for model selection one the hyper-parameter side, would automatically be applicable on the "model" selection side. From an abstract point of view, one could argue that there is no difference between "model selection" and "hyper-parameter" selection, or at least that one is just a special case of the other, no?)

I think this solution also has fewer conceptual overheads - it simply composes two operations that the user already understands: optimization as model wrapper - and estimating the performance of a single model. 

@rikhuijzer Your thoughts?

rikhuijzer:
> Your thoughts?

Sounds like a much more MLJ idiomatic solution. I was already thinking with this PR that I must be doing something wrong, because I couldn't find a way to blend the code in with the rest of the codebase :+1: 

> From an abstract point of view, one could argue that there is no difference between "model selection" and "hyper-parameter" selection, or at least that one is just a special case of the other, no?

Agreed!

> I think this solution also has fewer conceptual overheads - it simply composes two operations that the user already understands: optimization as model wrapper - and estimating the performance of a single model.

As long as it's documented, I agree. I've been looking all over the MLJ place for nested cross-validation and was, therefore, very surprised when you said on Slack that it was possible for hyperparameters :slightly_smiling_face:

Issue #598, opened 2021-07-27
Have evaluate! automatically run the right kinds of predictions for each metric


ablaom:
See https://github.com/alan-turing-institute/MLJ.jl/issues/795#issuecomment-848374803

PR #599, opened 2021-07-30
Enable `evaluate!` to infer automatically different operations for different measures

ablaom:
Addresses #598 according to [this suggestion](https://github.com/alan-turing-institute/MLJ.jl/issues/795#issuecomment-848374803). 

(This enhancement should mitigate greatly one of the most common gotchas for beginner users.)

Also improves the `show` for measures, to make printed output of evaluate less ambiguous.

My apologies for the messy PR - the 4th commit, which changes a lot of files, is only to suppress noisy logging during tests.  The key commits to review are the 3rd and 5th commits. 

From the updated doc-string from `evaluate!`:

---

The type of operation (`predict`, `predict_mode`, etc) to be
associated with `measure` is automatically inferred from measure
traits where possible. For example, `predict_mode` will be used for a
`Multiclass` target, if `model` is probabilistic but `measure` is
deterministic. Abmiguities, such as `predict_mean`/`predict_median`
are logged to `Info`. To suppress the logging, lower `verbosity` or
explicitly specify `operation=...`. If `measure` is a vector, then any
`operation` specified must be a single operation, which will be
associated with all measures, or a vector of the same length.

### DIFF
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -43,7 +43,8 @@ import Dates
 # Distributed computing
 using Distributed
 using ComputationalResources
-using ComputationalResources: CPUProcesses
+import ComputationalResources: CPU1, CPUProcesses, CPUThreads
+
 using ProgressMeter
 import .Threads
 
@@ -102,6 +103,9 @@ export fit, update, update_data, transform, inverse_transform,
 # data operations
 export matrix, int, classes, decoder, table,
     nrows, selectrows, selectcols, select
+
+# re-export from ComputationalResources.jl:
+export CPU1, CPUProcesses, CPUThreads
 
 # re-exports from ScientificTypes
 export Unknown, Known, Finite, Infinite,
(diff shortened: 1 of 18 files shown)


ablaom:
@CameronBieganek Any change you could have a look over this?

CameronBieganek:
Sure, I can probably take a look at it this weekend.

ablaom:
One possible alternative I've just thought of: Instead of generating log in ambiguous case (`predict_mean` or `predict_median`) we simply report the operations used as part of the return object, which would then appear in its tabular representation. What do you think?

CameronBieganek:
> One possible alternative I've just thought of: Instead of generating log in ambiguous case (`predict_mean` or `predict_median`) we simply report the operations used as part of the return object, which would then appear in its tabular representation. What do you think?

I think that sounds like a good idea.

Sorry I didn't get to this PR this weekend, but I'll try to take a look tomorrow.

ablaom:
Okay, I've implemented the alternative at #600.

PR #628, opened 2021-09-03
Outlier Detection Integration

davnn:
Implementation of the proposed outlier detection draft with "Abstract" removed from the types for consistency.


ablaom:
@davnn I'm now working on resolving the conflict, upating the MLJModelnterface compat in preparation to merge.

Issue #630, opened 2021-09-05
Handling of NaN micro/macro aggregation


ablaom:
I think it might be worth reviewing the current handling of `NaN` in multiclass micro/macro aggregation and in binary `FScore`. 
At present a method called `_mean` substitutes zero for any `NaN` in the one-against-many fscores, before aggregation. Perhaps the `NaN`' should just be skipped?? Also, it appears the `NaN` will arise whenever the precision or recall is zero. However, even in the binary case, FScore could be well-defined even if both precision and recall are zero:

precision = TP/(TP +FP)
recall = TP/(TP + FN)

score is usually defined as harmonic mean of these (and so undefined if either vanishes); however, if you simplify the harmonic mean expression (elminate a common factor of TP from final denominator and numerator) you get:

score = 2TP/(2TP + FP + FN)

which is only undefined if every sample is TN.  

cc @ven-k 

**added after discussion below**

To do: Replace harmonic mean of precision and recall with explicit reduced formula involving basic counts, as indicated above.

- [x] MulticlassFScore
- [x] Vanilla binary FScore
- [ ] Change substitution with 0 with skip; see [comment below](https://github.com/JuliaAI/MLJBase.jl/issues/630#issuecomment-922236280)

ven-k:
- `MulticlassFScore` returns a NaN only when both `Precision` and `Recall` are zero. 
- But it's a pronounced possibility for Micro_avg as both are the same. I have modified the code to return `recall` so that `NaN` never occurs. (harmonic mean of like quantities is the quantity itself) 
- Even with expanded formula, if a class has all values as true negatives, that entry will be a NaN when `no_avg` is used. So `macro_avg` will need the current `_mean` (even with the current changes). However, using precision and recall separately for FScore will place multiple `_mtp` calls.
- #637 takes care of multiclass fscore

---
- The `_mean` function seems to be doing the right thing- replace NaN with 0 so that `n = no. of entries`.

ablaom:
@ven-k Thanks for the prompt review, clarifications and fixes!

I'm not sure I agree we should count the classes that contribute nonsensical scores (that is, we should skip instead of setting to zero) but I also concede this is a weird corner case where there may not be a clear "right" answer. So happy to support your take on it. Again many thanks.

Leaving open until binary case sorted - see updated opening comment.

ven-k:
[...]

PR #637, opened 2021-09-12
Use expanded formula of MulticlassFScore

ven-k:
This addresses #630 for multiclass FScores.


ablaom (approved):
@ven-k Looks good to me. Many thanks

PR #639, opened 2021-09-17
Pipelines without macros

ablaom:
This PR is a step towards replacement of `@pipeline` with non-macro alternatives (https://github.com/alan-turing-institute/MLJ.jl/issues/594). It implements new parametric types for pipelines and a new constructor `Pipeline`, following the suggestions of @CameronBieganek in the cited issue.

The new pipelines offer all the features of @pipeline with the exception of target transformations, which are to be [provided with a separate wrapper](https://github.com/alan-turing-institute/MLJ.jl/issues/594#issuecomment-751863495). Additional features are:

- `inverse_transform` is implemented, where that makes sense (there was an issue raised, which I cannot find)
- implement `predict` for unsupervised pipelines ending in a model that implements `predict` (clustering) (https://github.com/JuliaAI/MLJClusteringInterface.jl/issues/10)
- Alternative constructor using `|>`. When two pipelines are concatenated, the result is flat (does not introduce an extra layer of hyper-parameter nesting).

The new doc-string is copied below under "details".

### Conflict with existing feature

The overloading of `|>` conflicts with the "arrow syntax" described [here](https://juliaai.github.io/DataScienceTutorials.jl/getting-started/learning-networks/#arrow_syntax) but never actually documented in the manual. It's purpose is to reduce code needed to build learning networks. I have not seen it used anywhere outside of the cited tutorial and would suggest linear pipelines, as a more common use-case, would be a better use. A non-breaking alternative would be to overload the generic associative operator `*`. 

What do others think?

<details>

    Pipeline(component1, component2, ... , componentk; options...)
    Pipeline(name1=component1, name2=component2, ..., componentk; options...)
    component1 |> component2 |> ... |> componentk

Create an instance of composite model type which sequentially composes
the specified components in order. This means `component1` receives
inputs, whose output is passed to `component2`, and so forth. A
"component" is either a `Model` instance, a model type (converted
immediately to its default instance) or any callable object.

At most one of the components may be a supervised model, but this
model can appear in any position.

The `@pipeline` macro accepts key-word `options` discussed further
below.

Ordinary functions (and other callables) may be inserted in the
pipeline as shown in the following example:

    Pipeline(X->coerce(X, :age=>Continuous), OneHotEncoder, ConstantClassifier)

### Syntactic sugar

The `|>` operator is overloaded to construct pipelines out of models,
functions, and existing pipelines:

```julia
LinearRegressor = @load LinearRegressor pkg=MLJLinearModels add=true
PCA = @load PCA pkg=MultivariateStats add=true

pipe1 = MLJBase.table |> ContinuousEncoder |> Standardizer
pipe2 = PCA |> LinearRegressor
pipe1 |> pipe2
```

### Special operations

If all the `components` are invertible unsupervised models
(transformers), then `inverse_transform` is implemented for the
pipeline. If there are no supervised models, then `predict` is
nevertheless implemented, assuming the last model (such as `KMeans`
clustering) also implements it. Similarly, calling `transform` on a
supervised pipeline calls `transform` on the supervised component.

### Optional key-word arguments

- `prediction_type`  -
  prediction type of the pipeline; possible values: `:deterministic`,
  `:probabilistic`, `:interval` (default=`:deterministic` if not inferable)

- `operation` - operation applied to the supervised component model,
  when present; possible values: `predict`, `predict_mean`,
  `predict_median`, `predict_mode` (default=`predict`)

- `cache` - whether the internal machines created for component models
  should cache model-specific representations of data. See [`machine`](@ref).

!!! warning "Set `cache=false` to guarantee data anonymization"

    This precaution applies to composite models, and only to those
    implemented using learning networks.

To build more complicated non-branching pipelines, refer to the MLJ
manual sections on composing models.

</details>

@CameronBieganek Any depth of review here would be greatly appreciated.

### DIFF
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -147,12 +147,15 @@ export machine, Machine, fit!, report, fit_only!
 export make_blobs, make_moons, make_circles, make_regression
 
 # composition:
-export machines, sources, @from_network, @pipeline, Stack,
+export machines, sources, @from_network, @pipeline, Pipeline, Stack,
     glb, @tuple, node, @node, sources, origins, return!,
     nrows_at_source, machine,
     rebind!, nodes, freeze!, thaw!, Node, AbstractNode,
     DeterministicSurrogate, ProbabilisticSurrogate, UnsupervisedSurrogate,
-    DeterministicComposite, ProbabilisticComposite, UnsupervisedComposite
+    DeterministicComposite, ProbabilisticComposite, UnsupervisedComposite,
+    DeterministicPipeline, ProbabilisticPipeline, IntervalPipeline,
+    UnsupervisedPipeline, StaticPipeline
+
 
 # aliases to the above,  kept for backwards compatibility:
 export  DeterministicNetwork, ProbabilisticNetwork, UnsupervisedNetwork
@@ -366,6 +369,7 @@ include("composition/models/methods.jl")
 include("composition/models/from_network.jl")
 include("composition/models/inspection.jl")
 include("composition/models/pipelines.jl")
+include("composition/models/pipelines2.jl")
 include("composition/models/_wrapped_function.jl")
 
 include("operations.jl")
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -361,9 +361,9 @@ include("composition/abstract_types.jl")
 include("composition/learning_networks/nodes.jl")
 include("composition/learning_networks/inspection.jl")
 include("composition/learning_networks/machines.jl")
-@static if VERSION ≥ v"1.3.0-"
-    include("composition/learning_networks/arrows.jl")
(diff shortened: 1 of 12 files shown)


tlienart:
I'm fine with the arrow syntax being dropped in favour of something that people consider more standard and readable :)

CameronBieganek:
Awesome!!! I'm busy this weekend, but I'll try to take a look at this as soon as possible.

CameronBieganek:
@ablaom Not sure how urgent this is. I'm pretty swamped right now, so I won't be able to look at this this week or this weekend. But maybe next week...

ablaom:
@CameronBieganek Thanks for the update!

 I think it's important to have a review and you are still my first choice. If you think it's likely more than a month away, let me know.

PR #646, opened 2021-09-26
fix bug in `_confmat`

OkonSamuel:
fixed bug in `_confmat` method causing errors in julia 1.6.3. These errors were reported by a user on slack.


ablaom:
@OkonSamuel Thanks for diagnosing and fixing this issue!

PR #650, opened 2021-09-30
Fix binary `FScore` corner case 

ablaom:
Addresses part of #630


ablaom:
@ven-k Could you review please (easy)

ven-k (approved):

PR #657, opened 2021-10-04
Fix input checks for unsupervised annotators

davnn:
Unsupervised annotators can be called in both a supervised and unsupervised fashion.


ablaom (approved):
Looks good to me, thanks.

PR #666, opened 2021-10-29
Allow nominal ranges consisting of different functions.

davnn:
Outlier detection models might tune different score normalization functions, which is not possible currently.

### DIFF
--- a/src/hyperparam/one_dimensional_ranges.jl
+++ b/src/hyperparam/one_dimensional_ranges.jl
@@ -52,6 +52,15 @@ function Base.show(stream::IO,
     return nothing
 end
 
+WARN_INFERRING_TYPE =
+    "Inferring the hyper-parameter type from the given "*
+    "model instance as the parameter is of type `Any`."
+
+WARN_AMBIGUOUS_UNION =
+    "The inferred hyper-parameter type is ambiguous, because "*
+    "the type union contains multiple real subtypes, using "*
+    "the first subtype to create the parameter range."
+
 """
     r = range(model, :hyper; values=nothing)
 
@@ -105,10 +114,18 @@ function Base.range(model::Union{Model, Type}, field::Union{Symbol,Expr};
 
     if model isa Model
         T = recursive_getpropertytype(model, field)
+        if T === Any
+            @warn WARN_INFERRING_TYPE
+            T = typeof(recursive_getproperty(model, field))
+        end
     else
         T = model
     end
-    if T <: Real && values === nothing
+    union_types = Base.uniontypes(T)
+    possible_types = filter(t -> t <: Real, union_types)
+    n_real = length(possible_types)
+    if n_real > 0 && values === nothing
+        n_real > 1 && @warn WARN_AMBIGUOUS_UNION 
         return numeric_range(T, D, field, lower, upper, origin, unit, scale)
     else
         return nominal_range(T, field, values)
--- a/src/hyperparam/one_dimensional_ranges.jl
+++ b/src/hyperparam/one_dimensional_ranges.jl
@@ -56,10 +56,9 @@ WARN_INFERRING_TYPE =
     "Inferring the hyper-parameter type from the given "*
     "model instance as the parameter is of type `Any`."
 
-WARN_AMBIGUOUS_UNION =
+ERROR_AMBIGUOUS_UNION =
     "The inferred hyper-parameter type is ambiguous, because "*
-    "the type union contains multiple real subtypes, using "*
-    "the first subtype to create the parameter range."
+    "the union type contains multiple real subtypes."
 
 """
     r = range(model, :hyper; values=nothing)
@@ -121,12 +120,11 @@ function Base.range(model::Union{Model, Type}, field::Union{Symbol,Expr};
     else
         T = model
     end
-    union_types = Base.uniontypes(T)
-    possible_types = filter(t -> t <: Real, union_types)
-    n_real = length(possible_types)
-    if n_real > 0 && values === nothing
-        n_real > 1 && @warn WARN_AMBIGUOUS_UNION 
-        return numeric_range(T, D, field, lower, upper, origin, unit, scale)
+    possible_types = filter(t -> t <: Real, Base.uniontypes(T))
+    n_possible_types = length(possible_types)
+    if n_possible_types > 0 && values === nothing
+        n_possible_types > 1 && error(ERROR_AMBIGUOUS_UNION) 
+        return numeric_range(first(possible_types), D, field, lower, upper, origin, unit, scale)
     else
         return nominal_range(T, field, values)
     end
--- a/src/hyperparam/one_dimensional_ranges.jl
+++ b/src/hyperparam/one_dimensional_ranges.jl
@@ -58,7 +58,9 @@ WARN_INFERRING_TYPE =
 
 ERROR_AMBIGUOUS_UNION =
     "The inferred hyper-parameter type is ambiguous, because "*
-    "the union type contains multiple real subtypes."
+    "the union type contains multiple real subtypes. You can "*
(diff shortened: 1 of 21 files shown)


ablaom:
This *should* be possible already, as in:

```julia
julia> range(Function, :f, values=[cos, sin, tan])
```

An error is thrown, but only because of a [bug](https://github.com/JuliaAI/MLJBase.jl/issues/667) in the  display. If I suppress display, it works:

```julia
julia> r = range(Any, :f, values=[cos, sin, tan]);

julia> iterator(r)
3-element Vector{Function}:
 cos (generic function with 13 methods)
 sin (generic function with 13 methods)
 tan (generic function with 12 methods)
```

However, I expect your complaint is with the case where you use a model in the first argument to automatically infer the type, as in 

```julia
mutable struct MyModel <: Deterministic
   f::Function
end

julia> range(MyModel(sin), :f, values=[sin, cos, tan])
ERROR: ArgumentError: `Function[sin, cos, tan]` must be an instance of type `AbstractVector{<:typeof(sin)}`.
Stacktrace:
 [1] nominal_range(T::Type, field::Symbol, values::Vector{Function})
   @ MLJBase ~/MLJ/MLJBase/src/hyperparam/one_dimensional_ranges.jl:176
 [2] range(model::MyModel, field::Symbol; values::Vector{Function}, lower::Nothing, upper::Nothing, origin::Nothing, unit::Nothing, scale::Nothing)
   @ MLJBase ~/MLJ/MLJBase/src/hyperparam/one_dimensional_ranges.jl:115
 [3] top-level scope
   @ REPL[25]:1
```

We have run into related issues frequently, [for example](https://github.com/JuliaAI/MLJBase.jl/issues/372) when a hyperparameter has type `Union{Nothing, Int}` and the instance used for inferring the type has `nothing` but the values are `Int`. A more generic solution to that suggested in this PR is to get the hyperparameter type from the model *type* instead of model *instance* . I feel there must have been a reason we dismissed this, but I can't think of one now!

@davnn Thoughts?

davnn:
> A more generic solution to that suggested in this PR is to get the hyperparameter type from the model typeinstead of model instance .

I was actually surprised that it doesnt work that was, but yes that should be the solution if it‘s doable imho.

Freundliche Grüße / Best regards
David Muhr

Sent from mobile.
________________________________

davnn:
> A more generic solution to that suggested in this PR is to get the hyperparameter type from the model typeinstead of model instance .

I was actually surprised that it doesnt work that was, but yes that should be the solution if it‘s doable imho.

davnn:
@ablaom awaiting your comments. We should probably add some tests ;)

ablaom:
@davnn Thanks for that. This looks good. 

There could be some unexpected behaviour for the occasional field that is not typed in the struct definition (ie, is `Any`). Perhaps if the type is `Any` we should fall back to the type of the instance and issue a warning like "Inferred type of range values is `Any`. Using $T. To override specify type explicitly, as in `range(Int64, :my_parameter, ...)`."?

Also, if we have `Union{Nothing, Integer}`, say, then we really want to pass `Integer` on to the more specialised methods or get unexpected behaviour (in numeric ranges, integers imply rounding, other values do not). 

What do you think?

By the way, I have in mind to factor this code out of MLJBase, in which case the first argument will always be a type (to avoid dependency on MLJ) and this "type inference" stuff will be MLJ-specific overloading of the externally defined method. 

And, yes some tests would be good.

Your revisions would trigger a breaking release, which could be a week or two away yet.

If you want to kick the can on this, I'd be happy to merge your original fix immediately, but I'm also happy to support the new effort. In the latter case, please rebase onto the branch "[for-0-point-19-release branch](https://github.com/JuliaAI/MLJBase.jl/tree/for-0-point-19-release)".

davnn:
> There could be some unexpected behaviour for the occasional field that is not typed in the struct definition (ie, is Any). Perhaps if the type is Any we should fall back to the type of the instance and issue a warning like "Inferred type of range values is Any. Using $T. To override specify type explicitly, as in range(Int64, :my_parameter, ...)."?

Sounds good, I've implemented the changes.

> Also, if we have Union{Nothing, Integer}, say, then we really want to pass Integer on to the more specialised methods or get unexpected behaviour (in numeric ranges, integers imply rounding, other values do not).

My proposal didn't really address this case, but I have added it now. The problem is what should we do if we get an ambiguous union like `Union{Float32, Integer, Nothing}`. We could always take the first `<: Real` and throw a warning, that's how I've implemented it for now.

> If you want to kick the can on this, I'd be happy to merge your original fix immediately, but I'm also happy to support the new effort. In the latter case, please rebase onto the branch "for-0-point-19-release branch".

Rebased.

ablaom:
Double checking that branch: https://github.com/JuliaAI/MLJBase.jl/pull/669

ablaom:
Okay that branch is not passing tests. I'll investigate and get back to you.

src/hyperparam/one_dimensional_ranges.jl
```diff
@@ -105,10 +114,18 @@ function Base.range(model::Union{Model, Type}, field::Union{Symbol,Expr};
...
         T = recursive_getpropertytype(model, field)
+        if T === Any
+            @warn WARN_INFERRING_TYPE
+            T = typeof(recursive_getproperty(model, field))
+        end
     else
         T = model
     end
-    if T <: Real && values === nothing
+    union_types = Base.uniontypes(T)
```
ablaom, line 124: Mmm. I suspect `uniontypes`  is not part of the public API, right? (Although I'm very glad to learn about this method, as it's better than `T.a` and `T.b` !) Additionally, I don't think "first" is well-defined, because `Union{Int,Char} == Union{Char,Int}`. 

How about we revert to the typeof instance in the case of ambiguous unions as well? I feel this is a corner of a corner case anyway.

ablaom:
That branch is now fixed. It seems you need to make new commits to re-trigger CI.

src/hyperparam/one_dimensional_ranges.jl
ablaom: Okay, this is little more subtle than I thought at first. What about using `Base.typesplit` insead? This is used in the implementation of `MIssings.nonmissingtype`:

```julia
julia> Base.typesplit(Union{Nothing,Float64,Int}, Nothing)
Union{Float64, Int64}
```

davnn: > Mmm. I suspect `uniontypes` is not part of the public API, right? (Although I'm very glad to learn about this method, as it's better than `T.a` and `T.b` !) Additionally, I don't think "first" is well-defined, because `Union{Int,Char} == Union{Char,Int}`.

Nope it's not public, we could also copy the implementation (see below). It's first as in `first` of the `uniontypes` array, but we should remove that from the warning message and show the used type instead.

```julia
_uniontypes(x::Union, ts) = (_uniontypes(x.a,ts); _uniontypes(x.b,ts); ts)
_uniontypes(@nospecialize(x), ts) = (push!(ts, x); ts)
uniontypes(@nospecialize(x)) = _uniontypes(x, Any[])
```

Not sure how `Base.typesplit` would help us here? What range would you create with the resulting `Union{Float64, Int64}`? You'd still have to guess what type is meant or not? I mean instead of guessing we could just error when there is an ambiguity.

ablaom: > You'd still have to guess what type is meant or not?

You are right!

> I mean instead of guessing we could just error when there is an ambiguity.

Yes, let's throw error in case of ambiguity. With a suggestion to explicitly specify the type, as we already have in existing warnings.

davnn: > Yes, let's throw error in case of ambiguity. With a suggestion to explicitly specify the type, as we already have in existing warnings.

Done, see tests for examples.

PR #670, opened 2021-11-02
Implement conditional testing

davnn:
It should be possible to conditionally test some sub-modules.

Edit: This PR allows us to call `Pkg.test` with test arguments, e.g. `Pkg.test("MLJBase", test_args=["interface"])` would only run the interface tests.


ablaom (approved):
Didn't know about this feature. Many thanks!

PR #671, opened 2021-11-02
Enable stacking of outlier detection models

davnn:
support probabilistic and deterministic outlier detection models.

### DIFF
--- a/src/composition/models/stacking.jl
+++ b/src/composition/models/stacking.jl
@@ -245,16 +245,28 @@ trainrows(X::AbstractNode, folds::AbstractNode, nfold) =
 testrows(X::AbstractNode, folds::AbstractNode, nfold) =
     node((XX, ff) -> selectrows(XX, ff[nfold][2]), X, folds)
 
+throw_if_missing(y) = any(ismissing.(y)) && error("Stack models cannot use labels with missing values")
 
-pre_judge_transform(ŷ::Node, ::Type{<:ProbabilisticTypes}, ::Type{<:AbstractArray{<:Union{Missing,Finite}}}) =
+pre_judge_transform(ŷ::Node,
+                    ::Type{<:ProbabilisticTypes},
+                    ::Type{<:AbstractArray{<:Union{Missing, Finite}}}) = begin
+    throw_if_missing(ŷ)
     node(ŷ -> pdf(ŷ, levels(first(ŷ))), ŷ)
+end
 
-pre_judge_transform(ŷ::Node, ::Type{<:ProbabilisticTypes}, ::Type{<:AbstractArray{<:Continuous}}) =
+pre_judge_transform(ŷ::Node,
+                     ::Type{<:ProbabilisticTypes},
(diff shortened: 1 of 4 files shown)


davnn:
Not sure how to best deal with models that can work with missing values in the labels (semi-supervised models). I would just dissallow missing values in the labels of stacking models.

ablaom:
If you are stacking outlier detection models (the base learners) then presumably the new composite is also a outlier detection model. But this is not what is implemented here. Rather the composite has either type `ProbabilisticComposite <: Probabilistic` or `DeterministicComposite <: Deterministic`. It seems to me we need the kind of generalization worked out for `BinaryThresholdPredictor` ([link](https://github.com/JuliaAI/MLJModels.jl/blob/dev/src/builtins/ThresholdPredictors.jl#L19)) with separate composite types for each detector type, no?

davnn:
> If you are stacking outlier detection models (the base learners) then presumably the new composite is also a outlier detection model.

I think it would be fine if the models are a `<: Probabilistic` or `<: Deterministic` as the metalearner has to be supervised anyways.

ablaom:
Okay, so let me see if I understand. The idea is not to stack a bunch of (base) outlier detection models to produce a new outlier detection model, but rather to include one or more outlier detection models in the stack of (otherwise supervised) learners to define a new supervised model that one hopes is more robust to outliers? Is that right?

That's sounds interesting. Has this been tried elsewhere? Have already tried this out in the current implementation?

Would you be willing to add a very basic integration test? There is a dummy detector model you can copy from [here](https://github.com/JuliaAI/MLJModels.jl/blob/c9e8c4a4f40ec0c83496bb5f3a631675d5d4d14a/test/builtins/ThresholdPredictors.jl#L116) if you need to.

davnn:
> Okay, so let me see if I understand. The idea is not to stack a bunch of (base) outlier detection models to produce a new outlier detection model, but rather to include one or more outlier detection models in the stack of (otherwise supervised) learners to define a new supervised model that one hopes is more robust to outliers? Is that right?

That's an interesting idea and should be explored ;). My thinking was different: even if the models only consists of detectors, the resulting model is a binary classifier (either probabilistic or deterministic) as the metalearner learns a decision boundary instead of a scoring function.

> Would you be willing to add a very basic integration test? There is a dummy detector model you can copy from here if you need to.

I will add some tests 👍

davnn:
Working with semi-supervised models (`target_scitype <: Union{Missing, T}`) is pretty tricky considering Julia's invariant type system. I think it makes things easier to update the `target_scitype` of detectors to `AbstractVector{<:OrderedFactor{2}}` and leave the special case of semi-supervised detectors for later. We might work something out (using traits?) to add `Missing` support if a model has a semi-supervised trait.

davnn:
Tests should work once https://github.com/JuliaAI/MLJModelInterface.jl/pull/123 is merged.

ablaom:
Closing for now as probably out-of-scope without a significant overhaul of  MLJModelInterface.

PR #678, opened 2021-11-17
Add wrapper to transform targets of (semi)supervised models

ablaom:
This PR adds a model wrapper `TransformedTargetModel` implementing [this suggestion](https://github.com/alan-turing-institute/MLJ.jl/issues/594#issuecomment-751863495) of  @CameronBieganek. Together with the `Pipeline`model already implemented (on the target branch) this wrapper renders the `@pipeline` macro redundant and whence resolves [this Pluto notebook issue](https://github.com/alan-turing-institute/MLJ.jl/issues/865).

The doc-string for the new model appears below.

This PR is not breaking but I am basing it on the 0.19 release branch as it was convenient to use some utilities introduced there for the new pipelines. 

To do:

- [ ] add deprecation warnings for `@pipeline`
- [ ] review test coverage

<details>

    TransformedTargetModel(model; target=nothing, inverse=nothing, cache=true)

Wrap the supervised or semi-supervised `model` in a transformation of
the target variable.

Here `target` is either:

- The `Unsupervised` model that is to transform the training target.
  By default (`inverse=nothing`) the parameters learned by this
  transformer are also used to inverse-transform the predictions of
  `model`, which means `target` must implement the `inverse_transform`
  method. If this is not the case, specify `inverse=identity` to
  suppress inversion.

*or*

- A callable object for transforming the target, such as `y ->
  log.(y)`. In this case 
  a callable `inverse`, such as `z -> exp.(z)`, should be specified.

Specify `cache=false` to prioritize memory over speed, or to guarantee data
anonymity.

Specify `inverse=identity` if `model` is a probabilistic predictor, as
inverse-transforming sample spaces is not supported. Alternatively,
replace `model` with a deterministic model, such as `Pipeline(model,
y -> mode.(y))`.

### Examples

A model that normalizes the target before applying ridge regression,
with predictions returned on the original scale:

```
@load RidgeRegressor pkg=MLJLinearModels
model = RidgeRegressor()
tmodel = TransformedTargetModel(model, target=Standardizer())
```

A model that instead applies a static `log` transformation to the data, again
returning predictions to the original scale:

```
tmodel2 = TransformedTargetModel(model, target=y->log.(y), inverse=z->exp.(y))
```

</details>

### DIFF
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -154,6 +154,8 @@ export machines, sources, @from_network, @pipeline, Stack,
     ProbabilisticPipeline, DeterministicPipeline, UnsupervisedPipeline,
     StaticPipeline, IntervalPipeline
 
+export TargetTransformed
+
 # aliases to the above,  kept for backwards compatibility:
 export  DeterministicNetwork, ProbabilisticNetwork, UnsupervisedNetwork
 
@@ -366,6 +368,7 @@ include("composition/models/from_network.jl")
 include("composition/models/inspection.jl")
 include("composition/models/pipelines.jl")
 include("composition/models/pipelines2.jl")
+include("composition/models/target_transformed_model.jl")
 include("composition/models/_wrapped_function.jl")
 
 include("operations.jl")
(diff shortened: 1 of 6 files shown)


ablaom:
@CameronBieganek Would be great if you could comment on the doc-string. If you also have time for a more detailed review over the next two weeks, let me know.

PR #679, opened 2021-11-17
Add R-square (coefficient of determination)

rikhuijzer:
This PR adds the R² metric.

R² is more informative for linear regressions than MSE and RMSE (Chicco et al, [2021](https://doi.org/10.7717/peerj-cs.623)) because the scale is dimensionless (a percentage).

I value -3 in the test is manually calculated by me. All looks good.

EDIT: It doesn't work yet in CV. I get an `_check(::RSquare, Vector{Int})`. I'll debug it tonight or so

### DIFF
--- a/src/measures/continuous.jl
+++ b/src/measures/continuous.jl
@@ -77,8 +77,7 @@ metadata_measure(RSquared;
     instances               = ["rsq", "rsquared"],
     target_scitype          = InfiniteArrMissing,
     prediction_type         = :deterministic,
-    orientation             = :loss,
-    aggregation             = RSquared()
+    orientation             = :loss
 )
 
 const RSQ = RSquared
--- a/src/measures/continuous.jl
+++ b/src/measures/continuous.jl
@@ -69,6 +69,8 @@ call(::RootMeanSquaredError,
 # -------------------------------------------------------------------------
 # R-squared (coefficient of determination)
 
+foo(x) = x
+
 struct RSquared <: Aggregated end
 
 metadata_measure(RSquared;
@@ -91,11 +93,21 @@ This is because R² is a percentage whereas MSE and RMSE have arbitrary ranges.
 Let ``\\overline{y}`` denote the mean of ``y``, then
 
 ``\\text{R^2} = 1 - \\frac{∑ (\\hat{y} - y)^2}{∑ \\overline{y} - y)^2}.``
-""",
-footer="")
+""")
 
 function call(::RSquared, ŷ::ArrMissing{<:Real}, y::ArrMissing{<:Real})
     num = (ŷ .- y).^2 |> skipinvalid |> sum
+    mean_y = mean(y)
+    denom = (mean_y .- y).^2 |> skipinvalid |> sum
+    return 1 - (num / denom)
+end
+
+function call(::RSquared,
+        ŷ::ArrMissing{<:Real},
+        y::ArrMissing{<:Real},
+        w::Arr{<:Real}
+    )
+    num = (ŷ .- y).^2 .* w |> skipinvalid |> sum
     mean_y = mean(y)
     denom = (mean_y .- y).^2 |> skipinvalid |> sum
     return 1 - (num / denom)
--- a/src/measures/continuous.jl
+++ b/src/measures/continuous.jl
@@ -69,8 +69,6 @@ call(::RootMeanSquaredError,
 # -------------------------------------------------------------------------
 # R-squared (coefficient of determination)
 
-foo(x) = x
-
 struct RSquared <: Aggregated end
 
 metadata_measure(RSquared;
@@ -108,7 +106,7 @@ function call(::RSquared,
     )
     num = (ŷ .- y).^2 .* w |> skipinvalid |> sum
     mean_y = mean(y)
-    denom = (mean_y .- y).^2 |> skipinvalid |> sum
+    denom = (mean_y .- y).^2 .* w |> skipinvalid |> sum
     return 1 - (num / denom)
 end
 
--- a/src/measures/continuous.jl
+++ b/src/measures/continuous.jl
@@ -75,8 +75,7 @@ metadata_measure(RSquared;
     instances               = ["rsq", "rsquared"],
     target_scitype          = InfiniteArrMissing,
     prediction_type         = :deterministic,
-    orientation             = :loss
-)
+    orientation             = :score)
 
 const RSQ = RSquared
 @create_aliases RSquared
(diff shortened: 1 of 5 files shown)


rikhuijzer:
I don't get it (yet):
```
~/git/MLJBase.jl> julia --project

julia> using MLJBase

julia> using MLJModelInterface

julia> MMI = MLJModelInterface;

julia> include("test/_models/Constant.jl")

julia> m = ConstantRegressor();

julia> X, y = make_regression();

julia> e = evaluate(m, X, y; resampling=CV(), measures=[rms]);
Evaluating over 6 folds: 100%[=========================] Time: 0:00:00

julia> e = evaluate(m, X, y; resampling=CV(), measures=[rsq]);
Evaluating over 6 folds: 100%[=========================] Time: 0:00:00
ERROR: MethodError: no method matching _check(::RSquared, ::Vector{Float64})
Closest candidates are:
  _check(::MLJBase.Measure, ::Any, ::Any, ::AbstractArray) at /home/rik/git/MLJBase.jl/src/measures/measures.jl:61
  _check(::MLJBase.Measure, ::Any, ::Any) at /home/rik/git/MLJBase.jl/src/measures/measures.jl:53
  _check(::MLJBase.Measure, ::Any, ::Any, ::Any) at /home/rik/git/MLJBase.jl/src/measures/measures.jl:57
  ...
Stacktrace:
  [1] (::RSquared)(args::Vector{Float64})
    @ MLJBase ~/git/MLJBase.jl/src/measures/measures.jl:116
  [2] aggregate(v::Vector{Float64}, measure::RSquared)
    @ MLJBase ~/git/MLJBase.jl/src/measures/measures.jl:164
  [3] (::MLJBase.var"#298#304"{Vector{RSquared}, Vector{Vector{Float64}}})(k::Int64)
    @ MLJBase ~/git/MLJBase.jl/src/resampling.jl:1157
  [4] iterate
    @ ./generator.jl:47 [inlined]
```

rikhuijzer:
I've read `measures/README.md` and checked the code and tests, but still have no idea whatsoever what an "aggregation" is. So, be careful when reviewing this PR in whether the aggregation is correctly specified (by being left empty).

test/resampling.jl
```diff
@@ -359,7 +359,7 @@ end
     for cache in [true, false]
         model = Models.DeterministicConstantRegressor()
         mach = machine(model, X, y, cache=cache)
-        result = evaluate!(mach, resampling=cv, measure=[rms, rmslp1],
+        result = evaluate!(mach, resampling=cv, measure=[rms, rsq, rmslp1],
```
rikhuijzer, line 362: This is a smoke test which will break when aggregation is specific incorrectly.

src/measures/continuous.jl
```diff
@@ -66,6 +66,50 @@ call(::RootMeanSquaredError,
...
+end
+
+function call(::RSquared,
+        ŷ::ArrMissing{<:Real},
+        y::ArrMissing{<:Real},
+        w::Arr{<:Real}
+    )
+    num = (ŷ .- y).^2 .* w |> skipinvalid |> sum
+    mean_y = mean(y)
+    denom = (mean_y .- y).^2 .* w |> skipinvalid |> sum
```
rikhuijzer, line 109: I just copied this `w` from similar `call` functions but have not tested it or found a formula somewhere which stated that this is correct.

OkonSamuel:
> I've read `measures/README.md` and checked the code and tests, but still have no idea whatsoever what an "aggregation" is. So, be careful when reviewing this PR in whether the aggregation is correctly specified (by being left empty).

Thanks @rikhuijzer for this PR.
**"aggregation"** is a concept used in resampling in two ways
1. To aggregate the results gotten from calling `UnAggregated` performance measures on observations. (see section on `per_fold` in documentation for `evaluate!` function.)  See [here](https://github.com/JuliaAI/MLJBase.jl/blob/dev/src/resampling.jl#L1144-L1152) also.
2. To aggregate the results from several measures to have one common numerical value. This is useful when calling `evaluate`/`evaluate!` functions with a vector of measures. See [here](https://github.com/JuliaAI/MLJBase.jl/blob/dev/src/resampling.jl#L1155-L1158)

Now the reason why your previous commit errored was because, you defined a wrong aggregation in your metadata function 
```julia
aggregation             = RSquared()
```
Although `RSquared()` is a valid julia object (since you created it), it isn't a recognized `StatisticalTraits.AggregationMode` object. See [here](https://github.com/JuliaAI/MLJBase.jl/blob/dev/src/measures/measures.jl#L154-L164) for definitions of existing aggregations and how to define new ones.

src/measures/continuous.jl
```diff
@@ -66,6 +66,50 @@ call(::RootMeanSquaredError,
...
+# -------------------------------------------------------------------------
+# R-squared (coefficient of determination)
+
+struct RSquared <: Aggregated end
+
+metadata_measure(RSquared;
+    instances               = ["rsq", "rsquared"],
+    target_scitype          = InfiniteArrMissing,
+    prediction_type         = :deterministic,
+    orientation             = :loss
```
ablaom, line 78: I believe you want `orientation = :score` here, as bigger is better, right?

src/measures/continuous.jl
```diff
@@ -66,6 +66,50 @@ call(::RootMeanSquaredError,
...
+)
+
+const RSQ = RSquared
+@create_aliases RSquared
+
+@create_docs(RSquared,
+body=
+"""
+The R² (also known as R-squared or coefficient of determination) is very suitable for interpreting regression analysis (Chicco et al., [2021](https://doi.org/10.7717/peerj-cs.623)).
+This is because R² is a percentage whereas MSE and RMSE have arbitrary ranges.
```
ablaom, line 88: This sounds a bit more like an advertisement than a doc-string 😉 . Feel free to leave it, but probably this coefficient is so-well known the reference is enough. Perhaps "linear regression" is clearer than just "regression".

src/measures/continuous.jl
ablaom: I don't know if there is a commonly accepted convention for weighting rsq. As rsq is non-linear (in the sense that it's value on the union of two sets of observations  is not the sum or mean of it's value on the individual sets) one can argue no particular convention is universally superior to another. So either drop support (my inclination) or be sure to add your formula to the doc-string.

ablaom (changes requested):
Thanks for this contribution @rikhuijzer ! 

As @OkonSamuel correctly points out "aggregation" refers to how scores are combined in when we resample to estimate the expected value of the score on unseen data. So if I am doing 3-fold cross-validation, I'll get 3 R-squared scores and the questions is how should I combine them? Since R-squared is rather non-linear (in the sense I described above) I don't believe there's an obvious choice, so falling back to `Mean()`, as you currently have, is my recommendation. 

If this was number-of-false-positives in classification, you would use `Count()`. 

For `rms` its possible to define an aggregation function `f` with the property that `f([rms(A1), rms(A2), rms(A3)]) = rms(cat(A1, A2, A3))` (assuming `A1`, `A2`, `A3` are data sets of same size) so that case has a special aggregator.

rikhuijzer:
Thanks both for the comments, @ablaom @OkonSamuel :smile:. I've implemented them.

rikhuijzer:
Now that the weighted `call` is removed, should `supports_weights` or `supports_class_weights` also be set to false?

ablaom (approved):
Looks good to me. Thanks again 🙏🏾

ablaom:
> Now that the weighted call is removed, should supports_weights or supports_class_weights also be set to false?

Oops, good catch. They should both be set to `false`. Can you push the change?

ablaom:
Nevermind, I'll do it when I merge your other PR.

PR #697, opened 2021-12-20
Make `PerformanceEvaluation` object a struct

### LINKED ISSUE #408: `evaluate` and `evaluate!` should return a `struct` instead of a `NamedTuple`
with the recent increase in the size of `NamedTuple` returned by `evaluate` and `evaluate!`, I would suggest that these methods return a custom immutable `struct` rather than a `NamedTuple`. The reason for this is so we can customize the display of the struct and define some cool methods for extracting the parameters of the struct. This change should not affect other methods which are dependent on `evaluate` and `evaluate!`.
```julia
julia> struct A
            x::Int
          end
 
julia> st = A(1)
A(1)

julia> nt = (x=1,)
(x = 1,)

julia> st.x
1

julia> nt.x
1

end
```
thoughts Anyone?

ablaom:
Resolves #408.


OkonSamuel (approved):
LGTM!

PR #711, opened 2022-01-05
add target_scitype trait for supervised pipelines

olivierlabayle:
To solve https://github.com/JuliaAI/MLJBase.jl/issues/710

### DIFF
--- a/test/composition/models/pipelines.jl
+++ b/test/composition/models/pipelines.jl
@@ -565,10 +565,6 @@ end
     @test MLJBase.as_type(:unsupervised) == Unsupervised
 end 
 
-@testset "Traits" begin
-    
-end
-
 end
 
 true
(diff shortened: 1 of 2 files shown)


test/composition/models/pipelines.jl
```diff
@@ -554,6 +565,10 @@ end
     @test MLJBase.as_type(:unsupervised) == Unsupervised
 end 
 
+@testset "Traits" begin
+    
+end
+
```
ablaom, line 571: I'm assuming 568-570 can be removed?

ablaom (changes requested):
Looks good to me @olivierlabayle, thanks.

olivierlabayle:
For the future, is that fine to merge it myself once approved or do you prefer to do it?

PR #712, opened 2022-01-06
Stacking internal evaluation

olivierlabayle:
I have played around with the stack again. Related to the new functionality [permitting to expose the internal state of a learning network](https://alan-turing-institute.github.io/MLJ.jl/dev/composing_models/#Exposing-internal-state-of-a-learning-network) that you offered in 0.19. The goal would be to optionally report the internal cross validation result of the stack for a given set of measures.

Even though not complete, I think the following provides a poc for this. Before moving forward I wanted to see if you could give me some feedback and guidance on where to go next. A few ideas that come to my mind:

- Formatting of the report, maybe something matching the output of `evaluate!`?
- Maybe I should check that the measures match the stack type?

I hope you can find some time to look into it and happy to take feedback.

### DIFF
--- a/src/composition/models/stacking.jl
+++ b/src/composition/models/stacking.jl
@@ -30,9 +30,10 @@ mutable struct DeterministicStack{modelnames, inp_scitype, tg_scitype} <: Determ
    models::Vector{Supervised}
    metalearner::Deterministic
    resampling
-   function DeterministicStack(modelnames, models, metalearner, resampling)
+   internal_measures
+   function DeterministicStack(modelnames, models, metalearner, resampling, internal_measures)
         inp_scitype, tg_scitype = input_target_scitypes(models, metalearner)
-        return new{modelnames, inp_scitype, tg_scitype}(models, metalearner, resampling)
+        return new{modelnames, inp_scitype, tg_scitype}(models, metalearner, resampling, internal_measures)
    end
 end
 
@@ -40,9 +41,10 @@ mutable struct ProbabilisticStack{modelnames, inp_scitype, tg_scitype} <: Probab
     models::Vector{Supervised}
     metalearner::Probabilistic
     resampling
-    function ProbabilisticStack(modelnames, models, metalearner, resampling)
+    internal_measures
+    function ProbabilisticStack(modelnames, models, metalearner, resampling, internal_measures)
         inp_scitype, tg_scitype = input_target_scitypes(models, metalearner)
-        return new{modelnames, inp_scitype, tg_scitype}(models, metalearner, resampling)
+        return new{modelnames, inp_scitype, tg_scitype}(models, metalearner, resampling, internal_measures)
     end
  end
 
@@ -135,7 +137,7 @@ evaluate!(mach; resampling=Holdout(), measure=rmse)
 ```
 
 """
-function Stack(;metalearner=nothing, resampling=CV(), named_models...)
+function Stack(;metalearner=nothing, resampling=CV(), internal_measures=nothing, named_models...)
     metalearner === nothing &&
         throw(ArgumentError("No metalearner specified. Use Stack(metalearner=...)"))
 
@@ -144,9 +146,9 @@ function Stack(;metalearner=nothing, resampling=CV(), named_models...)
     models = collect(nt)
 
     if metalearner isa Deterministic
-        stack =  DeterministicStack(modelnames, models, metalearner, resampling)
+        stack =  DeterministicStack(modelnames, models, metalearner, resampling, internal_measures)
     elseif metalearner isa Probabilistic
-        stack = ProbabilisticStack(modelnames, models, metalearner, resampling)
+        stack = ProbabilisticStack(modelnames, models, metalearner, resampling, internal_measures)
     else
         throw(ArgumentError("The metalearner should be a subtype
                     of $(Union{Deterministic, Probabilistic})"))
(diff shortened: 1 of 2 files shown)


ablaom:
Thanks for this further contribution!
 
Will take a look in due course. In the meantime, could you clarify if this is just adding features, or is it also breaking?

olivierlabayle:
No it shouldn't be breaking at all. It should be transparent to the user if using the general `Stack` constructor as it's only a new keyword argument that defaults to no evaluation.

ablaom:
This is a very cool application of the new interface. I'd be kind of surprised if any other stacking software exposes the kind of detail this PR adds. Nice work!

The POC looks good to me and I'd support it more-or-less as is, notwithstanding the two items you already mention.

I do see one wrinkle.  Base learners can be a mixture of probabilistic and deterministic models, yes? So, am I allowed to include a probabilistic measure? If so, I guess it needs to be ignored by the deterministic models. Or do we just rule out probabilistic measures altogether to rule out the added complexity. Probably. 

For the record, `evaluate!` allows the user to specify both kinds of measures if the model is probabilisitic. It also allows you to pass weights to the measures. Maybe you can make call some of that code to handle these kind of case distinctions, add some functionality, and to check the validity of the measures. Can provide further guidance on that code if necessary.

olivierlabayle:
@ablaom Thank you for your patience and sorry for the delay, I think this is now ready for review.

A few notes:

- I finally chose the `PerformanceEvaluation` type return for each model which makes this PR a bit more involved. The implementation is tested by comparing the result with an actual call to `evaluate` with the same strategy.

- I am not sure I have addressed the probabilistic measure issue you mention. I think it should work for a ProbabilisticStack since it currently can only combine probabilistic learners if I recall. For a DeterministicStack it may not as you mention. However I have made use of `_actual_operations` which should select the appropriate operation for the measure. It seems to boil down to weither a model implements the required operation. Do you have any testcase that I could use? I am happy to take your guidance on this, I could probably add a `_cheack_measures` call at the begining of the `fit` to fail fast if you think it is relevant. The problem being that `_cheack_measures` requires `y` and cannot be called at construction time.

- Finally there is no support for weights.

src/composition/models/stacking.jl
```diff
@@ -134,19 +141,29 @@ evaluate!(mach; resampling=Holdout(), measure=rmse)
...
+and provides a PerformanceEvaluation object for each model:
+
+```julia
+report(mach).cv_report
+```
+
 """
-function Stack(;metalearner=nothing, resampling=CV(), named_models...)
+function Stack(;metalearner=nothing, resampling=CV(), measures=nothing, named_models...)
     metalearner === nothing &&
```
ablaom, line 153: Minor suggestion. We could allow the user to specify *singular* `measure=...` by instead doing

```julia
function Stack(;metalearner=nothing, resampling=CV(), resapling=CV(), measure=nothing, measures=measure, named_models...)
```
with no need to change anything downstream.

src/composition/models/stacking.jl
```diff
@@ -253,9 +271,82 @@ pre_judge_transform(ŷ::Node, ::Type{<:Deterministic}, ::Type{<:AbstractArray{<
...
+
+internal_stack_report(m::Stack, verbosity::Int, y::AbstractNode, folds_evaluations::Vararg{Nothing}) = NamedTuple{}()
+function internal_stack_report(m::Stack, verbosity::Int, y::AbstractNode, folds_evaluations::Vararg{AbstractNode})
+    _internal_stack_report(y, folds_evaluations...) =
+        internal_stack_report(m, verbosity, y, folds_evaluations...)
+    return (report=(cv_report=node(_internal_stack_report, y, folds_evaluations...),),)
+end
+
+function internal_stack_report(stack::Stack{modelnames,}, verbosity::Int, y, folds_evaluations...) where modelnames
+
```
ablaom, line 287: I'm struggling a bit to understand what the role of each `internal_stack_report` is doing. There is one on 279, 282, and 286. I think a few code comments here might help me. 

I'm also struggling to find where the `fold_evaluations` are transformed from a list of `Nodes` to a list of "concrete" tuples. That is, where does the actual evaluation happen?

ablaom:
> @ablaom Thank you for your patience and sorry for the delay, I think this is now ready for review.
> 
> A few notes:
> 
> * I finally chose the `PerformanceEvaluation` type return for each model which makes this PR a bit more involved. The implementation is tested by comparing the result with an actual call to `evaluate` with the same strategy.

Awesome!

> * I am not sure I have addressed the probabilistic measure issue you mention. I think it should work for a ProbabilisticStack since it currently can only combine probabilistic learners if I recall. For a DeterministicStack it may not as you mention. However I have made use of `_actual_operations` which should select the appropriate operation for the measure. It seems to boil down to weither a model implements the required operation. Do you have any testcase that I could use? 

Looking over your tests, I think you have addressed my original issue. 

> I am happy to take your guidance on this, I could probably add a `_cheack_measures` call at the begining of the `fit` to fail fast if you think it is relevant. The problem being that `_cheack_measures` requires `y` and cannot be called at construction time.

I think such a check would probably be a good idea.

> * Finally there is no support for weights.

Could always be added later. This PR already exceeds my expectations.

ablaom (changes requested):
@olivierlabayle Thanks for going the extra mile here to recover the full `PerformanceEvaluation` objects. 

Your tests look pretty comprehensive to me and I'm happy with what codecov is reporting.

src/composition/models/stacking.jl
olivierlabayle: I added some doctrings and hope this is now easier to understand.

olivierlabayle:
As discussed, I found a usecase and added `check_stack_measures` call at the beginning of `fit`. I also updated the `setproperty!` method to encompass the `measures` field which was missing.

ablaom (approved):
All looks good to me. Thanks!

PR #723, opened 2022-01-20
Get rid of `ErrorNode`s and fix corner case bug in `replace`

ablaom:
In the recent implementation of `Pipeline` we introduced `ErrorNode <: AbstractNode` as a mechanism for throwing errors when an `UnsupervisedPipeline` did not support `inverse_transform`. This is not part of the public API.

This PR:

- Get's rid of `ErrorNodes` by allowing one to instead wrap an `Exception` object in an ordinary `Source` node. Calling the node in that case always *throws* the exception . This is a minor conceptual simplification. It also makes the next fix a little more straightforward. 

-  Fixes a bug in `replace`. This is an internal method used only for composite models exported using `@from_network`. The bug concerns the corner case where an "operation node" is also a `Source` node, and when there were previously `ErrorNode`s. 

**Some details.** The main change to `replace` is to iterate over *all* nodes, not just non-source nodes, when progressively reconstructing the network, and to add logic in the loop that stops the loop looking for ancestors in the special case of a `Source` node.

### DIFF
--- a/src/composition/learning_networks/inspection.jl
+++ b/src/composition/learning_networks/inspection.jl
@@ -27,7 +27,6 @@ function tree(W::Node)
     return NamedTuple{keys}(values)
 end
 tree(s::Source) = (source = s,)
-tree(n::ErrorNode) = (node = n,)
 
 # """
 #    args(tree; train=false)
--- a/src/composition/learning_networks/machines.jl
+++ b/src/composition/learning_networks/machines.jl
@@ -394,10 +394,8 @@ function return!(mach::Machine{<:Surrogate},
 
     verbosity isa Nothing || fit!(mach, verbosity=verbosity)
 
-    # anonymize the data, except at source nodes wrapping exceptions:
-    sources = filter(MLJBase.sources(glb(mach))) do s
-        !(s.data isa Exception)
-    end 
+    # anonymize the data
+    sources = MLJBase.sources(glb(mach))
     data = Tuple(s.data for s in sources)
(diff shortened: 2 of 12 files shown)


olivierlabayle:
Thanks Anthony for your swift reaction.

- Can I ask what would happen if we were **not** to have nodes containing exceptions? It seems to me the only problem is that the `inverse_transform` would not be defined in the signature but I am not sure why this is a problem?
- I like the idea of looping over all nodes. In light of reusing the `replace` function (or part of) for serialization we might on top of that dispatch the inner working of the loop on the Node type to update the various mappings adequatly. This can be done later on though.
- Regarding the potential removal of anonymization, do you mean that we may be able to remove the `cache.data` field on top of changing at least the `return!` and `update` functions? I would be up for it if possible since a serialisable machine will be deprived of any data anyway. However I didn't really have any use case for the anonymization stage in the first place so I might be missing something.

ablaom:
I have found a way to abandon the error-throwing source nodes and so remove more case distinctions. I agree this is a better solution. 

I have also relaxed the condition `state > 0` for operations called on machines  to `state !=0 `, which should address your concern at https://github.com/JuliaAI/MLJSerialization.jl/pull/17#discussion_r789005641 . Please confirm this is indeed the case. 

I am posting the new commits very shortly.

olivierlabayle (approved):
Thanks, it looks good to me!

I have also tested the predict function in MLJSerialization with the new state and it works fine !

PR #724, opened 2022-01-21
Drop `filename` arg from save(filename, model, ...)

ablaom:
In support of https://github.com/JuliaAI/MLJSerialization.jl/pull/17 .

Note PR is onto `for-a-0-point-20-release` branch, not onto `dev`.

cc @olivierlabayle

### DIFF
--- a/src/interface/model_api.jl
+++ b/src/interface/model_api.jl
@@ -31,26 +31,35 @@ predict_median(m, fitresult, Xnew, ::Type{<:BadMedianTypes}) =
 MLJModelInterface.implemented_methods(::FI, M::Type{<:MLJType}) =
     getfield.(methodswith(M), :name) |> unique
 
-# the following serialization fallbacks should live in
-# MLJModelInterface when version 2.0 i s released. At that time the
+# The following serialization fallbacks should live in
+# MLJModelInterface when version 2.0 is released. At that time the
 # hack block could also be removed.
+
 #####################
 # hack block begins #
 #####################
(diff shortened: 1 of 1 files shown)


src/interface/model_api.jl
```diff
@@ -31,8 +31,24 @@ predict_median(m, fitresult, Xnew, ::Type{<:BadMedianTypes}) =
...
+    "a model from XGBoost.jl, be sure to use MLJXGBoostInterface 2.0 or "*
+    "or higher. "
+)
+MLJModelInterface.save(filename, model, fitresult; kwargs...) =
+    throw(ERR_SERIALIZATION_FAILURE)
+###################
+# hack block ends #
+###################
+MLJModelInterface.save(model, fitresult; kwargs) = fitresult
 MLJModelInterface.restore(filename, model, serializable_fitresult) =
```
olivierlabayle, line 52: I think we will also want to drop the `filename` from `restore` right?

ablaom: Right you are.

olivierlabayle (approved):
Looks good to me. After this is merged, Would it be possible to rebase the https://github.com/JuliaAI/MLJBase.jl/tree/for-a-0-point-20-release on dev (or merging dev on this branch) so that I can develop MLJSerialization using this branch as a dependency?

ablaom:
> After this is merged, Would it be possible to rebase the https://github.com/JuliaAI/MLJBase.jl/tree/for-a-0-point-20-release on dev ?

Done.

PR #727, opened 2022-01-23
Fix the docs deployment

### LINKED ISSUE #726: Documentation is not deploying properly.
In the log we are getting "ERROR: LoadError: The repo path https://github.com/JuliaAI/MLJBase.jl should not contain the protocol".

No idea what this means, but I found this: https://discourse.julialang.org/t/documenter-jl-deployment-issue-with-repo-containing-protocol/64367

DilumAluthge:
Fixes #726

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -19,4 +19,4 @@ makedocs(;
     sitename="MLJBase.jl"
 )
 
-deploydocs(; repo="$(REPO).git", push_preview=false)
+deploydocs(; repo=REPO, push_preview=false)
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,7 +1,7 @@
 using Documenter
 using MLJBase
 
-const REPO="https://github.com/JuliaAI/MLJBase.jl"
+const REPO="github.com/JuliaAI/MLJBase.jl"
 
 makedocs(;
     modules=[MLJBase],
@@ -15,7 +15,7 @@ makedocs(;
         "Distributions" => "distributions.md",
         "Utilities" => "utilities.md"
     ],
-    repo="$REPO/blob/{commit}{path}#L{line}",
+    repo="https://$REPO/blob/{commit}{path}#L{line}",
     sitename="MLJBase.jl"
 )
 


ablaom:
Oops. Already slipped in a possible fix before seeing this one.

Thanks. If my fix doesn't work, I'll fall back to this one. I'm not sure I understand it though, as is seems not to make a change to the final `repo` string. The "protocol" `https:\\" is still there.

DilumAluthge:
See line 22. The "https://" will not be passed when we call `deploydocs`, which is the important part.

DilumAluthge:
The `repo` kwarg to `deploydocs` is a different concept from the `repo` kwarg to `makedocs`.

DilumAluthge:
In `makedocs`, you do still want to pass the "https://", which is why I'd recommend reverting the change you pushed to `dev`.

DilumAluthge:
In the `repo` kwarg to `deploydocs`, you should not include the `.git` suffix.

DilumAluthge:
PkgTemplates.jl will generate a `docs/make.jl` file that is compatible with the Documenter.jl requirements.

Here is a repository that I generated recently using PkgTemplates: https://github.com/bcbi/TableCellCounts.jl/blob/51ac308adbbc57f3bd4b21d3105f613079cf3458/docs/make.jl

OkonSamuel:
> In the `repo` kwarg to `deploydocs`, you should not include the `.git` suffix.

I guess both ways work then. I found `.git` option in the Documenter documentation

ablaom:
@OkonSamuel @DilumAluthge I'm happy for you two to take over here, thanks!

DilumAluthge:
> > In the `repo` kwarg to `deploydocs`, you should not include the `.git` suffix.
> 
> I guess both ways work then. I found `.git` option in the Documenter documentation https://github.com/JuliaDocs/Documenter.jl/blob/5e5d2db1d0600cddec14b4ea3139bc5acc48e401/src/Documenter.jl#L381-L383

Ah, good find! Yeah, it looks like either with or without will work. I think that PkgTemplates generates a template without, which is probably why that's why I tend to choose.

OkonSamuel (approved):

PR #731, opened 2022-01-25
remove specific model checks for general scitype check

pazzo83:
First pass at simplifying type checks for machines - we simply check the expected scitype signature instead of specific checks for supervised, unsupervised, etc.

### DIFF
--- a/src/machines.jl
+++ b/src/machines.jl
@@ -93,9 +93,17 @@ warn_scitype(model::Supervised, X) =
     "input_scitype(model) = $(input_scitype(model))."
 
 warn_generic_scitype_mismatch(S, F) =
-    "The scitype of `args` in `machine(model, args...; kwargs)` "*
-    "does not match the scitype "*
-    "expected by model's `fit` method.\n"*
+    "The number and/or types of data arguments do not " *
+    "match what the specified model supports. Commonly, " *
+    "but non exclusively, supervised models are constructed " *
+    "using the syntax `machine(model, X, y)` or `machine(model, X, y, w)` " *
+    "while most other models with `machine(model, X)`. " *
+    "Here `X` are features, `y` a target, and `w` sample or class weights. " *
+    "In general, data in `machine(model, data...)` must satisfy " *
+    "`scitype(data) <: MLJ.fit_data_scitype(model)` unless the " *
+    "right-hand side is `Unknown`.  Here, the scitype of `args` " *
+    "in `machine(model, args...; kwargs)` does not match the scitype " *
+    "expected by model's `fit` method.\n" *
     "  provided: $S\n  expected by fit: $F"
 
 warn_scitype(model::Supervised, X, y) =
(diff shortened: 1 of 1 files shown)


src/machines.jl
```diff
@@ -93,9 +93,17 @@ warn_scitype(model::Supervised, X) =
...
+    "but non exclusively, supervised models are constructed " *
+    "using the syntax `machine(model, X, y)` or `machine(model, X, y, w)` " *
+    "while most other models with `machine(model, X)`. " *
+    "Here `X` are features, `y` a target, and `w` sample or class weights. " *
+    "In general, data in `machine(model, data...)` must satisfy " *
+    "`scitype(data) <: MLJ.fit_data_scitype(model)` unless the " *
+    "right-hand side is `Unknown`.  Here, the scitype of `args` " *
+    "in `machine(model, args...; kwargs)` does not match the scitype " *
+    "expected by model's `fit` method.\n" *
     "  provided: $S\n  expected by fit: $F"
```
ablaom, line 107: I think we can drop the last sentence (the original message) as it is rendered reduntant by the preceding sentences.

src/machines.jl
```diff
@@ -93,9 +93,17 @@ warn_scitype(model::Supervised, X) =
...
-    "The scitype of `args` in `machine(model, args...; kwargs)` "*
-    "does not match the scitype "*
-    "expected by model's `fit` method.\n"*
+    "The number and/or types of data arguments do not " *
+    "match what the specified model supports. Commonly, " *
+    "but non exclusively, supervised models are constructed " *
+    "using the syntax `machine(model, X, y)` or `machine(model, X, y, w)` " *
+    "while most other models with `machine(model, X)`. " *
+    "Here `X` are features, `y` a target, and `w` sample or class weights. " *
+    "In general, data in `machine(model, data...)` must satisfy " *
```
ablaom, line 102: Line 102: Put "data" in back-quotes: "\`data\`".

src/machines.jl
ablaom, line 107: How about we replace the last sentence (from the original string, which is using different notation) with:

"In the present case:\n"*
"`scitype(data) = $S`\n"*
"and\n"*
"`fit_data_scitype(model) = $F`"

src/machines.jl
```diff
@@ -93,9 +93,17 @@ warn_scitype(model::Supervised, X) =
...
+    "`scitype(data) <: MLJ.fit_data_scitype(model)` unless the " *
```
ablaom, line 103: We can drop the qualifier `MLJ.` from `MLJ.fit_data_scitype(model)` as this is now exported.

ablaom:
@pazzo83 Regarding the failing log tests. Rather than something like

```julia
(:warn, r"The scitype of `X`")
```

let's make this robust to changes to the doc-string by instead doing something like

```julia
(:warn, MLJBase.warn_generic_scitype_mismatch(S, F))
```
and `S` and `F` have been defined appropriately for the test.

Make sense?

ablaom:
Okay, after looking again at the code, https://github.com/JuliaAI/ScientificTypes.jl/issues/182 is not an issue. I thought we were doing `S = scitype(args)`, but we actually have https://github.com/JuliaAI/MLJBase.jl/blob/3b85789f9c3b2b679344d089e06f8bde0d3e5941/src/machines.jl#L135, which should avoid the problem.

So, I think your fail is what I first suggested: outdated logging tests. The error message thrown has changed, so we have to update the `@test_logs`, no?

ablaom:
Hang on. I think `S = Tuple{elscitype.(args)...} ` should instead be `S = Tuple{scitype.(args)...}`, another bug.

pazzo83:
Yeah I had originally changed it to `scitype` but I keep getting warnings that specify this difference:
```
scitype(data) = Tuple{CallableReturning{Table{AbstractVector{Continuous}}}, CallableReturning{AbstractVector{Multiclass{2}}}}
```
vs
```
fit_data_scitype(model) = Union{Tuple{Table, AbstractVector{<:Finite}}, Tuple{Table, AbstractVector{<:Finite}, AbstractVector{Union{Continuous, Count}}}}
```

ablaom:
@pazzo83 I'll have a proper look into this tomorrow, thanks.

ablaom:
@pazzo83 Looking into this now and making progress. Thanks for your continued patience.

ablaom:
Closing in favour of #732

PR #733, opened 2022-01-28
Serialization

olivierlabayle:
Context: https://github.com/JuliaAI/MLJSerialization.jl/issues/15 

This PR contains:

- Routines for the serialization of machines/composite machines
- A small refactoring of the `Base.replace` to reuse some code for `save` of Composite models

Unfortunately I don't think we have CI enabled since this does not target dev. 

Happy to take your comments.


src/composition/learning_networks/machines.jl
```diff
@@ -414,9 +414,127 @@ end
...
+function copy_or_replace_machine(N::AbstractNode, newmodel_given_old, newnode_given_old)
+    train_args = [newnode_given_old[arg] for arg in N.machine.args]
+    return Machine(newmodel_given_old[N.machine.model],
+                train_args...)
+end
+
+"""
+    copy_or_replace_machine(N::AbstractNode, newmodel_given_old::Nothing, newnode_given_old)
+
+For now, two top functions will lead to a call of this function: `Base.replace` and
```
ablaom, line 428: line 428: Suggest `Base.replace` -> `Base.replace(::Machine, ...)`

src/composition/learning_networks/machines.jl
```diff
@@ -414,9 +414,127 @@ end
...
+`save`. If `save` is the calling function, the argument `newmodel_given_old` will be nothing
```
ablaom, line 429: line 429: Suggest replace first `save` with `save(::Machine, ...)`.

src/composition/learning_networks/machines.jl
```diff
@@ -414,9 +414,127 @@ end
...
+and the goal is to make the machine in the current learning network serializable.
```
ablaom, line 430: Maybe it's clearer if insead of "current learning network", you say "machine attached to node `N`". Isn't that what you mean here?

src/composition/learning_networks/machines.jl
```diff
@@ -414,9 +414,127 @@ end
...
+                train_args...)
+end
+
+"""
+    copy_or_replace_machine(N::AbstractNode, newmodel_given_old::Nothing, newnode_given_old)
+
+For now, two top functions will lead to a call of this function: `Base.replace` and
+`save`. If `save` is the calling function, the argument `newmodel_given_old` will be nothing
+and the goal is to make the machine in the current learning network serializable. 
+This method will be called. If `Base.replace` is the calling function, then `newmodel_given_old`
```
ablaom, line 431: Re : "This method will be called". Do you mean to delete this sentence. It's seems out of place. 

Add sentence saying method is mutating in case of `replace` but not in case of `save`.

ablaom: Thanks for these explanations, by the way.

src/composition/learning_networks/machines.jl
```diff
@@ -414,9 +414,127 @@ end
...
+"""
+    update_mappings_with_node!(
+        newnode_given_old, 
+        newmach_given_old, 
+        newmodel_given_old, 
+        N::AbstractNode)
+
+For Nodes that are not sources, update the appropriate mappings 
+between elements of the learning to be copied and the copy itself.
+"""
```
ablaom, line 449: Just a suggestion. Maybe the work "connections" is more suggestive that "mappings"??

src/composition/learning_networks/machines.jl
```diff
@@ -414,9 +414,127 @@ end
...
+
+"""
+    update_mappings_with_node!(
+        newnode_given_old, 
+        newmach_given_old, 
+        newmodel_given_old, 
+        N::AbstractNode)
+
+For Nodes that are not sources, update the appropriate mappings 
+between elements of the learning to be copied and the copy itself.
```
ablaom, line 448: "learning" -> "learning network"

src/composition/learning_networks/machines.jl
```diff
@@ -414,9 +414,127 @@ end
...
+
+update_mappings_with_node!(
+    newnode_given_old, 
+    newmach_given_old, 
+    newmodel_given_old, 
+    N::Source) = nothing
+
+"""
+    copysignature(signature, newnode_given_old; newmodel_given_old=nothing)
+
```
ablaom, line 477: We should probably call this `copysignature!` and state below that only the dictionaries are mutated.

src/composition/learning_networks/machines.jl
```diff
@@ -476,58 +587,65 @@ function Base.replace(mach::Machine{<:Surrogate},
...
+
+"""
+    save(model::Composite, fitresult)
+
+Returns a new `CompositeFitresult` that is a shallow copy of the original one.
+To do so,  we build a copy of the learning network where each machine contained 
+in it needs to be called `serializable` upon.
+
+Ideally this method should "reuse" as much as possible `Base.replace`.
```
ablaom, line 614: I guess you can now delete line 614

src/machines.jl
```diff
@@ -378,6 +378,25 @@ function machine(model::Model, arg1::AbstractNode, args::AbstractNode...;
...
+    machine(file::Union{String, IO}, raw_arg1=nothing, raw_args...)
+
+Rebuild from a file a machine that has been serialized using the default 
+Serialization module.
+"""
+function machine(file::Union{String, IO}, raw_arg1=nothing, raw_args...)
+    smach = deserialize(file)
+    smach.state == -1 || 
+        @warn "Deserialized machine state is not -1 (=$(smach.state)). "*
+             "It means that the machine has not been saved by a conventional MLJ routine."
```
ablaom, line 391: "It" -> "This"

I'd add to the warning: "For example, it's possible original training data is accessible from the deserialised object. "

src/machines.jl
```diff
@@ -796,3 +815,154 @@ function training_losses(mach::Machine)
...
+
+
+"""
+    serializable(mach::Machine)
+
+Returns a shallow copy of the machine to make it serializable, in particular:
+    - Removes all data from caches, args and data fields
+    - Makes all `fitresults` serializable
+    - Annotates the state as -1
+"""
```
ablaom, line 832: I think we can make this doc-string a little more user-friendly, as this for the general user. The only field of `mach` that is ostensibly "public" is the `model` field. 

How about:

> Returns a shallow copy of the machine to make it serializable.  In particular, all training data is removed and, if necessary,  learned parameters are replaced with persistent representations. 

> A machine `mach` is serializable exactly when `mach.state = -1`.

The other comments could be made private "#" comments.

src/machines.jl
```diff
@@ -796,3 +815,154 @@ function training_losses(mach::Machine)
...
+    MLJ.save(filename, mach::Machine)
+    MLJ.save(io, mach::Machine)
+
+    MLJBase.save(filename, mach::Machine)
+    MLJBase.save(io, mach::Machine)
+
+Serialize the machine `mach` to a file with path `filename`, or to an
+input/output stream `io` (at least `IOBuffer` instances are
+supported) using the Serialization module.
+
```
ablaom, line 884: Let's add a second paragraph reading:

> To serialise using a different format, see [`serializable`](@ref).

test/composition/learning_networks/machines.jl
```diff
@@ -230,6 +231,142 @@ enode = @node mae(ys, yhat)
 end
 
 
+@testset "Test serializable of pipeline" begin
+    filename = "pipe_mach.jls"
+    X, y = TestUtilities.simpledata()
+    pipe = (X -> coerce(X, :x₁=>Continuous)) |> DecisionTreeRegressor()
+    mach = machine(pipe, X, y)
```
ablaom, line 238: Did you decide against a proper transformer here?

test/composition/learning_networks/machines.jl
```diff
@@ -230,6 +231,142 @@ enode = @node mae(ys, yhat)
...
+    mach = machine(model, X, y)
+    fit!(mach, verbosity=0)
+    serialize(filename, mach)
+    @test filesize(filename) > filesizes[1]
+
+    @test_logs (:warn, "Deserialized machine state is"*
+                       " not -1 (=1). It means that the"*
+                       " machine has not been saved by a"*
+                       " conventional MLJ routine.") machine(filename)
+
```
ablaom, line 367: Can you please define a warning string in the src, as in 

```julia
warn_bad_deserialization(state) = "Deserialized machine .... (got $state). This means that ..."
```

which you call in src in place of the literate string you have already. 

In your `@test_logs` do instead

```julia
 @test_logs (:warn, MLJBase.warn_bad_deserialization(1)), machine(filename)
```

This is better for maintenance. We're now trying to adhere to this practice with all errors and logging.

test/machines.jl
```diff
@@ -262,6 +262,86 @@ end
 
 end
 
+@testset "Test serializable method of Supervised Machine" begin
+    X, y = TestUtilities.simpledata()
```
ablaom, line 266: Instead of `simpledata` you can do `X, y = make_regression(100, 1)` and dump `simpledata`.

ablaom (commented):
Mammoth effort, thanks. 

I'm sorry for not catching this earlier but, unless I am missing something, there is no point in including `mach.cache` in machines to be serialised. In particular, `serializable` can return a machine with `cache` undefined. Once a machine has been trained, the only time `cache` is ever used is when we update the machine (instead of training ab initio). Once the original training data is gone a machine "update" no longer makes sense, it seems to me, and any attempt to `fit!` the machine should start from scratch (and hence not need `cache`). 

If you're not convinced, let's have another call to clarify.

ablaom:
What happens if the user attempts to:

- call `fit!` on a machine returned by `serializable`?
- call on operation, such as `predict`, on such an object?

src/composition/learning_networks/machines.jl
olivierlabayle: right!

src/composition/learning_networks/machines.jl
olivierlabayle: I don't think any of the `copy_or_replace_machine` is actually mutating its argument but agree that the sentence is unclear. I have added a docstring to the other method and tried to make those a bit more intelligible.

src/composition/learning_networks/machines.jl
olivierlabayle: right!

ablaom:
And I'm thinking that we should restore `state=0` when we `restore!` a machine it ensure `fit!(mach)` performs training from scratch; see https://github.com/JuliaAI/MLJBase.jl/blob/c8a2d56bfc5862c7e4a2adde98ded299ee4eb3fb/src/machines.jl#L583 .

And in the code just referenced we could catch `state==-1` and throw an error that machine needs restoring, yes?

test/composition/learning_networks/machines.jl
olivierlabayle: Since I cover a `Standarizer` in `test/machines.jl` line 299 I thought this would just cover more cases and kept it.

src/machines.jl
olivierlabayle: I have added an example with JLSO

olivierlabayle:
> What happens if the user attempts to:
> 
>     * call `fit!` on a machine returned by `serializable`?
> 
>     * call on operation, such as `predict`, on such an object?
> 
>     * call `serializable` on a machine that has an `Node` as an argument?  (I wouldn't think it would ever make sense to do that. )

I think everything works as I would expect:

- Fails because no data is bound to it
- Works fine as long as data is provided.
- Works fine

Do you want to add testcases for this?

olivierlabayle:
> And I'm thinking that we should restore `state=0` when we `restore!` a machine it ensure `fit!(mach)` performs training from scratch; see
> 
> https://github.com/JuliaAI/MLJBase.jl/blob/c8a2d56bfc5862c7e4a2adde98ded299ee4eb3fb/src/machines.jl#L583
> .
> 
> And in the code just referenced we could catch `state==-1` and throw an error that machine needs restoring, yes?

Currently because `rows_is_new` will be true, training will already happen from scratch again so I think there is technically no problem.

If we restore the state to 0 I think the various operations won't work anymore while the machine has already been fitted, since those operations have been changed to support `state != 0`.

From our discussion from last time I think you said that MLJ was not supporting incremental training. However in the `fit_only` process, the state is incrementally updated (`src/machines.jl` line 663) while I would expect it to be simply set to 1 if this was the case. This is a bit of a problem since the deserialized machine has state -1, fitting it will result in state 0 which is not appropriate indeed. It also seems that `MLJModelInterface.update` falls back to a full fit but that there is technically nothing preventing the user to define an update method that would allow incremental learning so I think I may have misunderstood something?

test/composition/learning_networks/machines.jl
ablaom: 👍🏾

ablaom:
> So, I have tried and it turns out the cache is important for Composite models at least (I also thought it would be important for TunedModels but it turns out not to be the case). The main problem is the network_model_names which lives in the cache and which is used for report and fitted_params. I agree however that it makes intuitive sense to delete the cache for serialization. It means we must move the network_model_names field though (perhaps to CompositeFitresult) or construct it again if not in the cache. Waiting for your thoughts.

Thanks for identifying the problem here.  After reviewing the code, I believe `network_mode_name` can be moved from cache (where it logically has no place) to  a field of `CompositeFitResult`. I've just grepped the method: the relevant files are: src/composition/models/methods.jl, /scr/composition/learning_networks/machines.jl, /src/composition/models/inspection.jl, test/composition/models/methods.jl .

By the way, I think ultimately we will be removing `data` from `cache` as well, as we will drop the existing data anonymization dance. Your `serialization` method already takes care of this, right?

ablaom:
> Currently because rows_is_new will be true, training will already happen from scratch again so I think there is technically no problem.

👍🏾 

> If we restore the state to 0 I think the various operations won't work anymore while the machine has already been fitted, since those operations have been changed to support state != 0.

You are right, good point.

> From our discussion from last time I think you said that MLJ was not supporting incremental training. However in the fit_only process, the state is incrementally updated (src/machines.jl line 663) while I would expect it to be simply set to 1 if this was the case.

The `state`s of each machine downstream of a given machine `mach` is stored in the `old_upstream_state` field, which is used to determine if some downstream machine has been trained since the last time  `mach` is trained. This is important information for the "smart fitting" logic. For this reason, each time a machine is fit or updated, the state must increase - it cannot be stuck at 1. (Perhaps it is confusing to include "trained 3 times" in the show for machines, as I guess the precise number is really an implementation detail. Do you agree? Perhaps, we should just say "trained", "untrained", or "deserialised", yes?) 


To clarify, there is two kinds of machine "updates": (i) in response to new data, and (ii) in response to new model hyper parameters  - eg, increase in `epochs` - where data is fixed and the *view* of the data (`rows`) is not changing. These are different and require different interfaces. Currently (i) is not supported as there is no machine interface point to "inject more data" (such as `inject!(mach, Xadditional, yadditional)`) and no separate model method "update_data" (only a stub [here](https://github.com/JuliaAI/MLJModelInterface.jl/blob/d207f88d0646f955900de95d40e22210c38b445c/src/model_api.jl#L51) to support this in the future - see also [this proposal](https://github.com/alan-turing-institute/MLJ.jl/issues/60#issuecomment-549589603)). 

It's true that `machine("mymachine.jlso", X, y)` is an opportunity to "inject" more data, but that was not introduced for this purpose - only to give the user an opportunity to retrain *from scratch* a  model (set of hyper-parameters) that had been serialised with some new data. Perhaps this is confusing and we should just drop it? The model can be extracted, as in `machine("mymachine.jlso").model` and rebound to new machine if needed. 

> This is a bit of a problem since the deserialized machine has state -1, fitting it will result in state 0 which is not appropriate indeed. 

If we dump syntax `machine("myymachine.jlso", data...)` - and just have `machine("mymachine")` - then this point is moot. Alternatively, we can change the training logic to replace `-1` to `1` instead of `0`. I'm kind of leaning to the former.  What do you think?

src/composition/learning_networks/machines.jl
ablaom: Ah, yes, you are right. Got it.

src/composition/learning_networks/machines.jl
```diff
@@ -480,9 +480,11 @@ update_mappings_with_node!(
     N::Source) = nothing
 
 """
-    copysignature(signature, newnode_given_old; newmodel_given_old=nothing)
+    copysignature!(signature, newnode_given_old; newmodel_given_old=nothing)
 
-Copies the given signature of a learning network.
+Copies the given signature of a learning network. Contrary to Julia's convention,
+this method is actually mutating `newnode_given_old`` and `newmodel_given_old`` and not
```
ablaom, line 486: Are there some misplaced back-quotes here?

src/machines.jl
```diff
@@ -825,12 +830,43 @@ end
...
+Any general purpose Julia serialization may be applied to the output of 
+serializable (eg, JLSO, BSON, JLD) but you must call restore!(mach) on 
+the deserialised object mach before using it. See the example below.
+
+If using Julia's standard Serialization library, a shorter workflow is 
+available using the [`save`](@ref) method.
+
+A machine returned by serializable is characterized by the property mach.state == -1.
+
```
ablaom, line 845: Add back quotes

src/machines.jl
```diff
@@ -825,12 +830,43 @@ end
...
+### Example using [JLSO](https://invenia.github.io/JLSO.jl/stable/)
+
+    using MLJ
+    using JLSO
+    tree = @load DecisionTreeClassifier
```
ablaom, line 850: Something not right here. `tree` needs to be a model *instance* not a type. So:

```julia
Tree = @load DecisionTreeClassifier
tree = Tree()
``` 

or 

```julia
tree = (@load DecisionTreeClassifier)()
```

Possibly you adapted an existing example that also has this old syntax?

Always a good idea to check your examples run 😁

src/machines.jl
```diff
@@ -825,12 +830,43 @@ end
...
+    X, y = @load_iris
+    mach = fit!(machine(tree, X, y))
+
+    # This machine can now be serialized
+    smach = serializable(mach)
+    JLSO.save("machine.jlso", machine => smach)
+    
+    # Some fitresults may have to be restored
```
ablaom, line 858: I try to avoid using the work `fitresult` in public documentation as this field not really public. Maybe:

```
# Deserialize and restore learned parameters to useable form:
```

src/machines.jl
```diff
@@ -894,11 +932,11 @@ constructor for retraining on new data using the saved model.
     X, y = @load_iris
```
ablaom, line 932: Ah here's that old `@load` syntax. See comment above.

ablaom (commented):
Thanks for these latest changes and continued patience.

src/machines.jl
olivierlabayle: Yep, taken from the old `save` method from MLJSerialization, changed in both places!

olivierlabayle:
Thank you for those in depth explanations! Here is the current state of the code:

- `network_model_names`: I have moved it to `CompositeFitresult` as decided. I also took the liberty to remove the `report_additions` field from the `CompositeFitresult` since it plays no role in any operation. I just call the report node when building the report of the Surrogate. The only problem is that [the following test](https://github.com/JuliaAI/MLJBase.jl/blob/b87339a8eab8219ecf195fd30e25f700c7dcd5ab/test/composition/models/from_network.jl#L586) fails. It still throws an error but I can't figure out why it's not the original one. That would be great if you could have a look, I can't even find the definition of the `CompositeException` which seems to be thrown.

- `cache`: With the previous change I have now made the cache from a serializable machine undefined. So yes the `serializable` procedure removes all data whatsoever from any field that could contain some since all those fields are now undefined.

- `machine(filename)`: I removed the possibility to provide additional arguments as I agree it makes very little sense at the present time.

- `state`: Thanks for the detailed description, I now understand much better the purpose of this field! However, it makes me feel we are trying to take too much out of it. The original rationale behind setting its value to `-1` was to know weither a machine has been built by `serializable` or not. From an internal point of view, why would we want to know that? Is it to provide an **informative** warning for operations (given data, otherwise it will fail fast anyway) that may fail with **uninformative** message because the fitresult is in serializable form? It is the only corner case I would think of. If we want to tackle this I think we should:
   - Reset the state back to 1 for all machines when restoring
   - Add a warning in operations when it is called with `state=-1`. In most cases it should still work fine I believe since the default `save` is just to return the original fitresult. By the way, if we set the XGBoost case aside, is there any current model that requires the non default `save` method?

What do you think?

ablaom:
> I have moved it to CompositeFitresult as decided.

👍🏾 

>  I also took the liberty to remove the report_additions field from the CompositeFitresult since it plays no role in any operation.

True, it does not play a role in any operation. But it needs to be available to `update(model::Union{Composite,Surrogate}, ...)` so it can be updated in a warm restart: https://github.com/JuliaAI/MLJBase.jl/blob/c8a2d56bfc5862c7e4a2adde98ded299ee4eb3fb/src/composition/models/methods.jl#L50 . I suppose logically it would belong to `cache` but then you need to change `update!` (the method you should put back) to see both `cache` and `fitresult`, as it needs the `signature`. 

BTW, `CompositeException` is a Julia type for collecting multiple exceptions thrown by a `Task`. It's not defined by MLJ.

> cache: With the previous change I have now made the cache from a serializable machine undefined. So yes the serializable procedure removes all data whatsoever from any field that could contain some since all those fields are now undefined.

👍🏾 

> machine(filename): I removed the possibility to provide additional arguments as I agree it makes very little sense at the present time.

Yeah. Doing `machine(mach.model, X, y)` is not hard and cognitively clearer.

>    - Reset the state back to 1 for all machines when restoring
 > - Add a warning in operations when it is called with `state=-1`. In most cases it should still work fine I believe since the default `save` is just to return the original fitresult. By the way, if we set the XGBoost case aside, is there any current model that requires 

An excellent proposal. Warning could be something like "The operation `$operation` has been called on a deserialised machine `mach` whose learned parameters may be unusable.  To be sure, first run `restore!(mach)`."??

ablaom:
> By the way, if we set the XGBoost case aside, is there any current model that requires the non default save method?

I'm pretty sure not, but I will locally test the new API on all registered models when the PR's are ready.

olivierlabayle:
>  I suppose logically it would belong to `cache` but then you need to change `update!` (the method you should put back) to see both `cache` and `fitresult`, as it needs the `signature`.

I haven't put it into the cache either. I actually have removed this method too and updated the `update` function for composite models instead.

> An excellent proposal. Warning could be something like "The operation $operation has been called on a deserialised machine mach whose learned parameters may be unusable. To be sure, first run restore!(mach)."??

Yes that sounds good!

ablaom:
> I haven't put it into the cache either. I actually have removed this method too and updated the update function for composite models instead.

Oh, right. I do see that. I'll have a closer look at this and your fail then.

olivierlabayle:
> Oh, right. I do see that. I'll have a closer look at this and your fail then.

I finally put my hand on the problem. The reason why the test did not pass was because the `network_model_names` function needs to be called before `fit!` in the `return!` function. Like [now](https://github.com/JuliaAI/MLJBase.jl/blob/6a32f25284534ca6e55f409b875bb9729030347c/src/composition/learning_networks/machines.jl#L389). However, unless I am missing something, I believe the problem is the test itself which raises anyway because y is not a table.

In any case I have also added the restoration to `state=1` and the warning we were talking about.

ablaom:
> I finally put my hand on the problem.

🚀  ❤️ I'm glad I was busy with something else yesterday 😄 . This must have been a woozy.
 
>  I believe the problem is the test itself which raises anyway because y is not a table.

You are right! Great catch. This slipped by because the `Standardizer` in the local test suite only handles tables, whereas the most recent version in MLJModels, handles vectors as well. I should have been more careful.

I'm working on fix for this test now. As this is orthogonal to the current PR I will make a PR on a separate branch and rebase `for-a-0-point-20-release`. 

Given the complexity of the current PR, I should like to go over this one more time before merge. However, this is definitely in good shape and I will move on to review the supporting PR's for now. 

Again many thanks for your perseverance.

ablaom:
Okay, I have made that fix and rebased `for-a-0-point-20-release`. It looks like this created no merge conflicts for you, so no action on your part should be needed.

PR #735, opened 2022-02-02
Fix CategoricalDistributions.jl reexports

davnn:
Additionally requires https://github.com/JuliaAI/CategoricalDistributions.jl/pull/24


ablaom (approved):
Thanks, @davnn.

PR #738, opened 2022-02-07
Added Kappa Measure

Rahulub3r:
I am new to contributing to open source repositories. Please let me know if there is something else that I need to do or if there is scope for improvement. Thanks.

### DIFF
--- a/src/measures/finite.jl
+++ b/src/measures/finite.jl
@@ -143,7 +143,9 @@ function (::Kappa)(cm::ConfusionMatrixObject{C}) where C
     p₀ = sum(diag(cm.mat))/sum(cm.mat)
 
     # probability of agreement due to chance - for each class cᵢ, this would be: (#predicted=cᵢ)/(#instances) x (#observed=cᵢ)/(#instances)
-    pₑ = sum(sum(cm[j, :]) * sum(cm[:, j]) for j in 1:C)/sum(cm.mat)^2
+    rows_sum = sum!(similar(cm.mat, 1, C), cm.mat) # 1 x C matrix
+    cols_sum = sum!(similar(cm.mat, C, 1), cm.mat) # C X 1 matrix
+    pₑ = first(rows_sum*cols_sum)/sum(rows_sum)^2
 
     # Kappa calculation
     κ = (p₀ - pₑ)/(1 - pₑ)
--- a/src/measures/finite.jl
+++ b/src/measures/finite.jl
@@ -113,6 +113,46 @@ function call(m::BACC, ŷm, ym, wm::Union{Nothing,Arr{<:Real}}=nothing)
     return score
 end
 
+# ---------------------------------------------------
+# kappa
+
+struct Kappa <: Aggregated end
+
+metadata_measure(Kappa;
+                 instances  = ["kappa"],
+                 target_scitype           = FiniteArrMissing,
+                 prediction_type          = :deterministic,
+                 orientation              = :score,
+                 supports_weights         = false)
+
+@create_aliases Kappa
+
+@create_docs(Kappa,
+body=
+"""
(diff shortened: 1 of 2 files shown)


OkonSamuel:
Thanks @Rahulub3r. Thanks for the contribution. Could you please add some tests.

ablaom:
@Rahulub3r The tests live in the test/ folder in a file of the same name as you edited in src/

Rahulub3r:
> @Rahulub3r The tests live in the test/ folder in a file of the same name as you edited in src/

@ablaom @OkonSamuel Added the test cases. Let me know if I need to do anything else.

ablaom (commented):
Code looks good to me. Just waiting on CI. 

@OkonSamuel You happy with this?

src/measures/finite.jl
```diff
@@ -113,6 +113,46 @@ function call(m::BACC, ŷm, ym, wm::Union{Nothing,Arr{<:Real}}=nothing)
...
+""",
+scitype=DOC_FINITE)
+
+# calling behaviour:
+function (::Kappa)(cm::ConfusionMatrixObject{C}) where C
+    # relative observed agreement - same as accuracy
+    p₀ = sum(diag(cm.mat))/sum(cm.mat)
+
+    # probability of agreement due to chance - for each class cᵢ, this would be: (#predicted=cᵢ)/(#instances) x (#observed=cᵢ)/(#instances)
+    pₑ = sum(sum(cm[j, :]) * sum(cm[:, j]) for j in 1:C)/sum(cm.mat)^2
```
OkonSamuel, line 146: The following should be more efficient, even in the case where C>>2. This take advantage of the fact that `ConfusionMatrixObject` wraps a square matrix of type `Matrix{Int}`.
```suggestion
    rows_sum = sum!(similar(cm.mat, 1, C), cm.mat) # 1 x C matrix
    cols_sum = sum!(similar(cm.mat, C, 1), cm.mat) # C X 1 matrix
    pₑ = first(rows_sum*cols_sum)/sum(rows_sum)^2
```

OkonSamuel (approved):
Aside from some minor technical details (which may be overlooked for now), this PR LGTM.

src/measures/finite.jl
Rahulub3r: Makes sense. How about we have something like this (a little more simpler) ?

```
rows_sum = sum(cm.mat, dims=2)
cols_sum  = sum(cm.mat, dims=1)
pₑ = first(cols_sum*rows_sum)/sum(rows_sum)^2
```

OkonSamuel: > rows_sum = sum(cm.mat, dims=2)
> cols_sum  = sum(cm.mat, dims=1)
> pₑ = first(cols_sum*rows_sum)/sum(rows_sum)^2

Yeah, that also works, but the only issue here is that `rows_sum = sum(cm.mat, dims=2)` isn't as efficient as the original implementation. But as the C grows in size this is still quite efficient. You may use this option if you want.
It's up to you.

OkonSamuel:
Thanks @Rahulub3r

ablaom:
@Rahulub3r Just wanted to thank you for this work. I appreciate it takes some work to wrap your head around the interface and if you are also new to git workflow, that only makes the path to a successful PR steeper. 

Great to have your contribution. 

I you are interested and available to help out further, there is a lot to be done in the measures area, in particular just adding missing measures.  See [this filtered project list](https://github.com/orgs/JuliaAI/projects/1?card_filter_query=label%3Ameasures) and in particular [this list](https://github.com/JuliaAI/MLJBase.jl/issues/95) (which needs updating).

Rahulub3r:
@ablaom You're welcome. I will look into the list and see how I can contribute.

PR #740, opened 2022-02-21
Fix `iteration_parameter` for pipelines in case supervised component has value `nothing`

ablaom:
Currently, if `iteration_parameter(supervised_model) == nothing` for a supervised component in a `Pipeline`, `pipe`, then `iteration_parameter(pipe)` throws an error. 

This PR fixes this (and addresses some dep warnings from test pkg MultivariateStats.jl).

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -47,22 +47,26 @@ function reduce_nested_field(ex)
 end
 
 """
-    prepend(s::Symbol, s::Symbol)
+    MLJBase.prepend(::Symbol, ::Union{Symbol,Expr,Nothing})
 
 For prepending symbols in expressions like `:(y.w)` and `:(x1.x2.x3)`.
 
-julia> compose(:x, :y)
+julia> prepend(:x, :y)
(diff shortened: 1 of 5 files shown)


src/utilities.jl
```diff
@@ -47,23 +47,27 @@ function reduce_nested_field(ex)
...
 :(x.y.z)
 
-julia> compose(:w, ans)
+julia> prepend(:w, ans)
 :(w.x.y.z)
 
+If the second argument is `nothing`, then `nothing` is returned.
+
 """
+prepend(s::Symbol, ::Nothing) = nothing
```
ablaom, line 66: Here is the fix.

ablaom:
@OkonSamuel Sorry for the cluttered PR. The essential additions are the line marked above and the new tests.

ablaom:
@OkonSamuel Sorry for the cluttered PR - The only change of substance is the fixed annotated above.

OkonSamuel (approved):
LGTM

PR #745, opened 2022-03-07
make the stack's resampling scheme more general

olivierlabayle:
Unless I am missing something I think the existing version of `getfolds` is restrictive. 
This is a simple extension of the capabilities of resampling schemes. 

@edit 
I am converting to draft as I am also trying to get rid of the assumption that the resampling method needs a `nfolds` field and rely only on `train_test_pairs`.

### DIFF
--- a/src/composition/models/stacking.jl
+++ b/src/composition/models/stacking.jl
@@ -248,16 +248,8 @@ MLJBase.package_license(::Type{<:Stack}) = "MIT"
 ################# Node operations Methods #################
 ###########################################################
 
-
-getfolds(X::AbstractNode, y::AbstractNode, cv, n::Int) =
-    node((XX, YY) -> train_test_pairs(cv, 1:n, XX, YY), X, y)
-
-trainrows(X::AbstractNode, folds::AbstractNode, nfold) =
-    node((XX, ff) -> selectrows(XX, ff[nfold][1]), X, folds)
-
-testrows(X::AbstractNode, folds::AbstractNode, nfold) =
-    node((XX, ff) -> selectrows(XX, ff[nfold][2]), X, folds)
-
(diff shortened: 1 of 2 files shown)


ablaom:
Happy to review when you're ready.

olivierlabayle:
> Happy to review when you're ready.

Thanks, I think this is ready for review!

src/composition/models/stacking.jl
```diff
@@ -292,10 +278,10 @@ When measure/measures is provided, the folds_evaluation will have been filled by
 not doing any heavy work (not constructing nodes corresponding to measures) but just unpacking all the folds_evaluations in a single node that
 can be evaluated later.
 """
-function internal_stack_report(m::Stack, verbosity::Int, y::AbstractNode, folds_evaluations::Vararg{AbstractNode})
-    _internal_stack_report(y, folds_evaluations...) =
-        internal_stack_report(m, verbosity, y, folds_evaluations...)
-    return (report=(cv_report=node(_internal_stack_report, y, folds_evaluations...),),)
+function internal_stack_report(m::Stack, verbosity::Int, ttp, folds_evaluations::Vararg{AbstractNode})
```
ablaom, line 281: So the name `ttp` caused me some head scratching. Can we please have a more informative name? If `train_test_pairs` is really too long, how about `tt_pairs`?

ablaom (commented):
This looks like a nice refactor. You'll want to update the docstring at line 94.

src/composition/models/stacking.jl
olivierlabayle: Yes sure, no problem!

ablaom (approved):
Looks good to me, thanks @olivierlabayle !

PR #747, opened 2022-03-14
When constructing a `TransformedTargetModel`, allow the user to override the `target_scitype`, `predict_scitype`, and `prediction_type` traits

DilumAluthge:
For example, this will allow the user to take a `Deterministic` regression model and transform it into a `Probabilistic` binary classification model.

TODO before merging:
- [ ] Remove the `evotrees.jl` tests.
- [ ] Add some simpler tests.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -50,7 +50,10 @@ julia = "1.6"
 [extras]
 DecisionTree = "7806a523-6efd-50cb-b5f6-3fa6f1930dbb"
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
+EvoTrees = "f6006082-12f8-11e9-0c9c-0d5d367ab1e5"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
+MLJ = "add582a8-e3ab-11e8-2d5e-e98b27df1bc7"
+MLJModels = "d491faf4-2d78-11e9-2867-c94bc002c0b7"
 MultivariateStats = "6f286f6a-111f-5878-ab1e-185364afe411"
 NearestNeighbors = "b8a86587-4115-5ab1-83bc-aa920d37bbce"
 StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
@@ -58,4 +61,4 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 TypedTables = "9d95f2ec-7b3d-5a63-8d20-e2491e220bb9"
 
 [targets]
-test = ["DecisionTree", "Distances", "Logging", "MultivariateStats", "NearestNeighbors", "StableRNGs", "Test", "TypedTables"]
+test = ["DecisionTree", "Distances", "EvoTrees", "Logging", "MLJ", "MLJModels", "MultivariateStats", "NearestNeighbors", "StableRNGs", "Test", "TypedTables"]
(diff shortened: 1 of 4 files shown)


DilumAluthge:
@ablaom @OkonSamuel This is now ready to review.

Instead of specifying the abstract type, the user instead specifies some traits, and we infer the appropriate abstract type based on those traits. Specifically, the user can provide the following three traits:
1. `target_scitype`
2. `predict_scitype`
3. `prediction_type`

Right now, I'm testing this PR by using an example drawn from EvoTrees.jl, which was the original motivation for this PR. I'm fine with keeping that test, but I realize that EvoTrees.jl is a heavy dependency, so I would also be fine with removing the EvoTrees-based tests and instead putting in some more lightweight tests.

Also, right now, I have MLJ.jl as a test dependency; obviously we want to get rid of that.

ablaom (commented):
Thanks @DilumAluthge  for spending time with `TransformedTargetModel`. There is one problem, for which you are not really to blame, which involves an issue we have resolved in other wrappers, but not here, which has led you to be misled. 

 In your PR you make `target_scitype` a type parameter of `TransformedModel`, but in fact the `target_scitpye` of this wrapper (and other model wrappers like, say, [`IteratedModel`](https://github.com/JuliaAI/MLJIteration.jl/blob/7d063f9c6eb1c8a8956426bf522c0c1d5839aab3/src/traits.jl#L14)) cannot be a parameter, because in general it cannot be determined from the *type* of the atomic model `model`, only from a model *instance*, and the user is permitted to mutate the `model` in ways which change its `target_scitype`. An example is when the atomic `model` is a pipeline model. See [this comment](https://github.com/JuliaAI/MLJBase.jl/blob/94c3d10b72884551b72c2a3d1a7d037964fb651e/src/composition/models/pipelines.jl#L3). At present the only guarantee is that the `abstract_type` of `model` will not change. (Even this might change in the longer term if we discard the abstract type hierarchy).

Traits were initially defined at the level of types, because this was very natural from the point of view of the model registry. So, in fact, every model in the registry has a `target_scitype` that is known from its type. However,  composite models complicate the story: one may want (in "architecture search", say) to mutate model components in ways which change the `target_scitype` (or other traits) (this could be as mild as `OrderedFactor{2}` -> `OrderedFactor`). So, moving forward, outside of the registry, traits should only be called on instances, and only implemented for types when safe to do so (a fallback for every trait gets the instance trait from the type trait).

Perhaps you have a better idea, but I cannot help but feel we are working very hard to resolve an issue here than will disappear in a future no-abstract-type refactoring of MLJ.  But it's also not clear how much support there is for that refactoring, which itself will be the most disruptive so far. 

Perhaps we can keep he type parameters you have added here but have them fallback to `Nothing` when the user does not specify the `traits` dictionary. Then the traits for the wrapped model are determined by their values on instances for the atomic model if the type parameters are `Nothing` and otherwise get them from the type parameters. But this seems very hacky, and difficult to understand/maintain later.

I will today have a look at a workaround for the original issue you raised on Slack -  it seems to me that `check_measures=false` is not toggling all checks off, as it should. And will cross-reference here for you to consider as an alternative resolution to your problem.

ablaom:
https://github.com/JuliaAI/MLJBase.jl/pull/748

DilumAluthge:
Closing in favor of https://github.com/JuliaAI/MLJBase.jl/pull/748 and https://github.com/JuliaAI/MLJBase.jl/issues/760

PR #749, opened 2022-03-23
Add a brief reminder that the fields of the `PerformanceEvaluation` struct are part of the public API

DilumAluthge:


ablaom (approved):
Thanks!

PR #753, opened 2022-03-29
Allow user to control the level of scitype checks in `machine` constructor

### LINKED ISSUE #908: Feature request: ability to convert scitype warnings into errors
Sometimes, when I do `mach = MLJ.machine(model, X, y)`, it prints one or more warnings of the form:

```
Warning: the scitype of `X` is incompatible with...
Warning: the scitype of `y` is incompatible with...
```

Is there a way to convert those warnings to errors that are thrown immediately?

ablaom:
**edited** (can now modify the global default value of `scitype_check_level` kwarg of `machine` constructor)

Closes https://github.com/alan-turing-institute/MLJ.jl/issues/908 .

Although not breaking, this PR has `for_a-0-point-20-release` as target, not `dev`. This is because it is not urgent and would otherwise have led to messy merge conflicts later on. 

In this PR we:

- (**enhancement**) Allow user to specify `scitype_check_level` when constructing machines, as in `machine(model, X, y, check_scitype_level=2)`, to control how strictly the constructor enforces scitype type compatibility between `model` and the data. Allow user to change the global default `scitype_check_level` using new method `scitype_check_level(i::Int)`. 

From the new docstring:

```
machine(model, args...; cache=true, scitype_check_level=1)
```
...

If `scitype_check_level > 0` (default is `1`) then the scitype of each `arg`
in `args` is computed, and this is compared with the scitypes expected
by the model, unless `args` contains `Unknown` scitypes and
`check_level < 4`, in which case no further action is taken. Warnings
are issued or errors thrown as detailed in the following table:

`check_level` | Inspect scitypes? | If `Unknown` in scitypes | If other scitype mismatch |
|:-----------:|:-----------------:|:------------------------:|:-------------------------:|
0             | ×                 |                          |                           |
1             | ✓                 |                          | warning                   |
2             | ✓                 | warning                  | warning                   |
3             | ✓                 | warning                  | error                     |
4             | ✓                 | error                    | error                     |

### DIFF
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -94,6 +94,9 @@ using Statistics, LinearAlgebra, Random, InteractiveUtils
 # ===================================================================
 ## CONSTANTS
 
+const HANDLE_GIVEN_ID = Dict{UInt64,Symbol}()
+const SHOW_COLOR = true
+
 const PREDICT_OPERATIONS = (:predict,
                             :predict_mean,
                             :predict_mode,
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -94,8 +94,7 @@ using Statistics, LinearAlgebra, Random, InteractiveUtils
 # ===================================================================
 ## CONSTANTS
 
-const HANDLE_GIVEN_ID = Dict{UInt64,Symbol}()
-const SHOW_COLOR = true
+# for variable global constants, see src/init.jl
 
 const PREDICT_OPERATIONS = (:predict,
                             :predict_mean,
@@ -184,7 +183,7 @@ const LOSS_FUNCTIONS = vcat(MARGIN_LOSSES, DISTANCE_LOSSES)
 # default_resource allows to switch the mode of parallelization
 
 default_resource()    = DEFAULT_RESOURCE[]
-default_resource(res) = (DEFAULT_RESOURCE[] = res)
+default_resource(res) = (DEFAULT_RESOURCE[] = res;)
 
 # ===================================================================
 # Includes
@@ -300,7 +299,7 @@ export flat_values, recursive_setproperty!,
     recursive_getproperty, pretty, unwind
 
 # show.jl
-export HANDLE_GIVEN_ID, @more, @constant, @bind, color_on, color_off
+export HANDLE_GIVEN_ID, @more, @constant, color_on, color_off
 
 # datasets.jl:
 export load_boston, load_ames, load_iris, load_sunspots,
@@ -312,7 +311,7 @@ export load_boston, load_ames, load_iris, load_sunspots,
 export source, Source, CallableReturning
 
 # machines.jl:
-export machine, Machine, fit!, report, fit_only!
+export machine, Machine, fit!, report, fit_only!, default_scitype_check_level
 
 # datasets_synthetics.jl
 export make_blobs, make_moons, make_circles, make_regression
--- a/src/machines.jl
+++ b/src/machines.jl
@@ -106,9 +106,9 @@ err_length_mismatch(model) = DimensionMismatch(
     "Differing number of observations "*
     "in input and target. ")
 
-check(model::Any, args...; kwargs...) =
+check(model::Any, args...) =
     throw(ArgumentError("Expected a `Model` instance, got $model. "))
-function check(model::Model, args...; full=false)
+function check(model::Model, args...)
     nowarns = true
 
     F = fit_data_scitype(model)
@@ -306,7 +306,7 @@ machine(model::Model, arg1::AbstractNode, arg2, args...; kwargs...) =
 
 function machine(model::Model, raw_arg1, raw_args...; kwargs...)
     args = source.((raw_arg1, raw_args...))
-    check(model, args...; full=true)
+    check(model, args...)
     return Machine(model, args...; kwargs...)
 end
 
@@ -555,7 +555,7 @@ function fit_only!(mach::Machine{<:Model,cache_data};
                     @warn "Some learning network source nodes are empty. "
                 @info "Running type checks... "
                 raw_args = map(N -> N(), mach.args)
-                if check(mach.model, source.(raw_args)... ; full=true)
+                if check(mach.model, source.(raw_args)...)
                     @info "Type checks okay. "
                 else
                     @info "It seems an upstream node in a learning "*
--- a/src/machines.jl
+++ b/src/machines.jl
@@ -14,12 +14,12 @@ The effect of the `scitype_check_level` option in calls of the form
 `machine(model, data, scitype_check_level=...)` is summarized below:
 
 `scitype_check_level` | Inspect scitypes? | If `Unknown` in scitypes | If other scitype mismatch |
-|:-----------:|:-----------------:|:------------------------:|:-------------------------:|
-0             | ×                 |                          |                           |
-1 (startup)   | ✓                 |                          | warning                   |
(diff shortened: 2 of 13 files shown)


DilumAluthge:
>  Base automatically changed from `for-a-0-point-20-release` to `dev` 9 days ago 

@ablaom Just wanted to make sure that you noticed that GitHub changed the base (target) branch of this PR.

ablaom:
> @ablaom Just wanted to make sure that you noticed that GitHub changed the base (target) branch of this PR.

Yes, I did notice but thanks! Will resolve new conflicts after easter.

ablaom:
@DilumAluthge I've resolved conflicts. Assuming tests pass, this is ready for your review.

ablaom:
@DilumAluthge Do you have any time to review this PR over the next week?

src/MLJBase.jl
```diff
@@ -297,7 +299,7 @@ export flat_values, recursive_setproperty!,
     recursive_getproperty, pretty, unwind
 
 # show.jl
-export HANDLE_GIVEN_ID, @more, @constant, @bind, color_on, color_off
+export HANDLE_GIVEN_ID, @more, @constant, color_on, color_off
```
DilumAluthge, line 302: Why was this export removed?

Also, is that a breaking change?

src/machines.jl
```diff
@@ -1,3 +1,34 @@
+## SCITYPE CHECK LEVEL
+
+"""
+   default_scitype_check_level()
+
+Return the current global default value for scientific type checking
+when constructing machines.
+
+   default_scitype_check_level(i)
```
DilumAluthge, line 9:
```suggestion
   default_scitype_check_level(i::Integer)
```

src/machines.jl
```diff
@@ -1,3 +1,34 @@
...
+The effect of the `scitype_check_level` option in calls of the form
+`machine(model, data, scitype_check_level=...)` is summarized below:
+
+`scitype_check_level` | Inspect scitypes? | If `Unknown` in scitypes | If other scitype mismatch |
+|:-----------:|:-----------------:|:------------------------:|:-------------------------:|
+0             | ×                 |                          |                           |
+1 (startup)   | ✓                 |                          | warning                   |
+2             | ✓                 | warning                  | warning                   |
+3             | ✓                 | warning                  | error                     |
+4             | ✓                 | error                    | error                     |
```
DilumAluthge, line 22: It would be nice to fix the whitespace so that this table appears aligned when you look at the source code.

src/machines.jl
```diff
@@ -161,12 +206,24 @@ mach = machine(model, X, y)
...
-to guarantee data anonymity when serializing composite models.
+Specify `cache=false` to prioritize memory management over speed.
 
 When building a learning network, `Node` objects can be substituted
-for the concrete data.
+for the concrete data but no type or dimension checks are applied.
+
+### Checks on the types of training data
+
+A model articulates it's data requirements using [scientific
```
DilumAluthge, line 216:
```suggestion
A model articulates its data requirements using [scientific
```

src/machines.jl
```diff
@@ -161,12 +206,24 @@ mach = machine(model, X, y)
...
+types](https://juliaai.github.io/ScientificTypes.jl/dev/), i.e.,
+using the [`scitype`](@ref) function instead of the `typeof` function.
+
+If `scitype_check_level > 0` then the scitype of each `arg` in `args`
+is computed, and this is compared with the scitypes expected by the
+model, unless `args` contains `Unknown` scitypes and
+`scitype_check_level < 4`, in which case no further action is
+taken. Whether warnings are issued or errors thrown depends the
+level. The method `default_scitype_check_level`](@ref) controls the
+default level (`1` at startup).
```
DilumAluthge, line 226: After this paragraph, I think it would be helpful to insert a copy of the table from above that shows the behavior at each value of `scitype_check_level`.

Probably the easiest approach would be to define the table in a global constant string, and then interpolate that constant into each location that we want the table to appear. Then, we only have to maintain one version of the table, but we can use it multiple times.

src/machines.jl
```diff
@@ -769,7 +832,11 @@ Returns a shallow copy of the machine to make it serializable. In particular,
 all training data is removed and, if necessary, learned parameters are replaced
 with persistent representations.
 
+<<<<<<< HEAD
+Any general purpose Julia serialization may be applied to the output of
+=======
 Any general purpose Julia serializer may be applied to the output of
+>>>>>>> dev
```
DilumAluthge, line 839: It looks like some Git merge conflict markers accidentally got committed.

DilumAluthge:
I'm seeing a lot of color related changes in this PR. Are those intentional, or did something go wrong during a rebase?

DilumAluthge (approved):

DilumAluthge:
> Addresses [alan-turing-institute/MLJ.jl#908](https://github.com/alan-turing-institute/MLJ.jl/issues/908) .

I would say this closes #908, right?

src/MLJBase.jl
ablaom: No @bind is some old experimental thing that was never public or documented. I think the code defining it is gone.

Not breaking.

src/machines.jl
ablaom: Mmm. I was just thinking this doc string is already mammoth, and that's why I sucked this out. How about I change on line 225 onwards to:
 
```
level. For details, see `default_scitype_check_level`](@ref), which controls the
default level (`1` at startup).
```

src/machines.jl
ablaom: fixed

src/machines.jl
ablaom: done

ablaom:
Yes.

ablaom:
Sorry, what do you mean by "color" changes?

src/machines.jl
DilumAluthge: Hmmm, that's a fair point. Okay, let's go with a `@ref` for now.

src/sources.jl
```diff
@@ -129,7 +129,7 @@ function Base.show(stream::IO, object::AbstractNode)
     str = simple_repr(typeof(object))
     show_handle(object) && (str *= " $(handle(object))")
     if false
-        printstyled(IOContext(stream, :color=> SHOW_COLOR),
+        printstyled(IOContext(stream, :color=> SHOW_COLOR[]),
```
DilumAluthge, line 132: @ablaom Here's an example of the "color" changes that appear in this PR but seem unrelated to the PR.

ablaom: Oh right. No, that is intentional. It brings `SHOW_COLOR` into line with `DEFAULT_SCITYPE_CHECK_LEVEL` (introduced in this PR) and `DEFAULT_RESOURCE`.  See [here](https://github.com/JuliaAI/MLJBase.jl/blob/430eff5cc05c88e07bf16aba70b8b3b8696b6338/src/init.jl#L5). We are now now using `Ref` to avoid redefining global constants.

DilumAluthge: Ah, yeah a constant `Ref` sounds better than redefining globals.

DilumAluthge (approved):
Overall, the feature looks good!

After you have implemented your suggestion in https://github.com/JuliaAI/MLJBase.jl/pull/753#discussion_r863303906, this is good to go from my point of view.

ablaom:
Thanks @DilumAluthge for reviewing this PR and raising the original issue.

PR #758, opened 2022-04-12
Fix wrong field name in a `show` method

rikhuijzer:
According to the `show` output, `PerformanceEvaluation` has a field `train_test_pairs` whereas this should actually be `train_test_rows`.

Should I also add a test to check whether all `fieldnames(PerformanceEvaluation)` are contained in the output of `show`?

### DIFF
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -538,7 +538,7 @@ function Base.show(io::IO, ::MIME"text/plain", e::PerformanceEvaluation)
             "with these fields:")
     println(io, "  measure, measurement, operation, per_fold,\n"*
             "  per_observation, fitted_params_per_fold,\n"*
-            "  report_per_fold, train_test_pairs")
+            "  report_per_fold, train_test_rows")
     println(io, "Extract:")
     show_color = MLJBase.SHOW_COLOR
     color_off()
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -503,9 +503,8 @@ struct.
   machine `mach` training in resampling - one machine per train/test
   pair.
 
-- `train_test_rows`: a vector of vectors containing the train-test
-  pairs used to evaluate the model.
-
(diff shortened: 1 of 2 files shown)


ablaom:
Good catch @rikhuijzer . 

I also notice this field is missing from the [doc string](https://github.com/JuliaAI/MLJBase.jl/blob/95efa58e9e7a9c8818ebeb917323e744c8606ae8/src/resampling.jl#L505). Do you mind adding that also?

And yeah, a test would be a bonus.

rikhuijzer:
> Good catch @rikhuijzer .
> 
> I also notice this field is missing from the [doc string](https://github.com/JuliaAI/MLJBase.jl/blob/95efa58e9e7a9c8818ebeb917323e744c8606ae8/src/resampling.jl#L505). Do you mind adding that also?
> 
> And yeah, a test would be a bonus.

Nice catch on the missing field in the docstring. I've added that and tests now.

I'm not super happy with the tests since testing whether the docstring matches the fieldnames seems like a common problem which shouldn't require hand-crafted code. I know that Fons has manually created a macro to allow writing docs inline (https://github.com/JuliaPluto/PlutoSliderServer.jl/blob/abad2ae5164c98dab1db6a7b2e653d3c87dd031b/src/Configuration.jl#L11-L24). Anyway, tests are easy to refactor so maybe we should leave this problem for later.

src/resampling.jl
```diff
@@ -503,6 +503,9 @@ struct.
   machine `mach` training in resampling - one machine per train/test
   pair.
 
+- `train_test_rows`: a vector of vectors containing the train-test
+  pairs used to evaluate the model.
+
```
ablaom, line 508:
```suggestion
- `train_test_rows`: a vector of tuples, each of the form `(train, test)`, where `train` and `test` 
   are vectors of row (observation) indices for training and evaluation respectively. 
```

ablaom (approved):
Many thanks @rikhuijzer.

PR #762, opened 2022-05-02
Add interface point for specifying `acceleration` in a composite models

ablaom:
This PR:

-  introduces a mechanism for implementing accelerated training of composite models that have been defined by "manually" exporting a learning network
- implements the mechanism for `Stack`s
 
An alternative approach with a different interface point for acceleration is provided by #759 . 

Supposing the composite model struct has `acceleration::AbstractResource` as a hyper-parameter, one simply adds `acceleration=acceleration` in the learning network machine constructor. Here's an example:

```julia
mutable struct MyAverageTwo <: DeterministicComposite
    rgs1 # deterministic regressor
    rgs2 # deterministic regressor
    acceleration::AbstractResource
end

function MLJModelInterface.fit(model::MyAverageTwo, verbosity, X, y)

    X = source(X)
    y = source(y)

    m1 = machine(model.rgs1, X, y)
    y1 = predict(m1, X)

    m2 = machine(model.rgs2, X, y)
    y2 = predict(m2, X)

    yhat = 0.5*y1 + 0.5*y2

    # learning network machine
    mach = machine(Deterministic(), X, y; predict=yhat, acceleration=model.acceleration)

    return!(mach, model, verbosity)

end
```

To support resources other than `CPU1()`, requires  new overloadings of  `fit!(::Node, acceleration; kwargs)` as shown [here](https://github.com/JuliaAI/MLJBase.jl/blob/fe637029266ebaf3c351127a4678c0aefe47a09a/src/composition/learning_networks/nodes.jl#L213) for `CPUThreads`.   In a separate [draft branch](https://github.com/JuliaAI/MLJBase.jl/tree/acceleration-with-composites-and-multithreading), I have done just that, and successfully tested a `Stack`.

TODO:

- [ ] Update documentation (for `fit!(::Node)` and `fit!(:Machine{<:Surrogate}` and `fit!(::Machine)`.
- [ ] Add test that  `acceleration=CPUThreads()` in `Stack` throws expected unsupported acceleration error. 
- [ ] See if integration into @from_network is possible

### DIFF
--- a/src/composition/learning_networks/machines.jl
+++ b/src/composition/learning_networks/machines.jl
@@ -35,6 +35,7 @@ function _operation_part(signature)
     ops = filter(in(OPERATIONS), keys(signature))
     return NamedTuple{ops}(map(op->getproperty(signature, op), ops))
 end
+
 function _report_part(signature)
     :report in keys(signature) || return NamedTuple()
     return signature.report
@@ -164,7 +165,7 @@ function check_surrogate_machine(::Union{Unsupervised},
     return nothing
 end
 
-function machine(model::Surrogate, _sources::Source...; pair_itr...)
+function machine(model::Surrogate, _sources::Source...; acceleration=CPU1(), pair_itr...)
 
     # named tuple, such as `(predict=yhat, transform=W)`:
     signature = (; pair_itr...)
@@ -185,7 +186,7 @@ function machine(model::Surrogate, _sources::Source...; pair_itr...)
 
     check_surrogate_machine(model, signature, _sources)
 
-    mach = Machine(model, _sources...)
+    mach = Machine(model, _sources...; acceleration=acceleration)
 
     mach.fitresult = CompositeFitresult(signature)
 
@@ -211,6 +212,7 @@ function machine(_sources::Source...; pair_itr...)
     return machine(model, _sources...; pair_itr...)
 
 end
+
 
 """
     N = glb(mach::Machine{<:Union{Composite,Surrogate}})
--- a/src/composition/learning_networks/machines.jl
+++ b/src/composition/learning_networks/machines.jl
@@ -248,7 +248,7 @@ See also [`machine`](@ref)
 """
 function fit!(mach::Machine{<:Surrogate}; kwargs...)
     glb_node = glb(mach)
-    fit!(glb_node; kwargs...)
+    fit!(glb_node; acceleration=mach.acceleration, kwargs...)
     mach.state += 1
     report_additions_ = _call(_report_part(signature(mach.fitresult)))
     mach.report = merge(report(glb_node), report_additions_)
--- a/src/composition/models/stacking.jl
+++ b/src/composition/models/stacking.jl
@@ -26,27 +26,64 @@ function input_target_scitypes(models, metalearner)
 end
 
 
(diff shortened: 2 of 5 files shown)


ablaom:
@olivierlabayle

olivierlabayle:
@ablaom Thanks, I've had a look at the PR, it's a bit different from what I expected. I originaly thought you would just add it as a model hyperparameter and pass the `acceleration` to the `return!` function. I am not sure I understand why acceleration is now a field of both `Machine` and `CompositeModel` types. 

What would be the use case for a non-composite model? it seems to create multiple entrypoints for acceleration which I find a bit confusing.

ablaom:
> it seems to create multiple entrypoints for acceleration

Not really, but I understand the confusion. I am frequently confusing the machine that wraps a composite model and concrete data (which has no "downstream" machines, because it's arguments are just `Source`s) with the machines in a the learning network that `fit!` generates under the hood (which generally does). Perhaps that's the source of confusion.

The main reason for adding `acceleration` as a field to `Machine` is as a way to implement get parallelism happening in models defined as shown above. The key modification is the re-implementation of `fit!(Machine{<:Surrogate})` [here](https://github.com/JuliaAI/MLJBase.jl/blob/4b8c65bf4fbf2b69131a995c2c18be6b2e82787e/src/composition/learning_networks/machines.jl#L251). I have made an analogous modification [here](https://github.com/JuliaAI/MLJBase.jl/blob/4b8c65bf4fbf2b69131a995c2c18be6b2e82787e/src/machines.jl#L625) but that is not necessary to get the above behaviour, nor does it conflict with it. I added it because it seemed natural to do so, and means I can fit a machine within a learning network (an ordinary machine with `Node` arguments, not a learning network machine) and get parallelism by specifying `fit!(mach, acceleration=...)`. I don't think this is an important use-case, because we're mainly interested in *exporting* learning networks as stand-alone types, and in that case, only the learning network machine is `fit!`. So I'm open to just dumping [this](https://github.com/JuliaAI/MLJBase.jl/blob/4b8c65bf4fbf2b69131a995c2c18be6b2e82787e/src/composition/learning_networks/machines.jl#L251) or at least not documenting the possibilities it affords, if you think it's confusing.

ablaom:
>  I originaly thought you would just add it as a model hyperparameter 

It is added as a hyperparameter. The parameter could be put in `return!` instead of the `Surrogate` machine. I hadn't consider that. Is there an advantage?

olivierlabayle:
> > I originaly thought you would just add it as a model hyperparameter
> 
> It is added as a hyperparameter. The parameter could be passed on using `return!` instead of the `Surrogate` machine constructor, I hadn't looked so see if that works. Is there an advantage?

I think the advantage is that you don't have to overburden the Machine with a new field that is only relevant for the Surrogate which is itself an artefact of Composition. I am quite confident it would work too since I have made similar thing [here](https://github.com/JuliaAI/MLJBase.jl/pull/759/commits/38a576fbe908b7a90942bcf83baa257c2e3ec4de) to prioritize the Composite's acceleration specification.

ablaom:
I agree, that having that the `acceleration` interface point in `return!` is preferable to the implementation in this PR. It's simpler, and the extra generality afforded by this PR is not worth the possibility confusion. It will also be a little easier to integrate the change in `@from_network`.  Let's continue our discussion of how to move forward at #759 .

PR #767, opened 2022-05-12
Stack cache and acceleration

olivierlabayle:
Following up on: https://github.com/JuliaAI/MLJBase.jl/pull/759

This adds two fields to the Stack to enable/disable submachines caching and the network's acceleration:

- `cache`: Whether submachines will be caching data or not.
- `acceleration`: Acceleration mode of the learning network.

**added by** @ablaom: This PR allows one to train a learning network node using multithreading, as in `fit!(node, acceleration=CPUThreads())`.


test/Project.toml
```diff
@@ -0,0 +1,25 @@
...
+ScientificTypes = "321657f4-b219-11e9-178b-2701a2544e81"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
+LossFunctions = "30fc2ffe-d236-52d8-8643-a9d8f7c094a7"
+ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
+StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
+Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
+Parameters = "d96e819e-fc66-5662-9728-84c9c7592b0a"
```
ablaom, line 25: I think you have forgotten to remove the corresponding entries from the core Project.toml: 
https://github.com/JuliaAI/MLJBase.jl/blob/37a5256fa85eec02d91e340b0f4c0a1ddcbbf44a/Project.toml#L51

src/composition/models/stacking.jl
```diff
@@ -95,6 +99,11 @@ When training a machine bound to such an instance:
   evaluation of the learners in the Stack while training. This is not for the
   evaluation of the Stack itself.
 
+- `cache`: Weither machines created in the learning network will cache data or not.
```
ablaom, line 102:
```suggestion
- `cache`: Whether machines created in the learning network will cache data or not.
```

src/composition/models/stacking.jl
```diff
@@ -95,6 +99,11 @@ When training a machine bound to such an instance:
...
+
+- `acceleration`: A supported `AbstractResource` to define the training parallelization
+  mode of the learning network.
```
ablaom, line 105: Some users will use `Stack` but not be familiar with the term "learning network", so I suggest dropping the reference here. If they try an unsupported resource, they get an informative message explaining that. 

```suggestion
- `acceleration`: an `AbstractResource` defining the mode of parallelization for training the stack
```

ablaom: But we should also update this line: https://github.com/JuliaAI/MLJBase.jl/blob/37a5256fa85eec02d91e340b0f4c0a1ddcbbf44a/src/composition/learning_networks/nodes.jl#L210

to include `CPUThreads()` as a valid resource.

ablaom:
@olivierlabayle This is great, thank you. 

Since this adds `CPUThreads` as an option for training *any* learning network node, it's probably prudent to add some more tests for this. I would like to see that the `CPUThreads` analogue of these tests also a pass:

https://github.com/JuliaAI/MLJBase.jl/blob/4d1ed14e7b1034fbe00c4dde3550144180da5225/test/composition/learning_networks/machines.jl#L194

Note that `@test_mach_sequence` already provides a list of multiple acceptable training sequences, to handle case of asynchronous training. So they "should" pass. 

Probably we can just wrap these tests in an `@test_accelerated` block, as we have elsewhere, but exclude ` CPUProcesses`. 

I don't think we need to check both options give same results. You already do this for `Stack`. What do you think?

ablaom (changes requested):
My comments and test suggestions not withstanding, I'm already happy with this "draft" PR.

test/Project.toml
olivierlabayle: Exact! 👍

olivierlabayle:
@ablaom Thank you for this early review. I had indeed only marked this as a draft because I wanted to be a bit more thorough about the testing of the new multithreading feature. 

I will need to have a proper look at the `@test_accelerated` and `@test_mach_sequence` macros that you suggest. If the models in the sequence are completely deterministic I think I should make sure the results are concordent at least to some extent.

olivierlabayle:
I have incremented the testset you suggested with `CPUThreads()` as a resource for the custom learning network. Unfortunately, I couldn't make use of the `@testset_accelerated` macro with the `exclude` keyword. Besides, what is the advantage over the classic `@testset for accel in accelerations end`? (This is what I have used)

ablaom:
> Besides, what is the advantage over the classic @testset for accel in accelerations end? (This is what I have used)

👍🏾 Probably that macro was written before that syntax was available?? I didn't know about it myself.

ablaom (approved):
I'm happy with this, @olivierlabayle

I'm going to recommend that @OkonSamuel also reviews this, as he is more familiar with multithreading issues. He is pretty busy, but I think it's worth getting an expert to look over this.

ablaom:
Following suggestion of @OkonSamuel I thing we should test multi-threading works for a variety of models. I am working on some enhancements of [MLJTestIntegration.jl](https://github.com/JuliaAI/MLJTestIntegration.jl) to automate such a test, and will check back here when that is ready.

olivierlabayle:
> Following suggestion of @OkonSamuel we should test multi-threading works for a variety of component models. I am working on some enhancements of [MLJTestIntegration.jl](https://github.com/JuliaAI/MLJTestIntegration.jl) to automate such a test, and will check back here when that is ready.

Great idea, I'll be waiting then!

src/composition/models/stacking.jl
```diff
@@ -209,6 +218,8 @@ function Base.getproperty(stack::Stack{modelnames}, name::Symbol) where modelnam
     name === :metalearner && return getfield(stack, :metalearner)
     name === :resampling && return getfield(stack, :resampling)
     name == :measures && return getfield(stack, :measures)
+    name === :cache && return getfield(stack, :cache)
```
ablaom, line 221: You need to add `:acceleration` and `:cache` to the list of `propertynames` in line 214

olivierlabayle: done

src/composition/models/stacking.jl
```diff
@@ -54,7 +58,7 @@ const Stack{modelnames, inp_scitype, tg_scitype} =
             ProbabilisticStack{modelnames, inp_scitype, tg_scitype}}
 
 """
-    Stack(;metalearner=nothing, resampling=CV(), name1=model1, name2=model2, ...)
+    Stack(;metalearner=nothing, resampling=CV(), name1=model1, cache=true, acceleration=CPU1(), name2=model2, ...)
```
ablaom, line 61: This doc-string got a little mangled, with the model1 and model2 being separated. As this has grown passed the length of the line, how about we not list all the options here:

```suggestion
    Stack(;metalearner=nothing, name1=model1, name2=model2, keyword_options...)
```

ablaom:
@olivierlabayle I needed to rebase this and am closing in favour of https://github.com/JuliaAI/MLJBase.jl/pull/785. The rebase includes a fix for the doc string marked above.

PR #771, opened 2022-05-20
Fix typo

KronosTheLate:
I have not tested anything, but this was clearly a typo. Should doctests not have caught this?


ablaom (approved):
Thanks for this. "encoder" was the intended name for the field (although you can call it anything).

PR #773, opened 2022-05-23
Bump LossFunctions.jl

juliohm:
Appreciate if you can bump the version of LossFunctions.jl. It is holding downstream packages from updating.


juliohm:
Thank you @OkonSamuel appreciate if you can release a new patch release.

Issue #779, opened 2022-06-06
Add Multiple targets to `make_regression`


josephsdavid:
For a lot of documentation examples, it would be quite nice to have a `n_targets` field in the make_regression function!

ablaom:
Closed by https://github.com/JuliaAI/MLJBase.jl/pull/811

PR #780, opened 2022-06-06
add multiple targets to make_regression

josephsdavid:
I am still having a few issues here and there with this, but have been using something very similar in my examples. In service of #779.

### DIFF
--- a/src/data/datasets_synthetic.jl
+++ b/src/data/datasets_synthetic.jl
@@ -198,7 +198,7 @@ function make_circles(n::Integer=100;
         throw(ArgumentError(
             "Factor argument must be strictly between 0 and 1."))
     end
-    
+
     rng = init_rng(rng)
     # Generate points on a 2D circle
     θs = runif_ab(rng, n, 1, 0, 2pi)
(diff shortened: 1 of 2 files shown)


ablaom (commented):
Thanks for the contribution, @josephsdavid 

- If `as_table=true` (the default) then the multi-target array needs to get wrapped as a table, just like the features `X` (see the `finalize_Xy` method). But you will need to manually generate some dummy column names (eg, `:target1`, `target2`, and so forth) and do `MLJBase.table(y; names=names)`. Otherwise the automatically generated names for the target will conflict with those for the features (`:x1`, and so forth). 

- Please add a test. At the least, add a test that for `X, y = make_regression(n_targets=3)` we have `MLJBase.Tables.istable(y)` is true and check the number of rows and the column names of `y` with `MLJBase.nrows(y)` and `MLJBase.Tables.columnnames(y)` respectively. 

- Update the doc string

ablaom:
Closed in favour of PR cross-referenced below, where CI fail is addressed.

PR #786, opened 2022-06-10
closes #784 by fixing typename + tests

### LINKED ISSUE #784: typename method for AbstractDataFrame incorrect equality
https://discourse.julialang.org/t/simple-table-operation-has-very-large-compilation-time-with-mlj/82503/5?u=jack_n

tlienart:
* fixes `typename` for `DataFrames.DataFrame` object, tested with `1.0` to `1.7` seems ok
* added simple tests for it (required adding `DataFrames` to the `extras`)

**Note**: this `isdataframe` stuff should eventually be removed  via improvements to Tables.jl (see [here](https://github.com/JuliaData/Tables.jl/pull/278)) but for now is necessary to avoid big bottlenecks.

### DIFF
--- a/src/interface/data_utils.jl
+++ b/src/interface/data_utils.jl
@@ -105,7 +105,7 @@ function MMI.selectcols(::FI, ::Val{:table}, X, c::Union{Symbol, Integer})
     end
 end
 
-function MMI.selectcols(::FI, ::Val{:table}, X, c::AbstractArray)
+function MMI.selectcols(::FI, ::Val{:table}, X, c::Union{Colon, AbstractArray})
     if !isdataframe(X)
         cols = Tables.columntable(X) # named tuple of vectors
         newcols = project(cols, c)
--- a/src/interface/data_utils.jl
+++ b/src/interface/data_utils.jl
@@ -99,7 +99,7 @@ end
 function MMI.selectcols(::FI, ::Val{:table}, X, c::Union{Symbol, Integer})
     if !isdataframe(X)
         cols = Tables.columns(X)
-        return cols[c]
+        return Tables.getcolumn(cols, c)
     else
         return X[!, c]
     end
--- a/src/interface/data_utils.jl
+++ b/src/interface/data_utils.jl
@@ -97,14 +97,22 @@ function MMI.selectrows(::FI, ::Val{:table}, X, r)
 end
 
 function MMI.selectcols(::FI, ::Val{:table}, X, c::Union{Symbol, Integer})
-    cols = Tables.columntable(X) # named tuple of vectors
-    return cols[c]
+    if !isdataframe(X)
+        cols = Tables.columntable(X) # named tuple of vectors
+        return cols[c]
+    else
+        return X[!, c]
+    end
 end
 
 function MMI.selectcols(::FI, ::Val{:table}, X, c::AbstractArray)
-    cols = Tables.columntable(X) # named tuple of vectors
-    newcols = project(cols, c)
-    return Tables.materializer(X)(newcols)
+    if !isdataframe(X)
+        cols = Tables.columntable(X) # named tuple of vectors
+        newcols = project(cols, c)
+        return Tables.materializer(X)(newcols)
+    else
+        return X[!, c]
+    end
 end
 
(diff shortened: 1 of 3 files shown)


src/interface/data_utils.jl
```diff
@@ -97,14 +97,22 @@ function MMI.selectrows(::FI, ::Val{:table}, X, r)
 end
 
 function MMI.selectcols(::FI, ::Val{:table}, X, c::Union{Symbol, Integer})
-    cols = Tables.columntable(X) # named tuple of vectors
-    return cols[c]
+    if !isdataframe(X)
+        cols = Tables.columntable(X) # named tuple of vectors
```
OkonSamuel, line 101:
```suggestion
        cols = Tables.columns(X)
```

src/interface/data_utils.jl
```diff
@@ -97,14 +97,22 @@ function MMI.selectrows(::FI, ::Val{:table}, X, r)
...
+        return cols[c]
```
OkonSamuel, line 102:
```suggestion
        return Tables.getcolumn(cols, c)
```

src/interface/data_utils.jl
OkonSamuel: In this case there is no need to generate a named tuple of vectors which would have bad compile time performance. In fact I say we do not need the `isdataframe` check in this case.

src/interface/data_utils.jl
```diff
@@ -97,14 +97,22 @@ function MMI.selectrows(::FI, ::Val{:table}, X, r)
...
+    else
+        return X[!, c]
+    end
 end
 
 function MMI.selectcols(::FI, ::Val{:table}, X, c::AbstractArray)
```
OkonSamuel, line 108:
```suggestion
function MMI.selectcols(::FI, ::Val{:table}, X, c::Union{Colon, AbstractArray})
```

tlienart:
thanks for the review @OkonSamuel feel free to take this over, I'm not close enough to the Tables API to know what's best so I trust your choices are better, it might be good to check the tests are right, maybe do a side test with what brought Jack_N to open the issue in the first place and check that there's no bottleneck anymore?

jackn11:
Thank you both for helping with the changes. Also, it may be a good idea to make the slight modification to the selectcols function so that the '!' is not needed in the if statement. The code is below so you can see what I mean. Removing the unnecessary double negative should help keep the code clean and understandable.

`function MMI.selectcols(::FI, ::Val{:table}, X, c::Union{Symbol, Integer})
    if isdataframe(X)
        return X[!, c]
    else
        cols = Tables.columntable(X) # named tuple of vectors
        return cols[c]
    end
end

function MMI.selectcols(::FI, ::Val{:table}, X, c::AbstractArray)
    if isdataframe(X)
        return X[!, c]
    else
        cols = Tables.columntable(X) # named tuple of vectors
        newcols = project(cols, c)
        return Tables.materializer(X)(newcols)
    end
end`

src/interface/data_utils.jl
```diff
@@ -124,7 +128,7 @@ function project(t::NamedTuple, indices::AbstractArray{<:Integer})
 end
 
 # utils for selectrows
-typename(X) = split(string(supertype(typeof(X)).name), '.')[end]
+typename(X) = split(string(supertype(typeof(X))), '.')[end]
```
ablaom, line 131:
```suggestion
# NOTE. THE FOLLOWING USES NON PUBLIC JULIA API. 
# It is a hack to avoid DataFrames as a dependency (see implementation
# of `selectrows`) until https://github.com/JuliaData/Tables.jl/pull/278
# is resolved
typename(X) = split(string(supertype(typeof(X))), '.')[end]
```

ablaom (changes requested):
@OkonSamuel I'm happy with this, ~~modulo my single suggestion.~~ Please merge and tag a new release at your leisure.

tlienart:
I just merged (it was my branch so not sure if sam could do it) will leave the tag & release to you guys

ablaom:
Thank you @tlienart and @OkonSamuel . A new release just tagged.

PR #793, opened 2022-06-16
Fix bugs and inconsistencies in documentation on `evaluate!`s `resampling` keyword

wolthom:
This pull request contains minor changes in `resampling.jl`.
The changes const of:
1. Fixing a bug in the `resampling` code example showing a vector of tuples
2. Improve the documentation to use "vector of tuples" instead of "vector of pairs" to prevent confusion with the `Pair` datatype
3. Improve the error message in case a wrong type was supplied for the `resampling` keyword


ablaom (approved):
Nice catch, thanks @wolthom 🙏🏾

PR #798, opened 2022-06-29
Add `feature_importances` machine inspection method.

OkonSamuel:
This PR implements `feature_importances` method that returns a list of feature=>importance pairs for fitted machines with underlying models that supports intrinsic feature importances. (i.e models that overload the `reports_feature_importances` trait.)

**This Requires**
- [x] https://github.com/JuliaAI/MLJModelInterface.jl/pull/148

### DIFF
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -258,7 +258,7 @@ export @mlj_model, metadata_pkg, metadata_model
 export fit, update, update_data, transform, inverse_transform,
     fitted_params, predict, predict_mode, predict_mean,
     predict_median, predict_joint,
-    evaluate, clean!, training_losses, intrinsic_importances
+    evaluate, clean!, training_losses, feature_importances
 
 # data operations
 export matrix, int, classes, decoder, table,
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -53,7 +53,7 @@ import MLJModelInterface: fit, update, update_data, transform,
     predict_mean, predict_median, predict_joint,
     evaluate, clean!, is_same_except,
     save, restore, is_same_except, istransparent,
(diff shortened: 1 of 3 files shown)


src/MLJBase.jl
```diff
@@ -53,7 +53,7 @@ import MLJModelInterface: fit, update, update_data, transform,
     predict_mean, predict_median, predict_joint,
     evaluate, clean!, is_same_except,
     save, restore, is_same_except, istransparent,
-    params, training_losses
+    params, training_losses, intrinsic_importances
```
ablaom, line 56:
```suggestion
    params, training_losses, feature_importances
```

src/machines.jl
```diff
@@ -821,7 +821,29 @@ function training_losses(mach::Machine)
     end
 end
 
+"""
+    feature_importances(mach::Machine)
+
+Return a list of feature => importance pairs for a fitted machine, `mach` if it's underlying model 
+supports intrinsic feature importances. Otherwise, returns `nothing`.
+
+"""
```
ablaom, line 831: I don't think a carriage return is allowed here:
```suggestion
```

src/machines.jl
```diff
@@ -821,7 +821,29 @@ function training_losses(mach::Machine)
     end
 end
 
+"""
+    feature_importances(mach::Machine)
+
+Return a list of feature => importance pairs for a fitted machine, `mach` if it's underlying model 
+supports intrinsic feature importances. Otherwise, returns `nothing`.
```
ablaom, line 828:
```suggestion
Return a list of `feature => importance` pairs for a fitted machine, 
`mach`,  if supported by the underlying model, i.e., if 
`reports_feature_importances(mach.model) == true`.  Otherwise return
`nothing`.
```

PR #822, opened 2022-08-10
[DO NOT MERGE] See if CI passes on Julia nightly when `JULIA_NUM_THREADS` is set to `2`

DilumAluthge:


ablaom:
@DilumAluthge Can this be closed now?

DilumAluthge:
Actually, we should merge this, so that we can detect if tests start failing on Julia nightly again.

ablaom:
I guess this sounds silly, but I don't like all those red crosses I get after my commits when nightly fails, which it frequently does. That's why I disabled nightly. Is there a workaround?

DilumAluthge:
Yeah, it is unfortunate that sometimes Julia nightly is failing. The problem is that if we don't test on Julia nightly, we don't catch bugs like https://github.com/JuliaLang/julia/issues/46405 until after a new version of Julia has been released.

PR #836, opened 2022-09-09
remove unbound type parameters

nsajko:
Unbound type parameters often cause performance issues and run time dispatch.

Issue found using https://github.com/JuliaLang/julia/pull/46608

### DIFF
--- a/src/hyperparam/one_dimensional_range_methods.jl
+++ b/src/hyperparam/one_dimensional_range_methods.jl
@@ -423,11 +423,11 @@ function Distributions.sampler(r::NominalRange{T,N}) where {T, N}
     return sampler(r, fill(1/N, N))
 end
 
-Base.rand(s::NominalSampler, dims::Integer...) where I<:Integer =
+Base.rand(s::NominalSampler, dims::I...) where I<:Integer =
     broadcast(idx -> s.values[idx], rand(s.distribution, dims...))
 Base.rand(rng::AbstractRNG,
           s::NominalSampler,
-          dims::Integer...) where I<:Integer =
+          dims::I...) where I<:Integer =
     broadcast(idx -> s.values[idx], rand(rng, s.distribution, dims...))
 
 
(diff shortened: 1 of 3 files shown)


src/hyperparam/one_dimensional_range_methods.jl
```diff
@@ -423,11 +423,11 @@ function Distributions.sampler(r::NominalRange{T,N}) where {T, N}
     return sampler(r, fill(1/N, N))
 end
 
-Base.rand(s::NominalSampler, dims::Integer...) where I<:Integer =
+Base.rand(s::NominalSampler, dims::I...) where I<:Integer =
```
ablaom, line 426: Seems to me this could be simplified further, as the concrete subtype `I <: Integer` is not used.  

```suggestion
Base.rand(s::NominalSampler, dims::Integer...) = 
```

Right?

src/hyperparam/one_dimensional_range_methods.jl
```diff
@@ -423,11 +423,11 @@ function Distributions.sampler(r::NominalRange{T,N}) where {T, N}
...
     broadcast(idx -> s.values[idx], rand(s.distribution, dims...))
 Base.rand(rng::AbstractRNG,
           s::NominalSampler,
-          dims::Integer...) where I<:Integer =
+          dims::I...) where I<:Integer =
```
ablaom, line 430: As above:

```suggestion
          dims::Integer...) =
```

ablaom (commented):
Thanks @nsajko.

src/hyperparam/one_dimensional_range_methods.jl
nsajko: Depends on whether you want to allow dims to be heterogeneous (for example, a mix of Int32 and Int64). The syntax I chose disallows that, while your simplified syntax allows it.

nsajko: I was wrong in the first place to make that choice arbitrarily, instead of asking first.

ablaom: > Depends on whether you want to allow dims to be heterogeneous (for example, a mix of Int32 and Int64). 

Fair point. Then this is fine as is.

ablaom:
@nsajko Thanks for this contribution.

Issue #831, opened 2022-08-29
What's wrong with the way we export learning networks as new model types


ablaom:
The learning networks API appeared in the very earlier versions of MLJ and over time has grown to become a bit of a hack, in response to a great many feature requests. Here are some complaints:

- From the user's point of view, the process for exporting a learning network is **too complicated** (eg, first define learning network machine, etc)
- It is also **too mysterious** (too much hidden knowledge)
- From the developer's point of view, it is also **too complicated** and **too mysterious** making it a challenge to maintain and enhance, even for those very familiar with it.
- There exist unnatural restrictions on what a user can do, such as one that led to https://github.com/JuliaAI/MLJBase.jl/issues/377. This becomes worse if we move to allowing immutable model structs (for, eg, integration with TableTransforms.jl)
- Retraining a composite is not "smart" if a component model is replaced (only if it is mutated)
- The fact that `fitted_params(::Machine)` and `report(::Machine)` need special casing for machines bound to composite models feels like an unnecessary complication and is making it difficult to design uniform interfaces; see eg, https://github.com/alan-turing-institute/MLJ.jl/issues/841#issuecomment-1229698817.

These issues do not just affect esoteric applications to model composition but are holding back some important developments, eg, in outlier detection, and TableTransforms.jl  integration.

The root causes for these issues are:

**1.** The way we currently establish a mapping from composite model hyperparameters for component models to the corresponding machines in the network that are meant to point to those component models. 

**2.** The fact that we prematurely merge reports from `fit` with reports from an operation, instead of keeping them separate in the machine and providing a model-specific method to say how to combine them.

I've had a pretty good idea on how to resolve **1**, but it will take a little time to put together. Fixing **2** after **1** will be easier, and what is needed there is more-or-less obvious. Stay tuned for PR's to address these. Thank you for your patience.

ablaom:
Resolved by https://github.com/JuliaAI/MLJBase.jl/issues/852

PR #841, opened 2022-09-16
Add new option for exporting learning networks as stand-alone composite model types

ablaom:
This ~~non-breaking~~ PR addresses #831.

It is mildly breaking: Where previously `report(mach)` returned `NamedTuple()`, it will now return `nothing` instead.

Requires:

- [x] https://github.com/JuliaAI/MLJModelInterface.jl/pull/160

Has been tested (with above MLJModelInterface PR) as non-breaking for MLJTuning, MLJEnsembles, MLJIteration.
Has been tested with MLJTestIntegration, which applies integration tests for majority of model providing packages. 

### Summary

This PR adds a new abstract type `NetworkComposite`, with subtypes `DeterministicNetworkComposite`, etc, for the purpose of "exporting" learning networks as new stand-alone model types. It is intended to render all existing alternatives obsolete, **but these are not touched in this PR**. It is simpler, in that there are no "learning network machines" , no `Surrogate` models, and no special `return!` method to call. Refer to #831 for a list of other drawbacks overcome by the new approach.

Below is a demonstration of exporting a learning network to combine predictions of two regressors. It additionally exposes in the new model's report the output of a node called `disagreement` (measuring disagreement between the two regressors, at time of training):

#### Step 1: Define the composite type - must subtype `NetworkComposite`

```julia
using MLJBase

mutable struct DoubleRegressor <: DeterministicNetworkComposite
    regressor1
    regressor2
    mix::Float64
end
```

#### Step 2:  Implement `MLJBase.prefit` (method added in this PR)

This method  has same signature as `fit(model, ...)` and returns a named tuple constituting an "interface" to a learning network constructed out of it's arguments. A novelty is that component models are represented by the names (symbols) of the relevant field of the composite.

```julia
function MLJBase.prefit(composite::DoubleRegressor, verbosity, X, y)
    Xs = source(X)
    ys = source(y)

    mach1 = machine(:regressor1, Xs, ys)   #  <--- symbol instead of model
    mach2 = machine(:regressor2, Xs, ys)

    yhat1 = predict(mach1, Xs)
    yhat2 = predict(mach2, Xs)

    # node to return agreement between the regressor predictions:
    disagreement = node((y1, y2) -> l2(y1, y2), yhat1, yhat2)

    # get the weighted average the predictions of the regressors:
    λ = composite.mix
    yhat = (1 - λ)*yhat1 + λ*yhat2

    return (
        predict = yhat,
        report= (; training_disagreement=disagreement)
    )
end
```

### Details

A new documentation PR at MLJ has already been posted. The  relevant section is [here](https://github.com/alan-turing-institute/MLJ.jl/blob/903f0b3a6163a06bec18f1a593ba3b77146e1785/docs/src/learning_networks.md)

There are two additional changes introduced in this PR which make the above possible:

#### Symbols instead of models in machines

One can bind data in a machine to a "virtual" model, which is just a **symbol**, as in `mach = machine(:transformer, X)`. If `my_composite` is any object with `:transformer` as a property, then `fit!(mach; composite=my_composite)` will train the machine with `:transformer` replaced with `getproperty(my_composite, :transformer)`. This "run-time interpolation" of models propagates to all machines in a learning network. The general user need not bother with this, but if they know about it, the export process above is less mysterious. 

#### Separate training reports and operation reports in machines

In a change purely internal to MLJBase, the `report` field of a machine is now a dictionary `d`, keyed on method (`:fit`, `:predict`, etc). When an [reporting operation](https://github.com/JuliaAI/MLJBase.jl/pull/806) is called on the machine, it's report is added to the dictionary; previously it was merged with the training report. Now when you call `report` on a *machine*, the various reports in dictionary get merged by calling a [new method](https://github.com/JuliaAI/MLJModelInterface.jl/pull/160) `MLJModelInterface.report(::Model, d)`. This method can be overloaded by a model but there is a fallback that does a pretty good job of it (the fallback unexpectedly "just worked" for the new `NetworkComposite` types). 

In the existing export process, `fitted_params(::Machine, ...)` and `report(::Machine, ...)` have to be special-cased for composite model types. With this new approach, that is no longer necessary, which restores uniformity to the interface. 

### To do

- [x] Add `save` and `restore` methods for serialising `<:NetworkComposite`
- [x] Add facility to expose result of calling an internal node in the composite model `fitted_params`, as we have already for `report`s.


ablaom:
@davnn This PR is finished or very nearly so. I wonder if you can confirm that this will serve your purposes. You can see from the example how to add "report nodes" to the model report. To test, you currently need MLJModelInterface#banana which may not be not merged yet. Where you previously subtyped `UnsupervisedDetectorComposite`, for example, you would now subtype `UnsupervisedDetectorNetworkComposite`. 

Worth noting that for `NetworkComposite` types, operations such as `predict` and `transform` will always be reporting operations, in the sense of https://github.com/JuliaAI/MLJBase.jl/pull/806. That is, calling them on *models* returns *two* items - the usual result and a report (which could be `nothing`). In your case, you have no *internal* models with this behavior, so these extra report items for the composite are just copies of the `fit` report and you can ignore them. I'm only mentioning all this in case you have model-level tests for your composite models which are not expecting this. Machine reports should be unchanged, with the possible exception that items with `nothing` or `NamedTuple()` as values are now dropped. 

Thank you for your patience.

davnn:
Thanks, I'll try to have a look asap.

ExpandingMan:
Thought I'd comment briefly on this approach since I tend to use `node` and learning networks.
- Having a user defined struct with fields for the machines seems like a *huge* improvement for the transparency of the resulting object.  The surrogate approach was quite opaque and if you had to get a machine out of the interior of it the task was non-trivial.
- I wonder if the `machine` with a `Symbol` argument is a necessary pun?  I'm not entirely sure what's happening underneath that, and I think it would be really nice if whatever that was could be user-facing so that it's totally obvious to users what is happening in `prefit`.
- In the long run, I wonder if pipelines and ensembles can be deprecated in favor of "shorthand" for these networks?  It always struck me as a bit odd that there are so many different ways of composing models when one approach seems to encompass all of the others.  On the other hand, it might be pretty hard to make it as dead-simple as it currently is to compose models in pipelines that way without excessive macro magic, which I suppose might have something to do with pipelines being their own types.

ablaom:
@ExpandingMan Thanks for the feedback!

> * I wonder if the `machine` with a `Symbol` argument is a necessary pun?  I'm not entirely sure what's happening underneath that, and I think it would be really nice if whatever that was could be user-facing so that it's totally obvious to users what is happening in `prefit`.

Good point. Indeed, there is a user facing aspect explained in the new docstring for the machine constructor: https://github.com/JuliaAI/MLJBase.jl/blob/d5394001072160e38a8d3d2aab683f2fb05d9dc5/src/machines.jl#L299

I am also working a complete re-write of the documentation, which will explain this. 

> * In the long run, I wonder if pipelines and ensembles can be deprecated in favor of "shorthand" for these networks?  It always struck me as a bit odd that there are so many different ways of composing models when one approach seems to encompass all of the others.  On the other hand, it might be pretty hard to make it as dead-simple as it currently is to compose models in pipelines that way without excessive macro magic, which I suppose might have something to do with pipelines being their own types.

I've found it very difficult to find a single solution that is both powerful and convenient. Exporting learning networks, as simple as I have tried to make them, are far more work than pipelines or stacks. 

In a sense a pipeline is just shorthand for a learning network; I mean that is how it is implemented. But you can define a pipeline without creating your own type each time, and mapping type fields to models in the network. 

Pipelines used to be generated using macros (`@pipeline` still exists) which auto-generates a new type each time, but that didn't work too well. I've found the [newer alternative](https://alan-turing-institute.github.io/MLJ.jl/dev/linear_pipelines/) to be far more robust. 

You may be right that some kind of DSL can be build on top the basics to tie everything together. I would certainly be interested in any attempt to do that.

ExpandingMan:
> You may be right that some kind of DSL can be build on top the basics to tie everything together. I would certainly be interested in any attempt to do that.

Not really what I was going for.  I think a simple implementation of pipelines might be preferable to a complicated implementation that's needed just to express a pipeline in the language of a generalized network.  Like I said in the other issue, pipelines are probably "good enough" in something like 95% of use cases so I don't necessarily think it would be worth complicating them just to have everything be in terms of a generalized network.

ablaom:
@olivierlabayle Did you want to comment on this PR? I realize you might not want to review it as it is quite long, but if you're willing and interested that would be extremely welcome.

ablaom:
Keen to move this along soon. 

@davnn Have you made any progress testing this PR with OutlierDetection.jl?

ablaom:
The MLJ doc update associated with the change proposed here has been posted. The new rewritten Learning Networks section is [here](https://github.com/alan-turing-institute/MLJ.jl/blob/903f0b3a6163a06bec18f1a593ba3b77146e1785/docs/src/learning_networks.md).

davnn:
> Keen to move this along soon.
> 
> @davnn Have you made any progress testing this PR with OutlierDetection.jl?

Hey, sorry I've been very busy. I'm trying to integrate it currently, my approach is as follows:

We defined wrappers for detectors (previously resulting in composite detectors, now in composite network detectors) that take a variabled number of arguments as base detectors. Previously these base detector were used in `machine(detector, data...)`, now we should use `machine(:detector_name, data...)`. Where does `machine(:detector_name, data...)` to `machine(getproperty(composite, :detector_name, data...)` happen?

I used something like `map(d -> augmented_transform(MLJ.machine(d, Xs), Xs), getfield(model, :detectors))` before, replacing that with `map(d -> augmented_transform(MLJ.machine(d, Xs), Xs), tuple_of_detector_symbols)` doesn't work, but how am I supposed to map over an iterable of submodels then?

olivierlabayle:
> @olivierlabayle Did you want to comment on this PR? I realize you might not want to review it as it is quite long, but if you're willing and interested that would be extremely welcome.

Sorry I've just realised I've missed that! I'll try to have a look in the next couple days, it looks really promising, but please feel free to go ahead since it's been open for some time already.

src/machines.jl
```diff
@@ -355,41 +336,51 @@ machine(T::Type{<:Model}, args...; kwargs...) =
...
     return Machine(model, args...; kwargs...)
 end
 
-function machine(model::Model, arg1::AbstractNode, args::AbstractNode...;
+function machine(model::Union{Model,Symbol}, arg1::AbstractNode, args::AbstractNode...;
                  kwargs...)
     return Machine(model, arg1, args...; kwargs...)
 end
 
+function machine(model::Symbol, arg1::AbstractNode, args::AbstractNode...;
```
olivierlabayle, line 380: I think this method is now redundant with the method above.

test/composition/models/network_composite.jl
```diff
@@ -0,0 +1,733 @@
...
+function MLJBase.prefit(composite::WatermelonComposite, verbosity, X)
+
+    Xs = source(X)
+
+    len = node(Xs) do X
+        size(Tables.matrix(X), 1)
+    end
+
+    W = transform(machine(:scaler), Xs)
+    ytrain = predict(machine(:clusterer), Xs)
```
olivierlabayle, line 88: I think I understand you want to support both the symbolic and attribute acceses in the definition of the learning network? What happens if I do: `composite.clusterer` inead of `:clusterer`? It seems to break the fit report, the clusterer will not be part of it anymore.

olivierlabayle:
Thanks @ablaom for this simplification of the learning network definition process! My only reservations are about the symbolic access of models in the definition of the network itself. For that magic to occur you've had to hack into and complexify `fit_only!`. Writing `composite.model` instead of `:model` doesn't seem like a huge burden to me and would also be easier to debug.  Is there another benefit beyond this short keyed access?

ablaom:
@davnn Thanks for looking at this. 

> Where does machine(:detector_name, data...) to machine(getproperty(composite, :detector_name, data...) happen?

Whenever you call `fit!(node, composite=composite)` where `node` is a node, then this will interpolate symbols in machines in the network subtended at `node` with the model values in the property accessible object `composite` (assuming those symbols are properties) before training the network. (See also proposed [docs](https://github.com/alan-turing-institute/MLJ.jl/blob/banana/docs/src/learning_networks.md#step-1---define-a-new-model-struct).) There is generic `MMI.fit(::NetworkComposite, verbosity, data...)` method which makes exactly this call on the greatest lower bound node in the network (including "report" nodes) defined by the "network interface" returned by `prefit`. 

> replacing that with map(d -> augmented_transform(MLJ.machine(d, Xs), Xs), tuple_of_detector_symbols) doesn't work

Mmm. I think I need more context to understand why this is not working, as I've by now forgotten the details. Can you please point me to the relevant code and I'll take a look. (I didn't think we had a `augmented_transform` in MMI, as we decided to abandon that?? Is this something you added locally?)

An alternative is to leave that particular line alone - that should still work. You don't have to do the symbol replacement. The problem will be that the reports and fitted_params for the individual detectors in your ensemble will not automatically appear in the composite model's report and fitted_params. In the case of reports, we can (probably) add them "manually" using report nodes; in the case of fitted_params I'm now thinking of adding a facility to do that also in this PR. But I should first like to see if we can avoid this workaround.

test/composition/models/network_composite.jl
ablaom: Exactly. You don't have to replace with symbols, but then you loose the contributions to the report and fitted_params. The symbols act as hooks into those machines we want reports/fitted_params for. Said differently, it makes the mapping from composite fields to (one or more) machines explicit. The old way of mapping models to machines was less robust and significantly complicating . For example, different machines pointing to *identical* model *instances* (but meant to correspond to different composite fields) could not be distinguished and led to a strange restriction (mentioned in #831). This issue get's worse if we allow immutable model structs (TableTransforms.jl, for example uses only immutable structs). 

The down side is that you may want to reference a component model outside of a `machine` constructor, maybe to internally change a hyperparameter, and then you can't do the substitution and must manually expose the report/fitted_params if wanted. (I still need to add a way to do this for fitted_params; we already have "report" nodes ).

ablaom:
@olivierlabayle Thanks for taking the time to look over this substantial PR and for your valuable feedback.

> Thanks @ablaom for this simplification of the learning network definition process! My only reservations are about the symbolic access of models in the definition of the network itself. For that magic to occur you've had to hack into and complexify fit_only!. Writing composite.model instead of :model doesn't seem like a huge burden to me and would also be easier to debug. Is there another benefit beyond this short keyed access?

Yes, there is some complication to `fit_only!`, but I think this is way less hacky than what is done now to expose component model reports and fitted_params, and so forth, no? Not sure what you mean by "short keyed access". The benefits are overcoming the shortcomings detailed in #831. As explained in comment above, writing `composite.model` will mean reports and fitted_params are missing from the composite model report/fitted_params, but is still allowed. 

> would also be easier to debug.

Yes, it's slightly more tedious to debug. When working with an unexported learning network with the symbol replacements, you need to add the `composite=...` keyword argument to your `fit!(node)` calls (see [here](https://github.com/alan-turing-institute/MLJ.jl/blob/banana/docs/src/learning_networks.md#step-1---define-a-new-model-struct))

test/composition/models/network_composite.jl
olivierlabayle: Oh, thank you for the explanation, I never really dived into the reporting/fitted_params parts of the API. I understand the issue now and the benefit of the symbolic access, All clear!

davnn:
> Mmm. I think I need more context to understand why this is not working, as I've by now forgotten the details. Can you please point me to the relevant code and I'll take a look. (I didn't think we had a `augmented_transform` in MMI, as we decided to abandon that?? Is this something you added locally?)

Yes, I've implemented `augmented_transform` internally for detector models, which would now have to be extended for the new Network interface. The implementations of augmented transform work on machines, but I see that the new network works on models and fitresults, i.e. <https://github.com/JuliaAI/MLJBase.jl/blob/banana/src/operations.jl#L224>. I'm not sure how to translate this API to augmented transform. I tried to add an augmented_transform node <https://github.com/OutlierDetectionJL/OutlierDetection.jl/blob/banana/src/mlj_wrappers.jl#L193>, but that doesn't work (maybe because signatures only work with a hard-coded set of operations? <https://github.com/JuliaAI/MLJBase.jl/blob/banana/src/composition/learning_networks/signatures.jl#L148>).

ablaom:
@davnn I'm going to continue the OutlierDetection specific discussion in a new thread which will appear shortly below.

PR #848, opened 2022-10-17
Relax type restriction on `y` in `train_test_pairs(::StratifiedCV,  rows, y)`

### LINKED ISSUE #847: allow `train_test_pairs` to work on generic vector `y`?
after a long time, I still cannot figure out how to get
```julia
	class_labels = vcat(["spam" for _ = 1:3], ["not spam" for i = 1:9])
	ttp = train_test_pairs(kf, 1:length(class_labels), class_labels)
	for (k, (ids_train, ids_test)) in enumerate(ttp)
		# train on ids_train
		# test on ids_test
	end
```
to work. would be nice to allow `train_test_pairs` to accept a generic vector of class labels so that this can be used outside the context of ML workflows in this package. I'm looking for something generic like scikitlearn's stratified K-folds. thanks!

the error I get is:

```
ArgumentError: Supplied target has scitype AbstractVector{ScientificTypesBase.Textual} but stratified cross-validation applies only to classification problems.

    train_test_pairs(::MLJBase.StratifiedCV, ::UnitRange{Int64}, ::Vector{String})@resampling.jl:407
    top-level scope@[Local: 3](http://localhost:1235/edit?id=c4c85f16-4c12-11ed-3b74-134639249bb1#)[inlined]
```

ablaom:
Resolves #847.

Also fixes some julia 1.8 type parameter warnings.

### DIFF
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -403,12 +403,6 @@ StratifiedCV(; nfolds::Int=6,  shuffle=nothing, rng=nothing) =
 function train_test_pairs(stratified_cv::StratifiedCV, rows, y)
 
     st = scitype(y)
-    if !(st <: AbstractArray{<:Finite})
-        throw(ArgumentError(
-            "Supplied target has scitype $st but stratified " *
-            "cross-validation applies only to classification problems. "
-        ))
-    end
 
     if stratified_cv.shuffle
         rows=shuffle!(stratified_cv.rng, collect(rows))
--- a/src/measures/finite.jl
+++ b/src/measures/finite.jl
@@ -530,12 +530,7 @@ for M in (:MulticlassTruePositiveRate, :MulticlassTrueNegativeRate,
             average::T
             return_type::Type{U}
         end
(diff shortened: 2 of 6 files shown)


src/resampling.jl
```diff
@@ -403,12 +403,6 @@ StratifiedCV(; nfolds::Int=6,  shuffle=nothing, rng=nothing) =
 function train_test_pairs(stratified_cv::StratifiedCV, rows, y)
 
     st = scitype(y)
-    if !(st <: AbstractArray{<:Finite})
```
OkonSamuel, line 406: Shouldn't we have at least a warning here, even if we aren't throwing an error.

ablaom: I think that in the long term, this functionality is going to be moved into a separate package that non-MLJ packages can use. For use-cases outside of MLJ, this warning is not going to be relevant (and so annoying). 

In the MLJ case, wrong encoding for classification targets are likely to be caught elsewhere in the workflow, no?

Issue #852, opened 2022-10-27
Towards a 0.21 release


ablaom:
To be merged onto the branch, [for-a-0-point-21-release](https://github.com/JuliaAI/MLJBase.jl/tree/for-a-0-point-21-release):

- [x] #841 
- [x] #857 
- [x] #854 
- [x] #858 
- [x]  #860 and make calling `@pipeline` an error. 
- [x] bump version to 0.21 in Project.toml

To be included in release notes:

- (**breaking**) Remove the previously deprecated `@pipeline` macro (replaced by `Pipeline` and syntactic sugar).
- (**enhancement**) Introduce new protocol for exporting learning networks via new family of supertypes  `*NetworkComposite` (#841, #831) 
- (**mildly breaking**) Remove key-value pairs from reports and fitted parameters that have `nothing` or `NamedTuple()` as a value. 
- (**breaking**) Re-implement `Stack`, `TransformedTargetModel` and pipelines (`Pipeline`) as `NetworkComposite` types. What is breaking is that some reports and fitted parameters will be missing some these keys: `machines`, `fitted_params_given_machine`, `report_given_machine`.  However, the only information contained in those items that is not rendered redundant by the existing items are considered private. 
- (**breaking**) Remove the formerly deprecated `@pipeline` macro, now rendered redundant by [`Pipeline`](https://alan-turing-institute.github.io/MLJ.jl/dev/linear_pipelines/) constructor (or `model1 |> model2 |> ... ` syntax) and  [`TransformedTargetModel`](https://alan-turing-institute.github.io/MLJ.jl/dev/target_transformations/#Target-Transformations).

PR #854, opened 2022-10-31
Re-implement `Stack` as a `NetworkComposite` model

ablaom:
Currently `Stack` is implemented as a `Composite` model, an abstract type to be deprecated in favour of `NetworkComposite` added by #841.

This is mildly breaking, as indicated in #852.


ablaom:
FYI @olivierlabayle

PR #886, opened 2023-03-06
Cleanup measures to lower burden for new implementations

ablaom:
**1**. When a new performance measure is added, the methods `MLJBase.single` (single observation) and/or `MLJBase.call` (batch of observations) must be implemented. Some current implementations seem unnecessary specific, with regards to the types of their arguments. When these are relaxed (as in this PR) many type instabilities associated with `Missing` appear. To resolve these ambiguities in a naive way requires adding a lot of code, here and for any new implementations. Instead, this PR introduces a small design change in the way missing values are handled, described below. This greatly simplifies the burden on implementations, as reflected in the diff for the updated file `/src/measures/REAMDE.md`. 

**2**. There is a roughly analogous source of method ambiguities in `default_measure` with a similar resolution provided in this PR but not detailed below. 

I can't think how these changes could be breaking, but as an extra reassurance I will:

- [x] Run MLJTestIntegration.jl tests 

to ensure that at least there are no breakages within the MLJ + model interfaces stack.

### Design changes

At present, `single(::Unaggregated, args...)` needs to work when one of the `args` is `missing`. After this PR, `single` no longer needs to work for `missing` values and the `single` method is instead wrapped in a new function `robust_single` which adds the `missing` value handling. Where `call` fallbacks previously broadcasted over `single`, they now broadcast over `robust_single`. 

Here's the key code replacing a previous naive fallback implementation for `single(m, ::Missing, ::Any)`, etc:

```julia
# `robust_single` can accept `missing` observations/predictions but is never overloaded;
# `single` is overloaded but does not need to handle missings. This factoring allows us
# to avoid method ambiguities which are cumbersome to avoid with only one function.

robust_single(args...) = single(args...)
robust_single(m, ::Missing, ::Missing) = missing
robust_single(m, ::Missing, η) = missing
robust_single(m, η̂, ::Missing) = missing
```

### DIFF
--- a/src/measures/README.md
+++ b/src/measures/README.md
@@ -52,34 +52,10 @@ specified as part of an aggregation of multi-observation measurements.
 
 ### Unaggregated measures implement `single`
 
-To implement an `Unaggregated` measure, it suffices to implement
-`single(measure, η̂, η)`, which should return a measurement (e.g., a
-float) for a single example `(η̂, η)` (e.g., a pair of
-floats). Behavior on `missing` values is handled by fallbacks:
-
-```julia
-single(::Unaggregated, η̂::Missing, η)          = missing
-single(::Unaggregated, η̂,          η::Missing) = missing
-```
-
-Be sure to annotate the type of `η̂` and `η` in the implementation
-of `single` to avoid a type ambiguity with these fallbacks. For example
-
-```julia
-single(::MyUnaggregatedMeasure, η̂::Real, η::Real) = ...
-```
-or
-
-```
-single(::MyAggregatedMeasure, η̂::UnivariateFinite, η::Label) = ...
-```
-
-Here `Label` is just a large union type defined in MLJBase excluding
-`missing` as an instance:
-
-```julia
-const Label = Union{CategoricalValue,Number,AbstractString,Symbol,AbstractChar}
-```
+To implement an `Unaggregated` measure, it suffices to implement `single(measure, η̂, η)`,
+which should return a measurement (e.g., a float) for a single example `(η̂, η)` (e.g., a
+pair of floats). There is no need for `single` to handle `missing` values. (Internally, a
+wrapper function `robust_single` handles these.)
 
 If only `single` is implemented, then the measure will automatically
 support per-observation weights and, where that makes sense, per-class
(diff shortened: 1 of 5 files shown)


ablaom:
@rikhuijzer Any chance you could look over this one in the next week or so?

src/measures/README.md
```diff
@@ -52,34 +52,10 @@ specified as part of an aggregation of multi-observation measurements.
...
-single(::MyAggregatedMeasure, η̂::UnivariateFinite, η::Label) = ...
-```
-
-Here `Label` is just a large union type defined in MLJBase excluding
-`missing` as an instance:
-
-```julia
-const Label = Union{CategoricalValue,Number,AbstractString,Symbol,AbstractChar}
-```
+To implement an `Unaggregated` measure, it suffices to implement `single(measure, η̂, η)`,
```
rikhuijzer, line 55: This is slightly outside the scope of this PR, but could we also annotate here which is for the true values and which is for the predictions like `x_true` and `x_pred`? Not everyone is familiar with the ` ̂y` vs. `y` convention, I think

src/measures/measures.jl
```diff
@@ -102,21 +102,27 @@ end
...
 # closure for broadcasting:
-single(measure::Measure) = (ηhat, η) -> single(measure, ηhat, η)
+robust_single(measure::Measure) = (ηhat, η) -> robust_single(measure, ηhat, η)
 
-call(measure::Unaggregated, yhat, y) = broadcast(single(measure), yhat, y)
-function call(measure::Unaggregated, yhat, y, w::Arr)
-    unweighted = broadcast(single(measure), yhat, y) # `single` closure below
+call(measure::Unaggregated, yhat, y) = broadcast(robust_single(measure), yhat, y)
+function call(measure::Unaggregated, yhat, y, w::AbstractArray)
+    unweighted = broadcast(robust_single(measure), yhat, y)
```
rikhuijzer, line 121: Ah, so this is the core of the whole PR. The idea is that this used to be a call to `single` implemented by the consumer, but instead `robust_single` is now called which handles `missing`s so that the consumer implementation doesn't have to.

If that is correct, then I would say this PR is good to go.

rikhuijzer (approved):

rikhuijzer:
> @rikhuijzer Any chance you could look over this one in the next week or so?

Sure. Done. From looking through the code, I don't see any possible problems with the implementation. If the integration tests also work, this doesn't break old implementation, and my understanding is correct as stated in https://github.com/JuliaAI/MLJBase.jl/pull/886#discussion_r1127523233, then I'd say this is good to go.

Looks like a nice improvement for the consumers. Annotating with special vector types looked way more complex than the current version.

src/measures/README.md
ablaom: Agreed. They are better names. But let's kick the can on this one. I'm working on exporting measures out to a new package and will do something like that there.

src/measures/measures.jl
ablaom: Yep. This is the essential point. I got the idea from the excellent julia docs on [mitigating method ambiguities](https://docs.julialang.org/en/v1/manual/methods/#man-method-design-ambiguities).

ablaom:
@rikhuijzer Thanks for the speedy review of this somewhat technical PR.

Issue #570, opened 2021-06-06
Investigate loading time of LossFunctions.jl


ablaom:
This used to be quite long, according to [this comment](See https://github.com/alan-turing-institute/MLJBase.jl/issues/77#issuecomment-549041739). This seems surprising, given the modest deps:

```
LearnBase = "7f8f8fb0-2700-5f03-b4bd-41f8cfc144b6"
Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
```

Worth reviewing.

juliohm:
We dropped most deps in a recent refactoring of the code. LossFunctions.jl v0.9 only depends on CategoricalArrays.jl.

ablaom:
Will check the timings again after resolving #896

ablaom:
Based on PR at #898 :

```julia
julia> @time_imports using MLJBase
[ Info: Precompiling MLJBase [a7f614a8-145f-11e9-1d2a-a57a1082229d]
      0.2 ms  Reexport
      0.5 ms  ScientificTypesBase
      0.1 ms  DataValueInterfaces
      1.2 ms  DataAPI
      0.1 ms  IteratorInterfaceExtensions
      0.1 ms  TableTraits
     22.8 ms  Tables
      7.1 ms  Missings
      0.3 ms  Requires
    112.5 ms  CategoricalArrays 16.61% compilation time (6% recompilation)
     83.7 ms  FixedPointNumbers
     73.2 ms  ColorTypes 6.43% compilation time
      0.7 ms  Formatting
     75.3 ms  StringManipulation
    100.1 ms  Crayons
      0.9 ms  LaTeXStrings
    228.8 ms  PrettyTables
      0.8 ms  Compat
     75.7 ms  DataStructures
      0.4 ms  SortingAlgorithms
      2.5 ms  DocStringExtensions 67.38% compilation time
      6.7 ms  IrrationalConstants
     73.2 ms  ChainRulesCore
      0.7 ms  ChangesOfVariables
      1.1 ms  InverseFunctions
      0.7 ms  LogExpFunctions
      0.3 ms  StatsAPI
     23.1 ms  StatsBase
     23.0 ms  PDMats
      1.3 ms  OpenLibm_jll
     27.3 ms  Preferences
[...]

PR #898, opened 2023-04-20
Bump compat LossFunctions = "0.9" and address breakage

### LINKED ISSUE #896: Update to LossFunctions.jl v0.9
@ablaom we refactored LossFunctions.jl and changed the order of output and prediction arguments to be more consistent with other ecosystems as you requested at some point in the past. Can you please update this package to use the latest release?

We also removed the ObsDim submodule and the OrdinalMarginLoss as they were not very useful nor very used in downstream projects we maintain.

ablaom:
Closes: #896 #570


src/measures/loss_functions_interface.jl
```diff
@@ -44,7 +44,7 @@ err_wrap(n) = ArgumentError("Bad @wrap syntax: $n. ")
 
 # We define amacro to wrap a concrete `LossFunctions.SupervisedLoss`
 # type and define its constructor, and to define property access in
-# case of paramters; the macro also defined calling behaviour:
+# case of paramters; the macro also defines calling behaviour:
```
ToucheSir, line 47:
```suggestion
# case of parameters; the macro also defines calling behaviour:
```
Drive-by comment since I've been following https://github.com/JuliaAI/MLJBase.jl/issues/570 and am interested in how to effectively use the new LossFunctions interface :)

ablaom:
Thanks @ToucheSir . FYI I am working on a major refactor of the MLJMeasures code, which will live in a separate package, which will better leverage MLUtils.jl.  I'm considering dropping LossFunctions.jl altogether there (see [this issue](https://github.com/JuliaML/LossFunctions.jl/issues/157)) and to just re-implement those functions. Given the infrastructure I have developed, this would be pretty simple, and I'll get multi-target versions of the loss functions for free (working on matrices, tables or arrays.)  I'll also make the basic type-free scalar functions public.

PR #900, opened 2023-04-24
Add precompile workload

rikhuijzer:
In response to the comments at https://github.com/alan-turing-institute/MLJ.jl/pull/1016, this PR suggests adding a precompile workload here to `MLJBase.jl`.

I'm running a bit into an issue though. I can't run `evaluate!` because that one calls `MLJModelInterface.nrows` at `src/resampling.jl#1289`. As far as I understand, the problem is that the interface mode in `MLJModelInterface` (`INTERFACE_MODE`) is not set during precompilation yet. However, I also cannot set the global constant because that will likely "bake in" the constant during precompilation via constant propagation, which then could give problems later on.

Is there a way we could get rid of the global constant? Maybe that could make precompilation easier. This might also have some other benefits down the line. I do understand that it's a big ask, though, so we can also continue without calling `evaluate!`.


ablaom:
Thanks for this. Let me give a little context to see if that helps you see a solution. The idea that `INTERFACE_MODE` is a switch to allow new model implementations to avoid certain package dependencies (such as Tables.jl) when they depend on MLJModelInterface.jl. (We had tried a Requires.jl solution at the time but got issues.) As we document [here](https://alan-turing-institute.github.io/MLJ.jl/dev/adding_models_for_general_use/#Important), these implementations don't actually function until MLJBase is imported, which changes the value of the constant.

> Is there a way we could get rid of the global constant? 

Unfortunately, I don't see a way to do this that isn't disruptive. I guess we could use the new optional dependencies feature of 1.9, but that would still be quite a bit of refactoring and would be breaking. I'm not sure breaking changes to MLJModelInterface.jl are even conceivable at this point. That's for MLJ2 ha ha.

A key point is that this: once MLJBase is imported, there is no reason for INTERFACE_MODE to have any value other than `FullInterface`. 

Currently the switch happens in MLJBase.jl/init.jl. Is it possible we move it out of init.jl? I have gotten into trouble trying to modify global constants in other packages outside of init.jl, so I could be missing something here.

ablaom:
Thanks @rikhuijzer for spending time looking into this. Appreciate it.

PR #909, opened 2023-05-25
Replace built-in measures with measures in StatisticalMeasures.jl

### LINKED ISSUE #502: Add measures for multi-targets
I suppose there are two types of composition here - homogeneous and inhomogeneous, which might be implemented differently:

- homogeneous: all columns of target have the same scitype and we are we are defining our measure as a n-product of a single fixed uni-target measure `m`. Needed for time-series. 

- inhomogeneous: any two measures (multi-target or uni-target), one can take their product `m1 x m2`, to get a new multi-target measure .

We should keep in mind that multi-target predictions are tables.

### LINKED ISSUE #529: Improve querying for measures
There used to be `measures(matching(y))` but that didn't really make sense because `y` doesn't tell you whether you want to running a probabilistic model or a deterministic one. @f0lie suggests basing the query on a ready-constructed machine, which makes more sense. See https://github.com/alan-turing-institute/MLJ.jl/issues/753#issuecomment-805378211 .

Maybe something like `measures(matching(mach))`??

### LINKED ISSUE #570: Investigate loading time of LossFunctions.jl

ablaom:
In this PR we:

- (**breaking**) Remove the definition of all measures (`mae`, `log_loss`, etc), which can instead be loaded from [StatisticalMeasures.jl](https://juliaai.github.io/StatisticalMeasures.jl/dev/), which provides like-named measures, with the exception of those previously wrapped from LossFunctions.jl, which now must be explicitly imported and wrapped. (Julia >= 1.9). Breaking behaviour and workarounds are detailed below. For Julia < 1.9, make StatisticalMeasures a hard dependency and re-export it's names. For Julia 1.9 provide default measures for models using a Pkg extension, including StatisticalMeasures as a weak dependency. 

- Remove the LossFunctions.jl dependency to close #570.

- Add StatisticalMeasures.jl as a weak dependency, so that default measures can be provided through an Pkg extension, ext/DefaultMeasuresExt.jl. 

- Adapt the resampling (`evaluate!`) code (and a semi-duplication for model `Stack`s) to the new measure API set out in [StatisticalMeasuresBase.jl](https://juliaai.github.io/StatisticalMeasuresBase.jl/dev/). 

This PR closes #502, as StatisticalMeasures.jl provides some out-of-the-box multi-target measures and StatisticalMeasuresBase.jl provides a wrapper `multimeasure` to get more. Similarly, closes #529.

## Breaking behaviour relevant to many users

- In Julia 1.9 or higher, StatisticalMeasures.jl must be explicitly imported to use measures that were previously part of MLJBase (but new releases of MLJ will import StatisticalMeasures.jl, so they will be available that umbrella pkg).

-  Measures that in MLJBase skipped `NaN` values will now (at least by default) propagate those values, when imported from StatsticalMeasures.jl. Missing value behaviour is unchanged, except some measures that previously did not support `missing` now do. 

- All measures return a *single* aggregated measurement. In other words, measures
  previously reporting a measurement *per-observation* (previously subtyping
  `Unaggregated`) no longer do so. To get per-observation measurements, use the new method
  `measurements(measure, ŷ, y[, weights, class_weights])` from StatisticalMeasures.jl.
  
- Measures with a "feature argument" `X`, as in `some_measure(ŷ, y, X)`, are no longer
  supported. See [What is a
  measure?](https://juliaai.github.io/StatisticalMeasuresBase.jl/dev/implementing_new_measures/#definitions)
  for allowed signatures in measures.

- Aliases for measure types have been removed. For example `RMSE` (alias for
  `RootMeanSquaredError`) is gone. Aliases for instances, such as `rms` and
  `cross_entropy` have been preserved. The exception is `precision`, for which `ppv` can
  be used in its place. (This is to avoid conflict with `Base.precision`, which was
  previously pirated; you can also do `StatisticalMeasures.precision`.)

- `info(measure)` has been decommissioned; query docstrings or access the new [measure
  traits](https://juliaai.github.io/StatisticalMeasuresBase.jl/dev/methods/#Traits)
  individually instead. These traits are now provided by StatisticalMeasures.jl and not
  re-exported.

- Behaviour of the `measures()` method, to list all measures and associated traits, has
  changed. It now returns a dictionary instead of a vector of named tuples;
  `measures(predicate)` is decommissioned, but `measures(needle)` is preserved. (This method, 
  owned by StatisticalMeasures.jl has some new search options.)
  
- Measures that were wraps of losses from LossFunctions.jl are no longer exposed by
  MLJBase.jl. To use such a loss, you must explicitly `import LossFunctions` and wrap the
  loss appropriately.  See [Using losses from
  LossFunctions.jl](https://juliaai.github.io/StatisticalMeasures.jl/dev/examples_of_usage/#Using-losses-from-LossFunctions.jl)
  for examples.

- Some user-defined measures working in previous versions of MLJBase.jl may not work
  without modification, as they must conform to the new [StatisticalMeasuresBase.jl
  API](https://juliaai.github.io/StatisticalMeasuresBase.jl/dev/implementing_new_measures/#definitions). See
  [this tutorial](https://juliaai.github.io/StatisticalMeasuresBase.jl/dev/tutorial/) on
  how define new measures.

- The default measure for regression models (used in `evaluate/evaluate!` when `measures` is unspecified) is changed from  `rms` to `l2=LPLoss(2)` (mean sum of squares). 

## Breaking behaviour likely relevant only to developers of some client packages

- The abstract measure types `Aggregated`, `Unaggregated`, `Measure` have been
  decommissioned. (A measure is now defined purely by its [calling
  behaviour](https://juliaai.github.io/StatisticalMeasuresBase.jl/dev/implementing_new_measures/#definitions).)

- What were previously exported as measure types are now only constructors.

- `target_scitype(measure)` is decommissioned. Related is
  `StatisticalMeasures.observation_scitype` which declares an upper bound on the allowed
  scitype of a single observation
  
- `prediction_type(measure)` is decommissioned. Instead use
  `StatisticalMeasures.kind_of_proxy`.

- The trait `reports_each_observation` is decommissioned. Related is
  `StatisticalMeasures.can_report_unaggregated` which is `false` when the new
  `measurements` method returns `n` copies of the aggregated measure, where `n` is the
  number of observations provided, instead of individual observation-dependent
  measurements.

- `aggregation(measure)` has been decommissioned. Instead use
  `StatisticalMeasures.external_mode_of_aggregation(measure)`.

- `instances(measure)` has been decommissioned; query docstrings for measure aliases, or
  follow this example: `aliases = measures()[RootMeanSquaredError].aliases`.

- `is_feature_dependent(measure)` has been decommissioned. Measures consuming feature data are not longer
  supported; see above.

- `distribution_type(measure)` has been decommissioned.

- `docstring(measure)` has been decommissioned.

- Behaviour of `aggregate` has changed.

- The following traits, previously exported by MLJBase, cannot be applied to measures:
  `supports_weights`, `supports_class_weights`, `orientation`, `human_name`. Instead use
  the traits with these names provided by StatisticalMeausures.jl (they will need to be
  qualified, as in `using StatisticalMeasures; StatisticalMeasures.orientation(measure)`).


ablaom:
I'm not expecting detailed review of this PR, but think someone should at least look over the planned breakages, and so forth.
@rikhuijzer, @OkonSamuel could you take a cursory look?

src/MLJBase.jl
```diff
@@ -355,28 +302,19 @@ export ResamplingStrategy, Holdout, CV, StratifiedCV, TimeSeriesCV,
...
 # -------------------------------------------------------------------
 # re-export from Random, StatsBase, Statistics, Distributions,
 # OrderedCollections, CategoricalArrays, InvertedIndices:
 export pdf, sampler, mode, median, mean, shuffle!, categorical, shuffle,
    levels, levels!, std, Not, support, logpdf, LittleDict
 
+# for julia < 1.9
+if !isdefined(Base, :get_extension)
+    include(joinpath("..","ext", "DefaultMeasuresExt.jl"))
+    @reexport using .DefaultMeasuresExt.StatisticalMeasures
```
rikhuijzer, line 317: So `StatisticalMeasures` is only available in the public (re-exported) scope in Julia 1.9 or does the new extensions system from Julia automatically bring things to the public scope? (I'm not familiar with the new system yet)

src/composition/models/stacking.jl
```diff
@@ -378,21 +378,30 @@ model_2, ...), ...)
...
+
+    # For each model we record the results mimicking the fields of PerformanceEvaluation
     results = NamedTuple{modelnames}(
         [(
             measure = stack.measures,
             measurement = Vector{Any}(undef, n_measures),
             operation = _actual_operations(nothing, stack.measures, model, verbosity),
             per_fold = [Vector{Any}(undef, nfolds) for _ in 1:n_measures],
-            per_observation = Vector{Union{Missing, Vector{Any}}}(missing, n_measures),
+            per_observation = [Vector{Vector{Any}}(undef, nfolds) for _ in 1:n_measures],
```
rikhuijzer, line 404: Just to double check, `per_observation` is now a `Vector{Vector{Vector{Any}}}`?

rikhuijzer (commented):
I don't understand all the nuances of this large change, but here are some extra thoughts:

- Due to the large amount of changes, is this a nice moment to go to v1.0.0? Versions of 1 and up have the benefit that there is more control due to MAJOR.MINOR.PATCH instead of MAJOR.MINOR/PATCH. Also there is a risk of being mocked at https://0ver.org/ when staying at <1.0.0 :stuck_out_tongue: 
- Should there also be a migration guide somewhere like in the docs? The PR description would be reasonable, but also contains non-essential information making it harder for clients to migrate.

rikhuijzer:
Overall this PR looks like an improvement. The main idea was to make `StatisticalMeasures.jl` easier available to other packages, right? Maybe also good to clarify for users.

In general I'm in favor of merging this. It looks like a clear improvement.

src/MLJBase.jl
ablaom: If `using MLJBase` without MLJ, then, in Julia 1.9 or higher, `StatisticalMeasures` must be explicitly imported to use measures that were previously part of MLJBase (the names are automatically re-exported by a pkg extension). For earlier of Julia versions, StatisticalMeasures.jl is a hard dependency (through a hack that is standard practice).  If `using MLJ`, then all previous measures are still available, because StatisticalMeasures.jl will be a hard dep of MLJ.

src/composition/models/stacking.jl
ablaom: Yeah, `per_observation[i][j][k]` is the the measurement for the kth observation, in the jth fold, for the ith measure. There no longer can be `missing`s because measures that don't compute per-observation measurements, simply report copies of the aggregated instead.

ablaom:
> Due to the large amount of changes, is this a nice moment to go to v1.0.0? Versions of 1 and up have the benefit that there is more control due to MAJOR.MINOR.PATCH instead of MAJOR.MINOR/PATCH. Also there is a risk of being mocked at https://0ver.org/ when staying at <1.0.0 😛

Version 1.0.0 it is. 

> Should there also be a migration guide somewhere like in the docs? The PR description would be reasonable, but also contains non-essential information making it harder for clients to migrate.

Good idea. https://github.com/alan-turing-institute/MLJ.jl/blob/measure/docs/src/performance_measures.md#migration-guide-for-changes-to-measures-in-mljbase-10

PR #912, opened 2023-07-01
Implementing a logging interface for MLFlow

pebeto:
In support of [this proposal](https://github.com/alan-turing-institute/MLJ.jl/issues/1029#issuecomment-1617093086), this PR shows the possibility to adapt a logging platform to MLJBase functionality. This completely modular implementation allows the creation and addition of multiple logging interfaces in the future by the use of [package extensions](https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)). This work includes a way to connect MLJ with [MLFlow](https://mlflow.org/),  and opens new opportunities with other platforms (i.e. Neptune, SageMaker, etc...). The changes will be listed below:

- [x] A general package extension named **LoggersExt**, that allows the possibility to extend main methods functionality and loading desired behaviors dynamically
- [x] A minimal constructor named **MLFlowLogger** inside MLJBase source that will be extended when [MLFlowClient](https://github.com/JuliaAI/MLFlowClient.jl) is loaded, bringing a **MLFlowInstance** struct with the configurations needed by the user.
- [x] A `save` function extension that logs the model hyperparameters and saves it into the experiment artifact location.
- [x] An `evaluate!` function extension that saves the model hyperparameters and measures. This implementation is a mockup that **will be changed** in the future.
- [x] A new way to flat parameters from a NamedTuple, preserving names as key trees.
```julia-repl
julia> t = (X = (x = 1, y = 2), Y = 3)
julia> flat_params(t)
LittleDict{...} with 3 entries:
"X_x" => 1
"X_y" => 2
"Y"   => 3
```
- [ ] Save resampling strategy and it's parameters (as run tags)
- [ ] Adapt the functionality for composite models
- [ ] Adapt the functionality for tuned models
- [ ] Adapt the functionality for iterative models

### Additional information:
- [MLJ repo proposal](https://github.com/alan-turing-institute/MLJ.jl/issues/1029)
- [MLFlow documentation](https://mlflow.org/docs/latest/index.html)


ablaom:
@pebeto Great work!

> A new way to flat parameters from a NamedTuple, preserving names as key trees.

We should adapt this to model structs, right? Or are you planning (nested) model structs -> nested named tuples -> flat named tuples? I suggest *double* underscores for the compostite names as model property names often have single underscores in them already.

src/resampling.jl
```diff
@@ -917,8 +918,12 @@ untouched.
 
 - `verbosity` level, an integer defaulting to 1.
 
-- `check_measure` - default is `true`
+- `check_measure` - default is `true`.
 
+- `logger` - Binds a logging platform. It receives an instance object with
+  the required client parameters. Default is `nothing`.
+
+See [Loggers](@ref) for logger details.
```
ablaom, line 926:
```suggestion
  the required client parameters. Default is `nothing`. See [Loggers](@ref) for logger details.
```

Also, I tried `?Loggers` and didn't get anything.  And `MLJBase.Loggers` does not exist.  I guess this is coming?

src/loggers.jl
```diff
@@ -0,0 +1,29 @@
...
+[`MLFlowClient.jl`](https://juliaai.github.io/MLFlowClient.jl/dev/), and logs
+to an experiment.
+
+### Summary of key-word arguments
+- `base_uri`: URI of the MLFlow server. Defaults to `localhost:5000`.
+- `experiment_name`: Name of the experiment to log to.
+- `artifact_location`: Location of the artifacts.
+
+If `experiment_name` is not provided, a new experiment with the name
+"MLJ.jl experiments" will be created.
```
ablaom, line 15: Here "experiments" is plural,  which is creating some cognitive dissonance for me, because `experiment_name' is singular, not plural. Mays "MLJ experiment" is better?

src/loggers.jl
```diff
@@ -0,0 +1,29 @@
...
+If `artifact_location` is not provided, MLFlow will use its default directory.
```
ablaom, line 16: I'm a little confused what this means: "MLFlow will use its default directory". Do you mean "MLJFlowClient.jl" has a default directory that is used, or that mlflow the application has a default that is used. It's not clear to me where I go to change the default. 

If you are not always doing so, please refer to mlflow, the application, using lowercase to distinguish between the `MLFlow` julia type.

src/loggers.jl
```diff
@@ -0,0 +1,29 @@
+"""
+    mlflow_logger(; base_uri="localhost:5000", experiment_name=missing)
+
+Constructor for MLFlowLogger. Creates an instance of MLFlow,
+as defined in
+[`MLFlowClient.jl`](https://juliaai.github.io/MLFlowClient.jl/dev/), and logs
+to an experiment.
+
+### Summary of key-word arguments
+- `base_uri`: URI of the MLFlow server. Defaults to `localhost:5000`.
```
ablaom, line 10: As you document, the keyword argument is called `base_uri`. However, in the display, shown below, it is called `baseuri`, which confused me when I couldn't remember the name:

```
julia> logger
LoggersExt.MLFlowLogger(MLFlow(
    baseuri = "http://127.0.0.1:5000", 
    apiversion = 2.0
), "MLJ experiments", missing)
```

In any case, we may want to make this compulsory positional argument instead of a kwarg. See my comment below.

ablaom:
Just starting a review of this great new feature. Let me relate my initial experience, which suggests to me some changes in the user API.

- I first looked at the `evaluate!` doc-strings to see clues of how to incorporate logging, but because the `Logger` ref is dead, I could not discover the `mlflow_logger` constructor I needed to learn about. (Found it from the code.)
- I knew (only because you have explained it to me) that I must install and launch the (python) mlflow application. I wasn't exactly sure how to launch it, but the mlflow docs seemed to suggest entering `> mlflow ui` would work, so I did that. A url appeared in the log and I successfully navigated my browser to that location, "http://127.0.0.1:5000". I know launching the application is a system specific operation; still, an concrete example of how to do this on a linux-like machine would have been helpful in the `mlflow_logger` docstring.
- I used the default constructor `logger=mlflow_logger()`, which created `baseuri = "localhost:5000"`.  I then tried to use `evaluate!` with `logger=logger` and I got a long intimidating unhelpful error. Here's a part of it:

```
ERROR: HTTP.Exceptions.StatusError(403, "GET", "/api/2.0/mlflow/experiments/get-by-name?experiment_name=MLJ%20experiments", HTTP.Messages.Response:
"""
HTTP/1.1 403 Forbidden
Content-Length: 0
Server: AirTunes/620.8.2
""")
Stacktrace:
 [1] getexperiment(mlf::MLFlow, experiment_name::String)
@ MLFlowClient ~/.julia/packages/MLFlowClient/yjzJX/src/experiments.jl:71
```

- I eventually figure out that I need to explicitly specify the `baseuri` given to me by the mlflow application and that worked (after I figured out that the kwarg name is `base_uri`, not `baseuri`). Then all was well.

Following this experience, I have some suggestions. Let me know what you think about them:

- Make `baseuri` a compulsory positional argument of `mlflow_logger`. This is less convenient if `localhost:5000` happens to work for a user, but it didn't in my case. Also, logically, since the name of server is really controlled by the mlflow app, it seems to me that logically the MLJ user needs to specify this information. Also, I think a common gotcha  for users will be not realising they need to launch mlflow manually. Having to specify the uri forces them to confront that issue.
- If it is possible, catch invalidly set `baseuri` values and throw a more helpful error message. If this is very difficult to do, then we can use `throw`/`catch` construct (always a method of last resort, in my view, but I can imagine it's hard to say what is a "valid" uri on the Julia side.)

Project.toml
```diff
@@ -54,11 +61,12 @@ DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 DecisionTree = "7806a523-6efd-50cb-b5f6-3fa6f1930dbb"
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
+MLFlowClient = "64a0f543-368b-4a9a-827a-e71edb2a0b83"
```
ablaom, line 64: You will also need to add a [compat] entry for MLFlowClient.jl, as this is requirement for all registered Julia packages: all [deps] must have a corresponding [compat] entry.

ext/LoggersExt/LoggersExt.jl
```diff
@@ -0,0 +1,12 @@
+module LoggersExt
+
+using MLJBase: info, name, Model,
+               Machine, Measure, flat_params
+
+import MLJBase: save, evaluate!, mlflow_logger
+
+include("utils.jl")
+
```
ablaom, line 9: Style: remove the spare line:

```suggestion
```

ext/LoggersExt/mlflow.jl
```diff
@@ -0,0 +1,64 @@
+using MLFlowClient: MLFlow, logparam, logmetric,
+                    createrun, MLFlowRun, updaterun,
+                    logartifact, getorcreateexperiment
+
+struct MLFlowLogger
+    client::MLFlow
+    experiment_name::String
+    artifact_location::Union{String, Missing}
```
ablaom, line 8: In Julia `missing` is used in the setting of missing data (in the sense of statistics). Here you should use `nothing` instead of `missing` for the fallback `location`. 

`missing` has some behaviour that makes it trickier to handle. For example, `x == missing` always returns `missing`, no matter what `x` is. The idea is that a `missing` values should get propagated everywhere you operate on them (eg, in arithmetic).  By contrast, very little behaviour is actually overloaded for the `Nothing` type, so it is more predictable.

ext/LoggersExt/mlflow.jl
```diff
@@ -0,0 +1,64 @@
...
+    artifact_location::Union{String, Missing}) =
+    MLFlowLogger(MLFlow(base_uri), experiment_name, artifact_location)
+
+function _logmodelparams(client::MLFlow, run::MLFlowRun, model::Model)
+    model_params = params(model) |> flat_params |> collect
+    for (name, value) in model_params
+        logparam(client, run, name, value)
+    end
+end
+
```
ablaom, line 20: Just a note that you don't need to annotate the types for these functions. It makes no difference to the compiler here. However, it does improve readability, so it's fine to leave it here 😄

ext/LoggersExt/mlflow.jl
```diff
@@ -0,0 +1,64 @@
...
+function _logmachinemeasures(client::MLFlow, run::MLFlowRun, measures::Vector{T},
+                             measurements::Vector{Float64}) where T<:Measure
```
ablaom, line 22: In this you appear to be unecessarily restricting the types, and this means less generic code. I mean, does MLFlowClient really care if a measurement is a `Float64`? If we are using a GPU, we might have `Float32` here, for example. And a measure like `true_postitive` returns an `Int`. Also, in MLJ, some measurement are not even numerical: we could have a float vector, class dictionary,  or a confusion matrix. 

Try to make this a generic as MLFlowClient will allow. If I understand rightly, MLFlowClient just calls `string` on everything. If that  is the case, I'd just leave this untyped (as in my suggestion). 

With this in mind, here are some non-standard classification measures to add in your tests: `true_positive` (integer-valued, data must be `OrderedFactor{2}`), `ConfusionMatrix()` (returns a special type wrapping a matrix), `MulticlassFScore(average=MLJBase.NoAvg())` (dictionary-valued). 

```suggestion
function _logmachinemeasures(client::MLFlow, run::MLFlowRun, measures, measurements)
```

ext/LoggersExt/utils.jl
```diff
@@ -0,0 +1,33 @@
...
+isamodel(::Model) = true
+
+"""
+    params(m::Model)
+
+Recursively convert any transparent object `m` into a named tuple,
+keyed on the property names of `m`. An object is *transparent* if
+`isamodel(m) == true`. The named tuple is possibly nested
+because `params` is recursively applied to the property values, which
+themselves might be transparent.
```
ablaom, line 11:
```suggestion
Recursively convert any object subtyping `Model` a named tuple,
keyed on the property names of `m`. The named tuple is possibly nested
because `params` is recursively applied to the property values, which
themselves might subtype `Model`.
```

ext/LoggersExt/utils.jl
```diff
@@ -0,0 +1,33 @@
...
+
+For most `Model` objects, properties are synonymous with fields, but
+this is not a hard requirement.
+
+Most objects of type `Model` are transparent.
+
```
ablaom, line 17:
```suggestion
```

src/resampling.jl
```diff
@@ -1157,6 +1163,12 @@ function measure_specific_weights(measure, weights, class_weights, test)
     return nothing
 end
 
+# Logging level that allows platforms to perform operations by multiple dispatch over `evaluate!`:
```
ablaom, line 1166:
```suggestion
# Workflow logging platforms, such as mlflow (provided by MLFlowClient.jl) overload the
# following fallback method but replace the `logger` argument with `logger::LoggerType`,
# where `LoggerType` is specific to the logging platform. Such an overloading will need to
# include a call to the `evaluate` method which does *not* include the final `logger`
# argument, namely the sole method called in the following fallback.
```

src/utilities.jl
```diff
@@ -33,6 +33,38 @@ function flat_values(params::NamedTuple)
...
+
+View a nested named tuple `t` as a tree and return, as a Dict, the key subtrees
+and the values at the leaves, in the order they appear in the original tuple.
+
+```julia-repl
+julia> t = (X = (x = 1, y = 2), Y = 3)
+julia> flat_params(t)
+LittleDict{...} with 3 entries:
+"X_x" => 1
+"X_y" => 2
```
ablaom, line 47:
```suggestion
"X__x" => 1
"X__y" => 2
```

ext/LoggersExt/mlflow.jl
```diff
@@ -0,0 +1,64 @@
...
+
+    _logmodelparams(logger.client, run, mach.model)
+    _logmachinemeasures(logger.client, run, performance_evaluation.measure,
+                        performance_evaluation.measurement)
+
+    updaterun(logger.client, run, "FINISHED")
+    return performance_evaluation
+end
+
+function save(logger::MLFlowLogger, mach::Machine)
```
ablaom, line 49: Let's have a doc-string for `save` here. It will get appended to the existing `MLJBase.save` docstring. 

By the way, how do I deserialise the artifact? I guess we haven't addressed this yet, right? A file is deserialised with `machine("file_name.jls")` but  `machine(logger::MLFlowLogger)` is not going work, because we need to point to a specific run, right?

test/extensions/loggers.jl
```diff
@@ -0,0 +1,60 @@
...
+        X = (x=rand(4),)
+        y = ["Chenta", "Missy", "Gala", "Wendy"] |> categorical
+
+        mach = machine(ConstantClassifier(), X, y)
+        fit!(mach, verbosity=0)
+
+        logger = mlflow_logger(; experiment_name=experiment_name, artifact_location=artifact_directory)
+
+        @testset "save" begin
+            run = MLJBase.save(logger, mach)
```
ablaom, line 32: Oh I see, `save` returns the appropriate `run`.

test/extensions/loggers.jl
```diff
@@ -0,0 +1,60 @@
...
+            experiment = getexperiment(logger.client, run.info.experiment_id)
```
ablaom, line 33: These tests are getting me thinking about how the user should do this. 

I accept that we can assume the user is  familiar with MLFlowClient.jl API, so `run.info.experiment_id` is okay. 
But she does not know to get inside `logger`, as in `logger.client`? So,  perhaps  we need to add and document a method
`client(logger::MLFlowLogger) = logger.client`? And any other access methods the user might need. 

Or, maybe better is to overload `MLFlowClient.getexperiment` in this way:

```julia 
"""
    getexperiment(logger::MLFlowLogger, run)

Calls `getexperiment(client, run)` where `client` is the client associated with `logger`.
"""
MLFlowClient.getexperiment(logger::MLFlowLogger, run) = MLFlowClient.getexperiment(logger.client, run)
```

These comments obviously apply to other MLFlowClient methods.

test/extensions/loggers.jl
```diff
@@ -0,0 +1,60 @@
...
+
+        @test logger.client isa MLFlow
+        @test logger.experiment_name == experiment_name
+        @test logger.artifact_location == artifact_directory
+    end # @testset "outside extension tests"
+
+    @testset "extension tests" begin
+        X = (x=rand(4),)
+        y = ["Chenta", "Missy", "Gala", "Wendy"] |> categorical
+
```
ablaom, line 25: As discussed in another comment above, we should add tests out some measures with different value types. 

I suggest making `y` an `OrderedFactor{2}` so that `true_positive` can be tested. So you can do, e.g., 

```julia
X, y = make_moons()
y = coerce(y, OrderedFactor)
```
to get some ordered factor binary data.

ablaom (commented):
Okay, was great to play around with this new feature. Very cool. Thanks a great start on this project! 🎉 

In addition to addressing my inline comments, can we please have:

- [x] A unit test for `flat_params`
- [x] Some tests on the client side of:
  - measures and measurements
  - parameters
  You may want to upgrade your `ConstantClassifier` to something more complicated like, `model = Standardizer() |> DecisionTreeClassifier()` to get some nested parameters to check. I'm pretty sure the testing module `Models` provides these two components.
- [x] The refactor of `Resampler` that we discussed in our July 18th call. 
- [ ]  serialisation and deserialization of a machine as artifact.

And I think we need to have a call to discuss that last one.

src/loggers.jl
pebeto: I don't see this as an option due to someone can run this with default arguments. This baseuri is just a mirror of the default uri assigned by the mlflow project. This is not related about any Julia library, but it's important to be defined.

ext/LoggersExt/mlflow.jl
pebeto: You're right. The only way is to search for that specific run and retrieve the desired artifact using MLFlowClient.jl functionality.

test/extensions/loggers.jl
pebeto: Our MLFlowLogger object will be used just to retrieve the related experiments. Crossing that line, we are just going to create auxiliary functions (like `client(logger::MLFlowLogger) = logger.client` that simplifies the calls from MLFlowClient. I don't think this could be a good idea because, by the nature of package extensions, we will import MLFlowClient and the end-user will be able to call anything from that package.

pebeto:
@ablaom, I'm moving the extension into an external package that will be a dependency of [MLJ.jl](https://github.com/alan-turing-institute/MLJ.jl) instead of MLJBase. [MLJFlow.jl](https://github.com/pebeto/MLJFlow.jl) will contain all the required logic related to the connection between all the MLJ dependencies and MLFlow (via [MLFlowClient.jl](https://github.com/JuliaAI/MLFlowClient.jl)).

src/MLJBase.jl
```diff
@@ -349,7 +349,7 @@ export TransformedTargetModel
 
 # resampling.jl:
 export ResamplingStrategy, Holdout, CV, StratifiedCV, TimeSeriesCV,
-    evaluate!, Resampler, PerformanceEvaluation
+    evaluate!, Resampler, PerformanceEvaluation, log_evaluation
```
ablaom, line 352: Let's not export `log_evaluation`; this is not for the general user. You package MLJFlow.jl will just need to explicitly import this function or explicitly reference it as `MLJBase.log_evaluation`.

src/resampling.jl
```diff
@@ -476,6 +476,9 @@ struct.
 
 - `measure`: vector of measures (metrics) used to evaluate performance
 
+- `model`: model used to create the performance evaluation. In the case a
```
ablaom, line 479: The order of fields is not idea, as I think it creates some cognitive dissonance to separate `measure` from `measurement`. Can I suggest putting `model` last or first?

src/resampling.jl
```diff
@@ -509,13 +512,15 @@ struct.
...
 struct PerformanceEvaluation{M,
+                             Model,
                              Measurement,
                              Operation,
                              PerFold,
                              PerObservation,
                              FittedParamsPerFold,
                              ReportPerFold} <: MLJType
     measure::M
+    model::Model
```
ablaom, line 523: Nit-picky suggestion. We frequently use `M` for a model type elsewhere. For consistency, can we do something like:

```suggestion
struct PerformanceEvaluation{Measure,
                             M,
                             Measurement,
                             Operation,
                             PerFold,
                             PerObservation,
                             FittedParamsPerFold,
                             ReportPerFold} <: MLJType
    measure::Measure
    model::M
```

src/resampling.jl
```diff
@@ -1157,6 +1161,22 @@ function measure_specific_weights(measure, weights, class_weights, test)
...
+"""
+    log_evaluation(logger, performance_evaluation)
+
+Logs a performance evaluation (the returning object from `evaluate!`)
+to a logging platform. The default implementation does nothing.
+Can be overloaded by logging interfaces, such as MLJFlow (MLFlow
+connection via MLFlowClient.jl), which replace the `logger` argument
+with `logger::LoggerType`, where `LoggerType` is specific to the logging
+platform.
+"""
```
ablaom, line 1177:
```suggestion
"""
    log_evaluation(logger, performance_evaluation)

Log a performance evaluation to `logger`, an object specific
to some logging platform, such as mlflow. If `logger=nothing`
then no logging is performed. The method is called at
the end of every call to `evaluate/evaluate!` using the
logger provided by the `logger` keyword argument.

# Implementations for new logging platforms

Julia interfaces to workflow logging platforms, such as
mlflow (provided by the MLFlowClient.jl interface) should
overload `log_evaluation(logger::LoggerType, performance_
evaluation)`, where `LoggerType` is a platform-specific type
for logger objects. For an example, see the implementation
provided by the MLJFlow.jl package.
"""
```

ablaom (commented):
I guess you're still working on this. I don't see a call to `log_evaluation` in any of the `evaluate!` methods, as discussed on slack, and copied below:

```julia
# In MLJBase:
log_evaluation(logger::Nothing, e) = nothing # (publicly extendable method)
function evaluate!(..., logger)
    <create evaluation object `e`>
    log_evaluation(logger, e)
end

# In MLJFlow.jl:
struct MLFlowLogger
    ...
end

MLJBase.log_evaluation(logger::MLFlowLogger, e)
    <log to the client in `logger`; you can find everything you need in `e`>
end
```

ablaom:
Also, I no longer see the `logger`  kwarg additions to `evaluate`/`evaluate!` (with the `nothing` fallback values). 

And I don't see any refactoring of `Resampler`, as previously discussed.

ablaom (commented):
Looking good.  We still have the refactoring of `Resampler` to go, yes?

pebeto:
This implementation has changed over time. Closing this PR in favor of JuliaAI/MLJBase.jl#925

PR #951, opened 2024-01-08
Forward `(::Machine).fitresult` to RecipesBase

MilesCranmer:
@ablaom

Apparently this is the idiomatic way to set up plotting recipes on custom types: https://docs.juliaplots.org/latest/RecipesBase/api/#RecipesBase.@recipe-Tuple{Expr}

With this change, a downstream package will be able to define:

```julia
@recipe function plot_my_regressor(fitresult::MyFitresultType)
    ...
end
```

and enable custom plotting of their fitted model.

---

Confirmed that this works for SymbolicRegression's regressors!

---

Fixes https://github.com/JuliaAI/MLJModelInterface.jl/issues/188


ablaom (approved):
Looks good. Thanks for this contribution, @MilesCranmer.

Issue #948, opened 2023-10-23
Reduce sigdigits in parameter range display


ablaom:
```julia
julia> range(Float64, :junk, lower=sqrt(2), upper=sqrt(3))
NumericRange(1.4142135623730951 ≤ junk ≤ 1.7320508075688772; origin=1.5731321849709863, unit=0.15891862259789102)
```
I think 4 sigdigits is sufficient.

The display is defined here: https://github.com/JuliaAI/MLJBase.jl/blob/a8e599f302a7233b25973274cc2b5638d604c98c/src/hyperparam/one_dimensional_ranges.jl#L48

adarshpalaskar1:
Hello, can I work on this issue?

I think this change should work:

https://github.com/JuliaAI/MLJBase.jl/blob/a8e599f302a7233b25973274cc2b5638d604c98c/src/hyperparam/one_dimensional_ranges.jl#L33-L34
```
#replacing line 34 with
value = round(getproperty(r, field), sigdigits=4)
```

New output:
```
julia> range(Float64, :junk, lower=sqrt(2), upper=sqrt(3))
NumericRange(1.414 ≤ junk ≤ 1.732; origin=1.573, unit=0.1589)
```

ablaom:
Looks good to me, thanks. PR welcome.

ablaom:
issue assigned to @adarshpalaskar1

ablaom:
Closed by #958

PR #958, opened 2024-01-30
Reduce sigdigits in parameter range display #948

adarshpalaskar1:
This commit addresses issue #948 by updating the display of parameter ranges to include reduced significant digits.

### DIFF
--- a/src/hyperparam/one_dimensional_ranges.jl
+++ b/src/hyperparam/one_dimensional_ranges.jl
@@ -32,7 +32,7 @@ end
 # throws an exception:
 function _repr(r::NumericRange{T}, field) where T
     value = getproperty(r, field)
-    if !(typeof(value) <: Integer)
+    if !(value isa Integer)
         value = round(value, sigdigits=4)
     end
     r.scale isa Symbol && return repr(value)
--- a/src/hyperparam/one_dimensional_ranges.jl
+++ b/src/hyperparam/one_dimensional_ranges.jl
@@ -31,7 +31,7 @@ end
 # case that `r.scale` is not a symbol, and returning "?" if applying the transformation
 # throws an exception:
 function _repr(r::NumericRange{T}, field) where T
-    value = getproperty(r, field)
+    value = round(getproperty(r, field), sigdigits=4)
     r.scale isa Symbol && return repr(value)
     return try
         scaled = (r.scale)(value)
--- a/src/hyperparam/one_dimensional_ranges.jl
+++ b/src/hyperparam/one_dimensional_ranges.jl
@@ -31,7 +31,10 @@ end
 # case that `r.scale` is not a symbol, and returning "?" if applying the transformation
 # throws an exception:
 function _repr(r::NumericRange{T}, field) where T
-    value = round(getproperty(r, field), sigdigits=4)
+    value = getproperty(r, field)
+    if !(typeof(value) <: Integer)
+        value = round(value, sigdigits=4)
+    end
     r.scale isa Symbol && return repr(value)
     return try
         scaled = (r.scale)(value)
--- a/test/hyperparam/one_dimensional_ranges.jl
+++ b/test/hyperparam/one_dimensional_ranges.jl
@@ -120,10 +120,13 @@ end
     io = IOBuffer()
     r1 = range(Int, :junk, lower=1, upper=10)
     r2 = range(Char, :junk, values=['c', 'd'])
+    r3 = range(Float64, :junk, lower=3.14159, upper=6.283185)
     show(io, r1)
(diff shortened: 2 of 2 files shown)


ablaom:
Looks like you need to update a test.

adarshpalaskar1:
Yes, I saw the logs and found this test is failing:
```
range(Int, :junk, lower=1, upper=10)
```

Log:
```
  Expression: String(take!(io)) == "NumericRange(1 ≤ junk ≤ 10; origin=5.5, unit=4.5)"
   Evaluated: "NumericRange(1.0 ≤ junk ≤ 10.0; origin=5.5, unit=4.5)" == "NumericRange(1 ≤ junk ≤ 10; origin=5.5, unit=4.5)"
```

Should I update the test, or is the updated code incorrect?

adarshpalaskar1:
If the issue is only in the case of integers can we add a check for excluding integers?
```
function _repr(r::NumericRange{T}, field) where T
    value = getproperty(r, field)
    if !(T <: Integer)
        value = round(value, sigdigits=4)
    end
    r.scale isa Symbol && return repr(value)
```

ablaom:
I think your changes are good. You just need to update the invalidated test. How about  you improve the test by using lower and upper values with more that 4 sigdigits and verifying (with a variation on the existing test) that extra digits get truncated?

ablaom:
Okay, I see. In the <:Integer case, we don't want to apply rounding at all. Sorry I missed that. So you need to add some logic.

adarshpalaskar1:
Okay, I have added a check before rounding off. I faced some errors while executing the tests locally so tried some test cases manually.

src/hyperparam/one_dimensional_ranges.jl
```diff
@@ -32,6 +32,9 @@ end
 # throws an exception:
 function _repr(r::NumericRange{T}, field) where T
     value = getproperty(r, field)
+    if !(typeof(value) <: Integer)
```
ablaom, line 35: More idiomatic is

```suggestion
    if !(value isa Integer)
```

ablaom:
Can we please add at test here:

https://github.com/JuliaAI/MLJBase.jl/blob/f3da249c11374630e242fc25feed29801d3bf75a/test/hyperparam/one_dimensional_ranges.jl#L122 ?

For example, something for `r3= range(Float64, :junk, lower=3.14159, upper=6.283185)`

src/hyperparam/one_dimensional_ranges.jl
adarshpalaskar1: Thank you for the review, I have committed the suggested changes.

adarshpalaskar1:
> Can we please add at test here:
> 
> https://github.com/JuliaAI/MLJBase.jl/blob/f3da249c11374630e242fc25feed29801d3bf75a/test/hyperparam/one_dimensional_ranges.jl#L122
> 
> ?
> For example, something for `r3= range(Float64, :junk, lower=3.14159, upper=6.283185)`

Sure, I have added the test and executed it locally.
```
julia> include("test\\hyperparam\\one_dimensional_ranges.jl")
Test Summary: | Pass  Total  Time
constructors  |   36     36  1.7s
Test Summary:                            | Pass  Total  Time
range constructors for nested parameters |    1      1  0.6s
Test Summary:       | Pass  Total  Time
Test Summary: | Pass  Total  Time
coverage      |    3      3  0.2s
true
```

ablaom (approved):
Perfect. 

Thanks @adarshpalaskar1 for your valuable contribution.

adarshpalaskar1:
Thank you @ablaom for helping me through my beginner doubts; this was my first contribution. If there's anything specific I should keep in mind for future contributions, and if there are more issues where I can help, please point me in the right direction.

ablaom:
Maybe you'd like to take a stab at https://github.com/JuliaAI/MLJDecisionTreeInterface.jl/issues/23 ?

ablaom:
You can find some decision tree examples at https://alan-turing-institute.github.io/MLJ.jl/dev/models/DecisionTreeClassifier_DecisionTree/#Examples .

Once we have that, you'd be in good shape to improve integration with [TreePlots.jl](https://github.com/JuliaAI/TreeRecipe.jl).

ablaom:
Congratulations, by the way, on your first contribution 🎈

adarshpalaskar1:
> You can find some decision tree examples at https://alan-turing-institute.github.io/MLJ.jl/dev/models/DecisionTreeClassifier_DecisionTree/#Examples .
> 
> Once we have that, you'd be in good shape to improve integration with [TreePlots.jl](https://github.com/JuliaAI/TreeRecipe.jl).

Thank you! I'll check out the decision tree examples and proceed.

PR #963, opened 2024-03-15
Allow `Pipeline` and `TransformedTargetModel` to support feature importances

ablaom:
A model that supports features importances exposes these via the method `feature_importances`. However, these importances are hidden if the model is wrapped in `TransformedTargetModel` or is part of a pipeline. This PR implements `feature_importances` (and the trait `reports_feature_importances`) for these two wrappers by forwarding the atomic methods. 

For testing purposes, the test model `DecisionTreeClassifier` in /test/_models/DecisionTree.jl has been updated to support feature importances. 

@OkonSamuel


OkonSamuel (approved):
Looks good. This should resolve https://github.com/alan-turing-institute/MLJ.jl/issues/1100

PR #968, opened 2024-03-30
Add prompt to docstring REPL example

abhro:
Add "julia> " to example REPL session to clearly indicate code. Also add semicolon to indicate that the output should not be considered.


ablaom (approved):
Thanks @abhro

Issue #959, opened 2024-02-18
Type instabilities in `predict`


MilesCranmer:
Due to the definition of `last_model` being an `Any`:

```julia
last_model(mach) = isdefined(mach, :old_model) ? mach.old_model : nothing
```

all of the operations in `OPERATIONS` cannot infer output type. This is okay for `report`, but for `predict`, the output type might be important for downstream operations, and so this can slow things down.

ablaom:
Nice observation. May be tricky to fix quickly. I'm curious how you came to identify the issue. Do you have a concrete use case with performance substantially compromised by this issue?

MilesCranmer:
Sure, here is a 50x slowdown with a really simple function. But keep in mind that this is just the most minimal example. Ultimately this type instability will infect any code it touches. The larger the codebase around each `predict` call, the worse the performance hit.

```julia
using SymbolicRegression, MLJBase, Random

rng = Random.MersenneTwister(0)
X = randn(rng, 30, 3)
y = @. cos(X[:, 1] * 2.1 - 0.9) * X[:, 2] - X[:, 3]

model = SRRegressor(deterministic=true, seed=0, parallelism=:serial, maxsize=10)
mach = machine(model, X, y)
fit!(mach)

# With predict(mach, X)
function f(mach, X)
    y = predict(mach, X)
    return mean(y), std(y)
end

# With low-level calling
function g(eqn, options, X)
    y, _ = eval_tree_array(eqn, X, options)
    return mean(y), std(y)
end
```

With the following performance comparison

```julia
@btime f($mach, $X)
#   5.521 μs (259 allocations: 18.79 KiB)
@btime g(eqn, options, $(copy(X'))) setup=(r=report(mach); eqn = r.equations[r.best_idx]; options=mach.fitresult.options)
#  134.497 ns (3 allocations: 368 bytes)
```

MilesCranmer:
Here is the Cthulhu.jl output when I descend into `f(mach, X)` – the `predict(mach, X)` call results in `::Any` output (in red):
<img width="1492" alt="Screenshot 2024-03-06 at 21 02 08" src="https://github.com/JuliaAI/MLJBase.jl/assets/7593028/af75b91c-c23e-4cf5-92c1-55cc626535a6">

Descending into the `predict` call, we can immediately see that the first appearance of an `::Any` is from accessing the `.fitresult`:

[...]

PR #969, opened 2024-04-08
Address some predict/transform type instabilities

ablaom:
This PR addresses a type instability for operations (`predict`, `transform`, etc) acting on machines, as  identified in #959 (although this PR does not resolve the particular issue there). 

I admit it is not clear to me that the performance gains here are likely to significantly benefit many use cases. But having done the work to identify these instabilities, I don't see much harm in including them. (Incidentally, the table below identifies some likely type instabilities *internal* to some third party implementations of the MLJ interface.)

The type instability is not difficult to address in the case of machines attached to ordinary models, by annotating a currently abstract type in the `Machine` struct. However, in the special case of a  machine attached to a *symbolic* model (which appear exclusively in learning networks), the type instability remains.

In 69 regression models, we compared the "high level"  `predict(::Machine, ...)` method with the "low level" `predict(::Model, ...)` method implemented by third party model providers. The benchmark code is hidden below:

<details>

```
using MLJTestInterface, MLJModels, MLJBase
using Tables
using Random
using BenchmarkTools
using Statistics
import DataFrames
import MLJModelInterface as MMI

const MODELS  = models() do m
    !(m.package_name in ["MLJText"]) &&
        AbstractVector{Continuous} <: m.target_scitype &&
        m.is_supervised
end

# This is a way to load all needed model code:
MLJTestInterface.test(MODELS, mod=@__MODULE__, level=1, throw=true)

rng = Random.MersenneTwister(0)
Xmat = randn(rng, 30, 3)
X = Tables.table(Xmat)
y = @. cos(Xmat[:, 1] * 2.1 - 0.9) * Xmat[:, 2] - Xmat[:, 3]

stats = []
for m in MODELS
    print("\rBenchmarking $(m.name) $(m.package_name).")
    model = eval(:(@load $(m.name) pkg=$(m.package_name) verbosity=0))()
    mach = machine(model, X, y)
    fit!(mach, verbosity=0)
    fitresult = mach.fitresult
    b_high = @benchmark predict($mach, $X)
    b_low = @benchmark predict($model, $fitresult, $(MMI.reformat(model, X))...)
    slow_down = median(b_high.times)/median(b_low.times)
    bloat = b_high.allocs/b_low.allocs
    push!(stats, (; model=m.name, pkg=m.package_name, slow_down, bloat))
    print(" Done.                           ")
end

@show length(MODELS)
#length(MODELS) = 69

stats = DataFrames.DataFrame([stats...])
filter(stats) do row
    row.slow_down > 1.75 || row.bloat > 2.0
end
```

</details>

In the tables below: 

- "slow_down" is the ratio of elapsed **time** of "high" to "low"
- "bloat" is the ratio of number of **allocations** of "high" to "low"

Only models with `slow_down > 1.75` or `bloat > 2` are reported.

### Before this PR

```julia
#  Row │ model                           pkg                           slow_down  bloat
#      │ String                          String                        Float64    Float64
# ─────┼───────────────────────────────────────────────────────────────────────────────────
#    1 │ CatBoostRegressor               CatBoost                       99.694    20.875
#    2 │ ConstantRegressor               MLJModels                       8.50998   4.0
#    3 │ DeterministicConstantRegressor  MLJModels                      12.8771    4.0
#    4 │ ElasticNetRegressor             MLJLinearModels                 4.67751   2.0
#    5 │ EvoLinearRegressor              EvoLinear                      18.5536    7.33333
#    6 │ EvoSplineRegressor              EvoLinear                       4.11118   2.05556
#    7 │ HuberRegressor                  MLJLinearModels                 4.18327   2.0
#    8 │ LADRegressor                    MLJLinearModels                 4.18333   2.0
#    9 │ LassoRegressor                  MLJLinearModels                 4.07216   2.0
#   10 │ LinearRegressor                 MLJLinearModels                 4.05314   2.0
#   11 │ LinearRegressor                 MultivariateStats               4.77403   2.5
#   12 │ PLSRegressor                    PartialLeastSquaresRegressor    2.11986   1.375
#   13 │ QuantileRegressor               MLJLinearModels                 4.33039   2.0
#   14 │ RidgeRegressor                  MLJLinearModels                 4.12531   2.0
#   15 │ RidgeRegressor                  MultivariateStats               4.8491    2.5
#   16 │ RobustRegressor                 MLJLinearModels                 4.17605   2.0
```

### After this PR:

```
#  Row │ model               pkg        slow_down  bloat
#      │ String              String     Float64    Float64
# ─────┼────────────────────────────────────────────────────
#    1 │ CatBoostRegressor   CatBoost    96.6566   20.8125
#    2 │ ConstantRegressor   MLJModels    1.55635   3.0
#    3 │ EvoLinearRegressor  EvoLinear   19.9274    7.0
#    4 │ EvoSplineRegressor  EvoLinear    4.52508   2.0
#    5 │ EvoTreeMLE          EvoTrees     2.46669   1.00031
```

Note this machines serialised using #master cannot be deserialised after this PR. But I don't consider this triggers a breaking release. 

To do:

- [ ] Run MLJ tests with integration tests switched on

### DIFF
--- a/src/composition/learning_networks/nodes.jl
+++ b/src/composition/learning_networks/nodes.jl
@@ -27,9 +27,9 @@ See also [`node`](@ref), [`Source`](@ref), [`origins`](@ref),
 [`sources`](@ref), [`fit!`](@ref).
 
 """
-struct Node{T<:Union{Machine, Nothing}} <: AbstractNode
+struct Node{T<:Union{Machine, Nothing},Oper} <: AbstractNode
 
-    operation   # eg, `predict` or a static operation, such as `exp`
+    operation::Oper   # eg, `predict` or a static operation, such as `exp`
     machine::T  # is `nothing` for static operations
 
     # nodes called to get args for `operation(model, ...) ` or
@@ -43,9 +43,11 @@ struct Node{T<:Union{Machine, Nothing}} <: AbstractNode
     # order consistent with extended graph, excluding self
     nodes::Vector{AbstractNode}
 
-    function Node(operation,
-                  machine::T,
-                  args::AbstractNode...) where T<:Union{Machine, Nothing}
+    function Node(
+        operation::Oper,
+        machine::T,
+        args::AbstractNode...,
+        ) where {T<:Union{Machine, Nothing}, Oper}
 
         # check the number of arguments:
         # if machine === nothing && isempty(args)
@@ -70,7 +72,7 @@ struct Node{T<:Union{Machine, Nothing}} <: AbstractNode
                 vcat(nodes_, (nodes(n) for n in machine.args)...) |> unique
         end
 
-        return new{T}(operation, machine, args, origins_, nodes_)
+        return new{T,Oper}(operation, machine, args, origins_, nodes_)
     end
 end
 
--- a/src/machines.jl
+++ b/src/machines.jl
@@ -47,10 +47,10 @@ caches_data_by_default(m) = caches_data_by_default(typeof(m))
 caches_data_by_default(::Type) = true
 caches_data_by_default(::Type{<:Symbol}) = false
 
-mutable struct Machine{M,C} <: MLJType
+mutable struct Machine{M,OM,C} <: MLJType
 
     model::M
-    old_model  # for remembering the model used in last call to `fit!`
+    old_model::OM  # for remembering the model used in last call to `fit!`
     fitresult
     cache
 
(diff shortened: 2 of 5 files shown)


OkonSamuel (approved):
Looks good to me!!!. 
@ablaom I think we should benchmark the compile time for this package with vs without this PR.
But in general, I think run time performance gains we would get from this should outweigh any added compile time increase, because these operations (e.g predict, etc.) are expected in general to be quite expensive.

ablaom:
Doesn't look like there's a significant difference. 

Before this PR:

```julia
julia> @time_imports import MLJBase
    413.2 ms  MLJBase 23.21% compilation time
```

After this PR:

```julia
@time_imports import MLJBase
    437.3 ms  MLJBase 22.21% compilation time
```

PR #972, opened 2024-04-23
Make test of `iterator(...)` more robust

### LINKED ISSUE #971: Tests failing on dev
This may be intermittent. Observed at #968  and locally (but not at #969).

ablaom:
Resolves #971

I am perplexed about the sudden appearance of the fail in #971. It seems like an RNG issue, but we use StableRNGs in the  testing and the git blame shows no change to the relevant MLJBase code in > 4 years. Interestingly, #969, which has passed tests, now fails for me locally also. Tests are failing locally for me in Julia 1.6 and Julia 1.10.

The test currently failing relies completely on the stability of the RNG used. In this PR I replace the test with a more robust one, which still uses StableRNGs but only checks that the RNG-generated counts are within their expected 95% confidence intervals, instead of checking their precise values. This is a more correct test anyhow. 

I played around locally to ensure the tests only fails about 5% of the time, and then fixed the seed to ensure a pass. So if the "stable" RNG behaviour changes again unexpectantly, we have a good chance the test will survive the change 😄 .

### DIFF
--- a/test/hyperparam/one_dimensional_range_methods.jl
+++ b/test/hyperparam/one_dimensional_range_methods.jl
@@ -233,31 +233,31 @@ end
 @testset "NominalSampler" begin
 
     r = range(Char, :(model.dummy), values=collect("cab"))
+    N = 10000
+
+    # to compute half-width of 95% confidence intervals, for counts of a Bernoulli process
+    # with probability `p`, sampled `N` times:
+    halfwidth(p, N) = 1.96*sqrt(p*(1 - p))*sqrt(N)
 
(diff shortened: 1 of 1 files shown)


OkonSamuel:
It looks like an issue with `StableRNG` to me.

ablaom:
Yes, you are right. Somehow I missed the dependence on Distributions and thought this was Random. Great diagnosis, thank you. 

With the source of the issue put to rest, any objections to my solution?

OkonSamuel (approved):
LGTM!!!

PR #975, opened 2024-04-24
Add `InSample` resampling strategy

### LINKED ISSUE #967: Add `InSample` resampling strategy
When you want to train on all specified data and test on the same data.

So, 

```julia
e = evaluate(model, X, y, resampling=InSample(), measure=l1)
e.measurement[1]
```

is equivalent to:

```julia
mach = machine(model, X, y) |> fit!
yhat = predict(mach, X)
l2(yhat, y)
```

Which is even more convenient when there is battery measures to be evaluated.

ablaom:
Closes #967

To do post-PR:

- re-export `InSample` from MLJ? (tracked)

### DIFF
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -111,6 +111,50 @@ function shuffle_and_rng(shuffle, rng)
 end
 
 # ----------------------------------------------------------------
+# InSample
+
+"""
+    in_sample = InSample()
+
+Instantiate an `InSample` resampling strategy, for use in `evaluate!`, `evaluate` and in
+tuning. In this strategy the train and test sets are the same, and consist of all
+observations specified by the `rows` keyword argument. If `rows` is not specified, all
+supplied rows are used.
+
+# Example
+
+```julia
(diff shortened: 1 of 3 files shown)


src/resampling.jl
```diff
@@ -345,7 +389,7 @@ end
                                    rng=Random.GLOBAL_RNG)
 
 Stratified cross-validation resampling strategy, for use in
-`evaluate!`, `evaluate` and in tuning. Applies only to classification
+`evaluate!`, `evaluate` and intuning. Applies only to classification
```
OkonSamuel, line 392: Did you really mean "intuning" or this is an error?

OkonSamuel (approved):
Just a minor comment. Asides that LGTM!!.

src/resampling.jl
ablaom: Good catch, thanks.

ablaom:
Thankyou @OkonSamuel for your review.

PR #973, opened 2024-04-24
Add `CompactPerformanceEvaluation` objects and the option in `evaluate!` to construct them

ablaom:
Partly addresses https://github.com/alan-turing-institute/MLJ.jl/issues/1105, and prepares for https://github.com/JuliaAI/MLJTuning.jl/pull/215.

~~This PR also drops the `per_fold` part of `PerformanceEvaluation` display. More often than not, the table does not fit with this in it. The display still includes the names of all accessible fields.~~ This PR also adjusts the display of `per_fold` information to mitigate issue with overly wide table.

Previously:

```julia
PerformanceEvaluation object with these fields:
  model, measure, operation, measurement, per_fold,
  per_observation, fitted_params_per_fold,
  report_per_fold, train_test_rows, resampling, repeats
Extract:
┌─────────────┬──────────────┬─────────────┬─────────┬──────────────────────────────────────
│ measure     │ operation    │ measurement │ 1.96*SE │ per_fold                            ⋯
├─────────────┼──────────────┼─────────────┼─────────┼──────────────────────────────────────
│ BrierLoss() │ predict      │ 0.933       │ 0.184   │ [0.719, 1.16, 0.844, 1.16, 0.667, 0 ⋯
│ Accuracy()  │ predict_mode │ 0.0         │ 0.0     │ [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]      ⋯
└─────────────┴──────────────┴─────────────┴─────────┴──────────────────────────────────────
 ```

After this PR:

```julia
PerformanceEvaluation object with these fields:
  model, measure, operation,
  measurement, per_fold, per_observation,
  fitted_params_per_fold, report_per_fold,
  train_test_rows, resampling, repeats
Extract:
┌───┬─────────────┬──────────────┬─────────────┐
│   │ measure     │ operation    │ measurement │
├───┼─────────────┼──────────────┼─────────────┤
│ A │ BrierLoss() │ predict      │ 0.933       │
│ B │ Accuracy()  │ predict_mode │ 0.0         │
└───┴─────────────┴──────────────┴─────────────┘
┌───┬──────────────────────────────────────────┬─────────┐
│   │ per_fold                                 │ 1.96*SE │
├───┼──────────────────────────────────────────┼─────────┤
│ A │ [0.719, 1.16, 0.844, 1.16, 0.667, 0.914] │ 0.184   │
│ B │ [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]           │ 0.0     │
[...]

PR #977, opened 2024-05-06
For a 1.3 release

ablaom:
#968 #975 #973 #969


ablaom:
For the release notes:

- (**Performance enhancement**) Remove type instability for `predict(mach::Machine, ...)` in the easy and typical case that `mach` does not wrap a `Symbol` model (#969)

- (**New feature**) Give `evaluate` and `evaluate!` the option `compact=true`, to return  a `CompactPerformanceEvaluation` object with minimal memory footprint (#973)

- (**New feature**) Add an `InSample()` resampling strategy that trains and tests on the same data (whatever is specified by `rows`, or all supplied data) (#975)

- (**Display improvement**) Split the table displayed as part of an `PerformanceEvaluation` object over two tables, if needed, to deal with overly wide tables (#973)

Issue #926, opened 2023-08-17
Add facility to specify a default logger different from `nothing`


ablaom:
This could be provided as either part of #925 or as a follow-up PR.

The idea is to define a global variable `DEFAULT_LOGGER` and to make the default value of the `logger` keyword option in `evaluate!` equal to `DEFAULT_LOGGER`, rather than `nothing`. (And later, do the same for `TunedModel` and `IteratedModel`; see https://github.com/alan-turing-institute/MLJ.jl/issues/1029.)

In `__init__` the `DEFAULT_LOGGER` would be set to `nothing` but MLJBase will provide a public method `default_logger` to set and inspect the value. For example:

```julia
# inspecting:
julia> default_logger()
nothing

# changing:
using MLFlow
julia> default_logger(MLFlowLogger())
MLFlowLogger(MLFlow(
    baseuri = "http://127.0.0.1:5000", 
    apiversion = 2.0
), "MLJ experiment", nothing)
```

The required code can be modelled exactly on the `DEFAULT_RESOURCE` global variable (see src/init.jl) and the `default_resource` setter/getter (see src/MLJBase.jl).

To do after this integration:

- [ ] Have MLJ re-export `default_logger` 

@pebeto

pebeto:
Does this change replace our keyword argument in `evaluate!` and `save`?

ablaom:
Yes, everywhere we have a keyword argument `logger=nothing` , we need to change to `logger=default_logger()`. At present, that's just `evaluate!`. 

In the case of `MLJBase.save`, there is nothing in MLJBase that connects `save` to logging. That happens only in MLJFlow.jl, where `MLJBase.save(::MLFlowLogger, ::Machine)` is overloaded. I suppose we could arrange that `MLBase.save(mach)` saves the machine to the default logger, ie, do something like this in MLJBase:

```julia
const ERR_INVALID_DEFAULT_LOGGER = ArgumentError(
    "`default_logger()` is currently `nothing`. Either specify an explicit path or stream as "
    "target of the save, or "use `default_logger(logger)` to change the default logger. "
)
"""
    <new doc string>
"""
MLJBase.save(mach::Machine) = MLJBase(default_logger, mach)
MLJBase.save(::Nothing, ::Machine) = throw(ERR_INVALID_DEFAULT_LOGGER)
```

pebeto:
[...]

PR #979, opened 2024-05-28
Add a default_logger

ablaom:
A step towards resolution of #926; see also https://github.com/JuliaAI/MLJ.jl/issues/1124.

This PR:

- Adds a global constant, initialized in `__init__`, for a default logger, for use with MLJFlow, for example. 
- Adds a function `default_logger` for returning or resetting the default logger
- Changes the default value of the `logger` kwarg in `evaluate`, `evaluate!`, and `Resampler` from `nothing` to `default_logger()`.
- Implements `MLJBase.save(mach) = MLJBase.save(default_logger(), mach)` to save machines (as artifacts) at the default logger (assuming it supports this).

To do:

- [x] Add some tests

### DIFF
--- a/src/MLJBase.jl
+++ b/src/MLJBase.jl
@@ -248,7 +248,8 @@ export params
 # -------------------------------------------------------------------
 # exports from this module, MLJBase
 
-# computational_resources.jl:
+# get/set global constants:
+export default_logger
 export default_resource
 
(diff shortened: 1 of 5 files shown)


ablaom:
@pebeto Any chance you could review this?

pebeto (approved):

PR #985, opened 2024-07-01
Relax restrictions on model type in resampling (`evaluate!`)

ablaom:
### Current behaviour

Currently a model in `evaluate`/`evaluate!` must subtype `Supervised` or `Annotator`. However, in the current implementation, all that's essential for the model is: 

- It implements a `predict` method
- It's `fit` method consumes a target `y` (which must be provided in the `evaluate` call, or `machine` constructor).
-  In the common case that `operation` is left unspecified, `MLJModelnterface.prediction_type` must be one of: `:deterministic`, `:probabilistic`, `:interval`. 

Subtypes of `Annotator` and `Supervised` (e.g,`Probabilistic`, `Deterministic`, `Interval`) have the obvious `prediction_type` fallbacks, but the fallback for `Unsupervised` models is `:unknown`. 

### What this PR does

This PR:

- Removes the current model type restriction: Now any `<:Model` is now allowed.
- Adds the currently missing check that a target `y` has been provided.
- In the case that an operation is not specified, and the `prediction_type` is not one of `:deterministic`, `:probabilistic`, `:interval`, the error which is already thrown in that case is made more informative. It now suggests two remedies: (i) explicitly specify `operation`(s), or (ii) post an issue to have the model's `prediction_type` reviewed. 

### Context

More and more we are relying on traits and less on the abstract model hierarchy. Currently, however, only `Unsupervised` models can have their `transform` method propagated in a `Pipeline` (and `predict` is propagated by `Supervised` models). So some models want to be `Unsupervised` for pipeline use, despite the fact they are also "supervised" - a case in point is `RecursiveFeatureElimination` (currently `Supervised` and so not pipelinable). 

### To do

Just to be sure, let's wait for 

- [x] Run local MLJ integration tests (@ablaom)

### Afterword

We should ultimately like to evaluate models that do *not* consume a target in training, such as clustering models. In that case, the target is only supplied for the test sets (e.g., for computing the [Rand index](https://en.wikipedia.org/wiki/Rand_index)). However, that requires a more substantial redesign of `evaluate`.

### DIFF
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -31,10 +31,6 @@ const ERR_INVALID_OPERATION = ArgumentError(
 _ambiguous_operation(model, measure) =
     "`$measure` does not support a `model` with "*
     "`prediction_type(model) == :$(prediction_type(model))`. "
-err_ambiguous_operation(model, measure) = ArgumentError(
-    _ambiguous_operation(model, measure)*
-    "\nUnable to infer an appropriate operation for `$measure`. "*
-    "Explicitly specify `operation=...` or `operations=...`. ")
 err_incompatible_prediction_types(model, measure) = ArgumentError(
     _ambiguous_operation(model, measure)*
     "If your model is truly making probabilistic predictions, try explicitly "*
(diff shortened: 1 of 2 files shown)


OkonSamuel (approved):
Looks Good to me!!!.

PR #1041, opened 2025-12-18
Stratify shuffle fix

tlienart:
This fixes a docstring and a bug

1. docstring: if a stratification vector is provided, the user should use `shuffle=true` to avoid bias that may be caused by dataset ordering 
2. double lookup in the case of stratification with vector and with shuffle replacement of

```julia
idxs  = [[i for i in rows if stratify[rows[i]] === v] for v in uv]
```

by

```julia
idxs  = [[i for i in eachindex(rows) if stratify[rows[i]] === v] for v in uv]
```

I'm a bit surprised this wasn't caught by the test suite but added an explicit extra test that specifically that path is correct where it wouldn't have been before.

### DIFF
--- a/src/data/data.jl
+++ b/src/data/data.jl
@@ -136,10 +136,9 @@ julia> (Xtrain, Xtest), (ytrain, ytest) = partition((X, y), 0.8, rng=123, multi=
   used, can be an integer seed. If specified, and `shuffle ===
   nothing` is interpreted as true.
 
-* `stratify=nothing`: if a vector is specified, the partition will
-  match the stratification of the given vector. In that case,
-  `shuffle` should be `true` to avoid any bias that may
-  come from the ordering of the rows.
+* `stratify=nothing`: if a vector is specified, the partition will match the
+  stratification of the given vector. In that case, `shuffle` should be `true` (or an
+  `rng` specified) to avoid any bias that may come from the ordering of the rows.
 
 * `multi=false`: if `true` then `X` is expected to be a `tuple` of
   objects sharing a common length, which are each partitioned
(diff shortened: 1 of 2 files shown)


ablaom (commented):
Thanks very much indeed for this PR and the prompt attention to my offline query. 

Is there a use-case for *not* specifying `shuffle=true`, the bias notwithstanding? If not, I'd prefer to throw an error here. What do you think?

ablaom:
Oh, I guess, the data might already be shuffled. Perhaps that was your original thinking (over 6 years ago?). Perhaps we do not add the warning, continue to allow `shuffle=false`, but just clarify the docs a bit to avoid head scratching.

tlienart:
that sounds good, could you maybe take over the docstring adjustments?

andreuvall:
That's interesting! I now see that, in @tlienart's view, it wasn't the docstring which was wrong but [this line of code](https://github.com/JuliaAI/MLJBase.jl/pull/1041/changes#diff-f8b609a9c46a4d0376c6dca85b606d5560c2eb66254919d6b05db65959e3bb5fL49).

I see it the other way around. I don't think it is MLJ's role to prevent the user from preparing the data this or that way. The risk of bias caused by dataset ordering is always there, with or without stratification. Why should MLJ suppose that the user doesn't know what they are doing? 

In summary, to me, the code seems correct, but the docstring is wrong (not so much because it doesn't describe what the code does, but rather because I disagree with MLJ constraining the possible data preparation options).

ablaom:
I agree. I have reverted the logic (no warning) but kept @tlienart's clarification of the docstring and bug fix.

PR #1044, opened 2026-02-12
Allow predicates in `unpack` to act on string names instead of symbols

### LINKED ISSUE #1043: Have `unpack` support `String` names for the columns
As in 

```julia
data = load_iris()
y, X = unpack(==("target")) # now only `==(:target)` is supported now
```

This can lower the congnitive burden for newcomers familiar with DataFrames.jl which supports string indexing of columns.

ablaom:
This PR resolves #1043.

After this PR:

```julia
julia> table = DataFrame(x=[1,2], y=['a', 'b'], z=[10.0, 20.0], w=["A", "B"])
2×4 DataFrame
 Row │ x      y     z        w      
     │ Int64  Char  Float64  String 
─────┼──────────────────────────────
   1 │     1  a        10.0  A
   2 │     2  b        20.0  B

julia> YW, _ = unpack(table, in(["y", "w"]); string_names=true)
julia> YW
2×2 DataFrame
 Row │ y     w      
     │ Char  String 
─────┼──────────────
   1 │ a     A
   2 │ b     B
```

### DIFF
--- a/src/data/data.jl
+++ b/src/data/data.jl
@@ -212,13 +212,14 @@ applied to the column names. Selection from the column names is without replacem
 column except the one with name `:id`.
 
 Predicates may also be formulated with the understanding that column names are strings
-instead of symbols, by specifying `string_names=true`, or by
+instead of symbols, by specifying `string_names=true`.
 
 
 Returns a tuple of tables/vectors with length one greater than the number of supplied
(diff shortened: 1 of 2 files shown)


ablaom:
I'm thinking there is a better solution. See the discussion at #1043 .

PR #1066, opened 2026-08-19
Clarify update log message: warm restart is not guaranteed

### LINKED ISSUE #1065: Improve misleading message on `fit!(mach)` in case `update` is dispatched.
As flagged [here](https://github.com/Evovest/EvoTrees.jl/pull/347#discussion_r3793261214): If `update` is dispatched then "Updating..." is logged by `fit_only!(mach)` but it may happen that `update` decides training from scratch is required after all, dending on model-specific logic. So this isn't a "update" in the sense of a warm restart. 

Since the deliberations are internal to `update`, their outcomes are not accessible in the current design, so the message should just be made clearer. That said, this could be complicated, because some testing around warm restarts and smart training of learning networks may depend on `@test_logs` and the exact wording of this message.  Here and in other repositories, such as MLJFlux and MLJIteration. There is a more robust test [`@test_model_sequence`](https://github.com/JuliaAI/MLJBase.jl/blob/6b9ef82b4069298a6fcb691882eb2438f4907176/test/test_utilities.jl#L87) but I don't think this was rolled out everywhere.

AJ0070:
Closes #1065.

Wording as agreed there. No new action symbol: `:update` covers both the row-change branch and the model-change branch, and "with warm restart if possible" is accurate for both. Adding a separate symbol changes the `MACHINE_CHANNEL` sequences that `@test_mach_sequence` asserts on, which breaks `test/machines.jl:417` among others.

### DIFF
--- a/src/machines.jl
+++ b/src/machines.jl
@@ -535,7 +535,7 @@ function fitlog(mach, action::Symbol, verbosity)
         put!(MACHINE_CHANNEL, (action, mach))
     elseif verbosity > 0
         action == :train && (@info "Training $mach."; return)
-        action == :update && (@info "Updating $mach."; return)
+        action == :update && (@info "Updating $mach (with warm restart if possible)."; return)
         action == :skip && begin
             @info "Not retraining $mach. Use `force=true` to force."
             return


ablaom (approved):
Looks good to me.

